PR #57, opened 2021-04-28
fix: syntax check in sdk create and drop db

### LINKED ISSUE #46: CreateDB API does not follow the syntax rules  leads that  use Database Command does not work

## Expected Behavior
CreateDB("121314") return false

## Current Behavior
CreateDB("121314") return true
image

## Steps to Reproduce

1.
2.
3.
4.

vagetablechicken:

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)

Add syntax check in create/drop db, by using hybridse parser.

Now, sdk allows any name. But hybridse parser only allows [^`/\\.\n]+.
So if db name contains the forbidden characters, we shouldn't pass it.

Nope.

### DIFF
--- a/src/sdk/sql_cluster_router.cc
+++ b/src/sdk/sql_cluster_router.cc
@@ -633,6 +633,13 @@ bool SQLClusterRouter::CreateDB(const std::string& db,
     if (status == NULL) {
         return false;
     }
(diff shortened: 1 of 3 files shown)


imotai (approved):
LGTM

PR #70, opened 2021-05-14
feat: restful api server

### LINKED ISSUE #48: Add API Server
HTTP interface is one of the most friendly interface and many developers like using http interface to develop their program. REST API also help us to show demo simply. So we want to support uing http api to access FEDB

Please refer [rfc](https://github.com/dl239/rfcs/blob/main/fedb/api-server.md) for more information

vagetablechicken:
Closes #48
Restful api server uses brpc http server to accept http requests, then the inner fedb client will forward the requests to fedb.
The url parser supports to parse urls like "/a/:arg1/b/:arg2/:arg3", but doesn't support wildcards.
Use `fedb --role=apiserver ...` to start it.


src/apiserver/api_server_impl.cc
```diff
@@ -0,0 +1,300 @@
...
+    return true;
+}
+
+void APIServiceImpl::Process(google::protobuf::RpcController* cntl_base, const HttpRequest*, HttpResponse*,
+                             google::protobuf::Closure* done) {
+    brpc::ClosureGuard done_guard(done);
+    auto* cntl = dynamic_cast<brpc::Controller*>(cntl_base);
+
+    // The unresolved path has no slashes at the beginning, it's not good for url parsing
+    auto unresolved_path = "/" + cntl->http_request().unresolved_path();
```
imotai, line 60: check if the path starts with `/`

src/apiserver/api_server_impl.cc
```diff
@@ -0,0 +1,300 @@
...
+    auto method = cntl->http_request().method();
+    LOG(INFO) << "unresolved path: " << unresolved_path << ", method: " << HttpMethod2Str(method);
```
imotai: better to use `DLOG`

src/apiserver/api_server_impl.cc
vagetablechicken: brpc ensures that, cntl.http_request().unresolved_path() is always normalized, namely no slashes at the beginning or the end, and no repeated slashes in the middle. Ref https://github.com/apache/incubator-brpc/blob/master/docs/en/http_service.md#restful-url

src/apiserver/api_server_impl.cc
vagetablechicken: fixed

src/apiserver/interface_provider.cc
```diff
@@ -0,0 +1,197 @@
...
+}
+
+Url ReducedUrlParser::parse(std::string const& urlString) {
+    Url url;
+    url.url = urlString;
+
+    // regex for extracting path, query, fragment
+    // (?:(?:(\/(?:(?:[a-zA-Z0-9]|[-_~!$&']|[()]|[*+,;=:@])+(?:\/(?:[a-zA-Z0-9]|[-_~!$&']|[()]|[*+,;=:@])+)*)?)|\/)?
+    // (?:(\?(?:\w+=(?:[\w-])+)(?:(?:&|;)(?:\w+=(?:[\w-])+))*))?(?:(#(?:\w|\d|=|\(|\)|\\|\/|:|,|&|\?)+))?)
+    std::regex rgx(
```
imotai: Is there a perfermance issue if we initialize `regex` when using it.

src/apiserver/api_service_impl.cc
```diff
@@ -0,0 +1,301 @@
...
+namespace http {
+
+APIServiceImpl::~APIServiceImpl() = default;
+
+bool APIServiceImpl::Init(const sdk::ClusterOptions& options) {
+    // If cluster sdk is needed, use ptr, don't own it. SQLClusterRouter owns it.
+    auto cluster_sdk = new ::fedb::sdk::ClusterSDK(options);
+    bool ok = cluster_sdk->Init();
+    if (!ok) {
+        LOG(ERROR) << "Fail to connect to db";
```
imotai: using `WARNING` to avoid `core dump`

src/apiserver/api_service_impl.h
```diff
@@ -0,0 +1,269 @@
...
+        default: {
+            LOG(ERROR) << "Invalid Column Type";
+            ar & "err";
+            break;
+        }
+    }
+}
+
+template <typename Archiver>
+Archiver& operator&(Archiver& ar, const ExecSPResp& s) {  // NOLINT
```
dl239: why overload the global &

src/apiserver/interface_provider.cc
vagetablechicken: I'll test it.

src/apiserver/api_service_impl.cc
vagetablechicken: `ERROR` won't cause core dump, `FATAL` will.

src/apiserver/api_service_impl.h
vagetablechicken: In this way, we can use 'writer & <struct>' to write, simpler to write.

src/apiserver/interface_provider.cc
vagetablechicken: |Benchmark           |         Time       |      CPU  | Iterations|
|---|---|---|---|
|BM_UrlParse         |    817332 ns    |   817312 ns    |      696|
|BM_UrlParseWithInit  |      787 ns     |     787 ns     |  896177|

It's better to init `std::regex` first, instead of creating it in every `parse`. I'll fix it.

src/apiserver/json_writer.h
```diff
@@ -0,0 +1,69 @@
...
+
+    static const bool IsReader = false;
+    static const bool IsWriter = !IsReader;
+
+    JsonWriter(const JsonWriter&) = delete;
+    JsonWriter& operator=(const JsonWriter&) = delete;
+
+ private:
+    // PIMPL idiom
+    void* mWriter;  ///< JSON writer.
```
imotai: `writer_`

src/apiserver/api_service_impl.cc
```diff
@@ -0,0 +1,301 @@
...
+        }
+        const auto& arr = value[0];
+        StringBuffer buffer;
+        Writer<StringBuffer> sql_writer(buffer);
+        arr.Accept(sql_writer);
+        std::string line(buffer.GetString());
+        std::string insert_sql = "insert into " + table + " values(" + line.substr(1, line.length() - 2) + ");";
+
+        hybridse::sdk::Status status;
+        auto ok = sql_router_->ExecuteInsert(db, insert_sql, &status);
```
imotai: It's better to use `PreparedStatement` or Apiserver will parse sql for every request.

imotai: Just take a look at https://github.com/4paradigm/fedb/blob/main/src/sdk/sql_cluster_router.h#L123

vagetablechicken: Oops, I thought `SQLInsertRow` doesn't have the schema before. I'll try `SQLInsertRow`.

src/apiserver/json_helper.cc
```diff
@@ -0,0 +1,318 @@
...
+}
+
+JsonWriter& JsonWriter::Member(const char* name) {
+    WRITER->String(name, static_cast<SizeType>(strlen(name)));
+    return *this;
+}
+
+bool JsonWriter::HasMember(const char*) const {
+    // This function is for JsonReader only.
+    assert(false);
```
vagetablechicken: todo

src/apiserver/api_service_impl.cc
```diff
@@ -0,0 +1,317 @@
...
+        if (ok) {
+            PutResp resp;
+            writer& resp;
```
vagetablechicken: add a layer, e.g. writer >> obj, reader << obj, still use operator& inside. 
More easier for caller.

src/apiserver/api_service_impl.h
```diff
@@ -0,0 +1,284 @@
...
+    void RegisterExecSP();
+    void RegisterGetSP();
+
+    static SimpleSchema TransToSimpleSchema(const hybridse::sdk::Schema* schema) {
+        SimpleSchema ss;
+        for (int i = 0; i < schema->GetColumnCnt(); ++i) {
+            ss.emplace_back(schema->GetColumnName(i), schema->GetColumnType(i), schema->IsConstant(i),
+                            schema->IsColumnNotNull(i));
+        }
+        return std::move(ss);
```
aceforeverd: I think `std::move` is not neccessary, there is copy elision

vagetablechicken: fixed

src/apiserver/api_service_impl.cc
vagetablechicken: fixed

src/apiserver/interface_provider.h
```diff
@@ -0,0 +1,190 @@
...
+class ReducedUrlParser {
+ public:
+    /**
+     *  parses a urlString to an url object.
+     */
+    bool parse(std::string const& urlString, Url* url);
+
+ private:
+    void parseQuery(std::string const& query, Url* url);
+    std::regex reg_{
```
imotai: it's better to add a static modifier

vagetablechicken: ok

src/apiserver/api_service_impl.cc
```diff
@@ -0,0 +1,371 @@
...
+
+        hybridse::sdk::Status status;
+        auto sp_info = sql_router_->ShowProcedure(db, sp, &status);
+        if (!sp_info) {
+            writer << err.Set(status.msg);
+            return;
+        }
+
+        GetSPResp resp;
+        resp.data.name = sp_info->GetSpName();
```
dl239: unnecessary data copy

src/apiserver/api_service_impl.h
```diff
@@ -0,0 +1,287 @@
...
+        ar.SetNull();
+        return;
+    }
+    switch (schema[i].type) {
+        case hybridse::sdk::kTypeInt32: {
+            ar & rs->GetInt32Unsafe(i);
+            break;
+        }
+        case hybridse::sdk::kTypeInt64: {
+            ar & rs->GetInt64Unsafe(i);
```
dl239: use "<<"

src/apiserver/api_server_test.cc
```diff
@@ -0,0 +1,304 @@
...
+    ASSERT_EQ(6, arr.Size());
+    ASSERT_EQ(butil::rapidjson::kStringType, arr[0].GetType());
+    ASSERT_EQ(butil::rapidjson::kNumberType, arr[1].GetType());
+    ASSERT_EQ(butil::rapidjson::kNumberType, arr[2].GetType());
+    ASSERT_EQ(butil::rapidjson::kStringType, arr[3].GetType());
+    ASSERT_EQ(butil::rapidjson::kNumberType, arr[4].GetType());
+    ASSERT_EQ(butil::rapidjson::kTrueType, arr[5].GetType());
+
+    butil::rapidjson::StringBuffer buffer;
+    butil::rapidjson::Writer<butil::rapidjson::StringBuffer> writer(buffer);
```
dl239: Add illegal url case

src/proto/http.proto
```diff
@@ -0,0 +1,10 @@
+package fedb.http;
+
+option cc_generic_services = true;
+
+message HttpRequest {};
+message HttpResponse {};
+
+service APIService {
```
dl239: rename to  APIServer and update the api_service_impl file name

src/apiserver/api_service_impl.cc
vagetablechicken: You mean the string copy?

src/proto/http.proto
vagetablechicken: It's for `brpc::Server::AddService`, `Server` is not a good name.

imotai: Just follow the naming style like `NameServer` and `TabletServer`

src/apiserver/api_service_impl.cc
dl239: string and vector

src/apiserver/api_service_impl.h
vagetablechicken: It's for both writer & reader, can't use "<<"

src/proto/http.proto
vagetablechicken: We will `StartAPIServer` in fedb.cc, not here

src/apiserver/api_server_test.cc
vagetablechicken: added

src/apiserver/api_service_impl.cc
vagetablechicken: If we really need to avoid data copy, it's better to get the result from sdk.
I'll reconsider about the schema copy. If it's hard to read, I'd prefer clean code.

vagetablechicken:
- [x] when need_schema==false,  ExecSPResp::data::schema is empty, can't use this to parse result set
- [x] date type in response

src/proto/http.proto
dl239: 1 Our module is called APIServer not APIService. 2 There is no format that if use AddXXX method the variables must be named XXX. 3 Just follow the existing naming style

src/apiserver/api_service_impl.cc
dl239: Why not use ProcedureInfo to pack json directly. Many responses wrapped in different structs is clean code and easy to read?

vagetablechicken: > Why not use ProcedureInfo to pack json directly. Many responses wrapped in different structs is clean code and easy to read?

The point is that we can JsonRead the response directly, so I transformed the schema(However, the data can't read). Pack is easy, but the unpack is too complicated. This pr will only support JsonWriter for execSp & getSp. Maybe redesign in the next time.

src/proto/http.proto
vagetablechicken: > 1 Our module is called APIServer not APIService. 2 There is no format that if use AddXXX method the variables must be named XXX. 3 Just follow the existing naming style

AddService(AServer) is really ok?

vagetablechicken: fixed

dl239 (approved):

imotai (approved):
LGTM

PR #74, opened 2021-05-21
feat: remove ts_name in request proto

### LINKED ISSUE #67: Do not use ts name in scan and get
Because we had updated the structure of index,  query request such as scan/get/traverse/count need not to use ts_name.

dl239:
Because we had updated the structure of index, query request such as scan/get/traverse/count need not to use ts_name.


imotai (approved):
LGTM

src/proto/tablet.proto
```diff
@@ -242,14 +242,13 @@ message ScanRequest {
...
     optional string idx_name = 8;
-    optional string ts_name = 9;
-    optional GetType st_type = 10 [default = kSubKeyLe];
-    optional GetType et_type = 11 [default = kSubKeyGt];
-    optional bool enable_remove_ts = 12 [default = false];
-    optional uint32 atleast = 13 [default = 0];
-    repeated uint32 projection = 14;
-    repeated uint32 pid_group = 15;
-    optional bool use_attachment = 16 [default = false];
+    optional GetType st_type = 9 [default = kSubKeyLe];
```
vagetablechicken, line 245: The unique numbered tags changed, do we consider about the backward compatibility?

dl239: Do not consider compatibility

vagetablechicken (approved):

PR #97, opened 2021-06-22
fix: api server api format bug and no db bug

altale:
Change some format in the response JSON
Fix the bug that happens when want to show all tables in a non-existent db


dl239 (approved):

src/apiserver/api_server_impl.cc
```diff
@@ -428,7 +428,25 @@ void APIServerImpl::RegisterGetTable() {
...
+                          writer << err.Set(status.msg);
+                          return;
+                      }
                       auto db = db_it->second;
+                      bool db_ok = std::find(dbs.begin(), dbs.end(), db) != dbs.end();
+                      if (!db_ok) {
+                          writer.StartObject();
+                          writer.Member("code") & -1;
+                          writer.Member("msg") & std::string("DB not found");
+                          writer.Member("tables");
```
vagetablechicken: If failed, do we need the 'tables' field? code & msg is enough?

altale: In my opinion, whether to keep the 'tables' field are both ok. Here I choose to keep it because: First, the additional content is only _"tables": []_. It is very short. Second, in my experience using MVVM front-end frameworks such as vue.js, it is better to keep it. Because the code would also work even if the developer doesn't do a special check. Nor the variable binding to the component will be an 'undefined'.

vagetablechicken:
But we’ve defined the error response format(code&msg only). If you think we need to keep the empty field, it should be consistent in all methods. Need more discussion.

src/apiserver/api_server_impl.cc
vagetablechicken: But we’ve defined the error response format(code&msg only). If you think we need to keep the empty field, it should be consistent in all methods. Need more discussion.

altale: Yes if other methods only have code & msg, deleting it is better.

src/apiserver/api_server_impl.cc
```diff
@@ -428,7 +428,22 @@ void APIServerImpl::RegisterGetTable() {
...
+                          writer << err.Set(status.msg);
+                          return;
+                      }
                       auto db = db_it->second;
+                      bool db_ok = std::find(dbs.begin(), dbs.end(), db) != dbs.end();
+                      if (!db_ok) {
+                          writer.StartObject();
+                          writer.Member("code") & -1;
+                          writer.Member("msg") & std::string("DB not found");
+                          writer.EndObject();
```
vagetablechicken: You could use `GeneralError` to simplify the code. No need to start/end object. Please ref other methods, e.g. `RegisterGetSP()`.

dl239 (approved):

PR #108, opened 2021-06-28
fix: delete showlimit cmd

### LINKED ISSUE #105: Delete setlimit cmd
Becase brpc does not support dynamic modification of concurrency in latest version, we plan to delete the setlimit cmd. More details refer [here](https://github.com/apache/incubator-brpc/blob/master/docs/en/server.md#limit-method-level-concurrency)

altale:
#105


dl239:
need delete relevant code in client/tablet_client.h client/tablet_client.cc  proto/tablet.proto  tablet/tablet_impl.cc tablet/tablet_impl.h

PR #200, opened 2021-08-03
fix: java smoke case

### LINKED ISSUE #166: openmldb-jdbc SQLRouterSmokeTest needs fix
Problem 1.
```
java.lang.ClassCastException: com._4paradigm.openmldb.jdbc.SQLResultSet cannot be cast to com._4paradigm.openmldb.ResultSet
        at com._4paradigm.openmldb.jdbc.SQLRouterSmokeTest.testSmoke(SQLRouterSmokeTest.java:87)
```
`executeSQL` returns java.sql.ResultSet(actually is `com._4paradigm.openmldb.jdbc.SQLResultSet`). Maybe we should cast it to `SQLResultSet`, cuz we want to get schema in test, can't use java.sql.ResultSet only.
But if we use `SQLResultSet`, there're so many Deprecated methods.  So shall we use `getFetchSize` to get result set size?

https://github.com/4paradigm/OpenMLDB/blob/9470d3f01dbda471b0710f130aaaafb15bc19a96/java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/SQLResultSet.java#L506-L512

And the getXX methods index is different wtih `com._4paradigm.openmldb.ResultSet`, e.g.

https://github.com/4paradigm/OpenMLDB/blob/9470d3f01dbda471b0710f130aaaafb15bc19a96/java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/SQLResultSet.java#L125-L131

- [ ] May have more problems, need check

dl239:
+ fix java smoke case
+ fix InsertPreparedStatement SetDate bug: Month should plus 1


imotai (approved):
LGTM

vagetablechicken (approved):

PR #300, opened 2021-08-24
fix: fix end2end test cases which may reuse the same spark session

tobegit3hub:
* Config skipAfterFailureCount for `mvn test`
* Fix the Spark util to check if dataframes are equal
* Update the  end2end unit tests to use the new SparkSession
* Add the utility method to disable Spark log
* Add the unit test for `approximateDfEqual`


zekai427 (approved):
LGTM

dl239 (approved):

PR #328, opened 2021-08-31
fix: importer fix and doc

vagetablechicken:
- must check csv header
- add doc


dl239 (approved):

zekai427 (approved):
LGTM

java/openmldb-import/src/main/java/com/_4paradigm/openmldb/importer/Importer.java
```diff
@@ -313,13 +322,13 @@ public static void main(String[] args) {
             return;
         }
 
-        logger.info("start bulk load");
+        System.out.println("start bulk load");
```
imotai, line 325: use `logger` to replace `system.out`

PR #341, opened 2021-09-01
fix: sql and ns client desc result

### LINKED ISSUE #331: bug: sql client `desc` result
Ns client `showschema` use `PrintColumnKey` to show indexes.
Currently, sql client `desc` can't show multi keys and ttl is not 'xx min'.
Sql client could use `PrintColumnKey` to reach an agreement.

vagetablechicken:
Closes #331 
Sql client & ns client use the same print functions.

### DIFF
--- a/src/cmd/display.h
+++ b/src/cmd/display.h
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include <base/texttable.h>
 #include <gflags/gflags.h>
 #include <snappy.h>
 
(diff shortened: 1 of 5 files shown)


src/cmd/display.h
```diff
@@ -25,6 +25,7 @@
 #include <utility>
 #include <vector>
 
+#include "base/texttable.h"
```
dl239, line 28: base/texttable.h is not exist

dl239 (approved):

zekai427 (approved):
LGTM

PR #381, opened 2021-09-07
fix: fix a bug in union with instance not in window

### LINKED ISSUE #374: Fix the issue to enable window skew optimization for window union case
Currently we fail to enable window skew optimization for some window union cases in integrated testing.

We need to address the issue and fix for more general purposes.

zekai427:
- Resolve https://github.com/4paradigm/OpenMLDB/issues/374
- Add a unit test for this bug fix


tobegit3hub (approved):
LGTM

dl239 (approved):

Issue #398, opened 2021-09-10
feat: support insert multiple rows into a table using a single SQL insert statement.


jingchen2222:
**Is your feature request related to a problem? Please describe.**

We are going to provide the ability to insert multiple database rows in a single SQL insert statement. The ANSI syntax for multi row is shown in the following demo.

**Example 1:** Insert 2 rows into table `t1`
```SQL
CREATE TABLE t1 (
col1 int not null,
col5 bigint not null,
std_ts timestamp not null,
index(key=(col1), ts=col5)
);
      
INSERT INTO t1 VALUES (1, 10000L, 1590738987000L), (2, 20000L, 1590738988000L);
```

**Example 2:** Insert 2 rows into table `t2` by specifying both the column names and the values to be inserted.
```SQL
CREATE TABLE t2 (
col1 int not null,
col2 smallint,
col3 float,
col4 double,
col5 bigint not null,
std_date date not null,
str string,
INDEX(key=(col1), ts=col5)
);
INSERT INTO t2 (col1, col5, std_date) VALUES (1, 10000L, '2020-05-27'), (2, 20000L, '2020-05-28');
```

**Describe the solution you'd like**

The problem is we only get one single row from given insert statement. So we are going to modify function `ExecuteInsert`.
```c++
bool SQLClusterRouter::ExecuteInsert(const std::string& db, const std::string& sql, ::hybridse::sdk::Status* status) {
  // TODO 
  // parse a sql into a set of rows and put rows into storage one by one
}
```
More related discussion can be checked from  #391.

**Describe alternatives you've considered**

**Additional context**

PR #399, opened 2021-09-10
feat: support insert multiple rows into a table using a single SQL insert statement

### LINKED ISSUE #391: Bug: SQL INSERT Statement with multi rows does not work as expected
I came across the bug when trying to insert multiple rows into a table with a INSERT Statement(the code below). The execution of the statement did not return any error. But when querying the table, only the first row had been inserted into the table. The bug exists both in sql_client and Python SDK of OpenMLDB, i haven't test it in Java SDK.

```sql
127.0.0.1:6527/>  CREATE DATABASE db_test;
Create database success
127.0.0.1:6527/> USE db_test;
Database changed
127.0.0.1:6527/db_test> CREATE TABLE user(userId string, createAt timestamp, index(key=userId, ts=createAt));
I0907 13:11:27.337503   163 zk_client.cc:56] node watcher with event type 3, state 3
127.0.0.1:6527/db_test> INSERT INTO user VALUES('user_0', 1628689181169), ('user_1', 1628689181169), ('user_2', 1628689181169);
127.0.0.1:6527/db_test> SELECT * FROM user;
 -------- ---------------
  userId   createAt
 -------- ---------------
  user_0   1628689181169
 -------- ---------------
```

## Expected Behavior
After success execution of the INSERT statement with multiple rows, the result of the querying should contains all the rows in the INSERT statement.

## Current Behavior
After success execution of the INSERT statement with multiple rows, the current result of the querying only contains the first row in the ```INSERT``` statement.

## Possible Solution
There are two possible solution for the bug.
- First One: Given an **Explicitly** error or warn when user trying to insert multiple rows in a ```INSERT``` statement.
- Second One: Support Insert multiple rows in a ```INSERT``` statement

I prefer to the second solution to support multiple rows inserting.

## Steps to Reproduce
```sql
CREATE DATABASE db_test;
USE db_test;
CREATE TABLE user(userId string, createAt timestamp, index(key=userId, ts=createAt));
INSERT INTO user VALUES('user_0', 1628689181169), ('user_1', 1628689181169), ('user_2', 1628689181169);
SELECT * FROM user;
```
## Context (Environment)
- Image: ghcr.io/4paradigm/openmldb:0.2.1

## Detailed Description
After a further investigation, i find that the SQL Parser already support parsing multiple rows(as code shown below).
``` c++
// hybridse/src/planv2/ast_node_converter.cc
base::Status ConvertInsertStatement(const zetasql::ASTInsertStatement* root, node::NodeManager* node_manager,
    ...
    node::ExprListNode* rows = node_manager->MakeExprList();
    for (auto row : root->rows()->rows()) {
        CHECK_TRUE(nullptr != row, common::kSqlError, "Un-support insert statement with null row")
        node::ExprListNode* row_values;
        CHECK_STATUS(ConvertExprNodeList(row->values(), node_manager, &row_values))
        for (auto expr : row_values->children_) {
            CHECK_TRUE(nullptr != expr &&
                           (node::kExprPrimary == expr->GetExprType() || node::kExprParameter == expr->GetExprType()),
                       common::kSqlError, "Un-support insert statement with un-const value")
        }
        rows->AddChild(row_values);
    }
    ...
```
However, the SQLClusterRouter only take the first row put build a ```default_map```.
```c++
// src/sdk/sql_cluster_router.cc
SQLClusterRouter::GetInsertInfo(const std::string& db, const std::string& sql, ::hybridse::sdk::Status* status,
    ...
    *default_map = GetDefaultMap(*table_info, column_map,
                                 dynamic_cast<::hybridse::node::ExprListNode*>(insert_stmt->values_[0]), str_length);
    if (!(*default_map)) {
        status->msg = "get default value map of " + sql + " failed";
        LOG(WARNING) << status->msg;
        return false;
    }
    ...
```
I am not family with C++, so i am not sure if the bug related to the code i find above.

## Possible Implementation
I think it is not hard to fix ```SQLClusterRouter::GetInsertInfo```, however when coming to error handling, the behavior after a row failed to be inserted should be discussed before fix the bug.

jingchen2222:

feature

#398 #391 

* **Other information**:
Further things to do:
  - API Server support multi-rows insert statement @vagetablechicken 
  - Integration Test @wuyou10206

### DIFF
--- a/src/sdk/sql_cluster_router.cc
+++ b/src/sdk/sql_cluster_router.cc
@@ -261,7 +261,86 @@ std::shared_ptr<SQLInsertRow> SQLClusterRouter::GetInsertRow(const std::string& 
     SetCache(db, sql, cache);
     return std::make_shared<SQLInsertRow>(table_info, cache->column_schema, default_map, str_length);
 }
-
+bool SQLClusterRouter::GetMultiRowInsertInfo(const std::string& db, const std::string& sql,
+                                             ::hybridse::sdk::Status* status,
+                                             std::shared_ptr<::openmldb::nameserver::TableInfo>* table_info,
+                                             std::vector<DefaultValueMap>* default_maps,
+                                             std::vector<uint32_t>* str_lengths) {
+    if (status == NULL || table_info == NULL || default_maps == NULL || str_lengths == NULL) {
+        status->msg = "insert info is null";
+        LOG(WARNING) << status->msg;
+        return false;
+    }
+    ::hybridse::node::NodeManager nm;
+    ::hybridse::plan::PlanNodeList plans;
+    bool ok = GetSQLPlan(sql, &nm, &plans);
+    if (!ok || plans.size() == 0) {
+        LOG(WARNING) << "fail to get sql plan with sql " << sql;
+        status->msg = "fail to get sql plan with";
+        return false;
+    }
+    ::hybridse::node::PlanNode* plan = plans[0];
+    if (plan->GetType() != hybridse::node::kPlanTypeInsert) {
+        status->msg = "invalid sql node expect insert";
+        LOG(WARNING) << "invalid sql node expect insert";
+        return false;
+    }
+    ::hybridse::node::InsertPlanNode* iplan = dynamic_cast<::hybridse::node::InsertPlanNode*>(plan);
+    const ::hybridse::node::InsertStmt* insert_stmt = iplan->GetInsertNode();
+    if (insert_stmt == NULL) {
+        LOG(WARNING) << "insert stmt is null";
+        status->msg = "insert stmt is null";
+        return false;
+    }
+    *table_info = cluster_sdk_->GetTableInfo(db, insert_stmt->table_name_);
+    if (!(*table_info)) {
+        status->msg = "table with name " + insert_stmt->table_name_ + " in db " + db + " does not exist";
+        LOG(WARNING) << status->msg;
+        return false;
+    }
+    std::map<uint32_t, uint32_t> column_map;
+    for (size_t j = 0; j < insert_stmt->columns_.size(); ++j) {
+        const std::string& col_name = insert_stmt->columns_[j];
+        bool find_flag = false;
+        for (int i = 0; i < (*table_info)->column_desc_size(); ++i) {
+            if (col_name == (*table_info)->column_desc(i).name()) {
+                if (column_map.count(i) > 0) {
+                    status->msg = "duplicate column of " + col_name;
+                    LOG(WARNING) << status->msg;
+                    return false;
+                }
+                column_map.insert(std::make_pair(i, j));
+                find_flag = true;
+                break;
+            }
+        }
+        if (!find_flag) {
+            status->msg = "can't find column " + col_name + " in table " + (*table_info)->name();
+            LOG(WARNING) << status->msg;
+            return false;
+        }
+    }
+    for (int i = 0; i < insert_stmt->values_.size(); i++) {
+        hybridse::node::ExprNode* value = insert_stmt->values_[i];
+        if (value->GetExprType() != ::hybridse::node::kExprList) {
+            status->msg =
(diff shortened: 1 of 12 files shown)


src/sdk/sql_cluster_router.cc
```diff
@@ -261,7 +261,86 @@ std::shared_ptr<SQLInsertRow> SQLClusterRouter::GetInsertRow(const std::string&
...
+                "invalid row expression, expect kExprList but " + hybridse::node::ExprTypeName(value->GetExprType());
+            LOG(WARNING) << status->msg;
+            return false;
+        }
+        uint32_t str_length = 0;
+        default_maps->push_back(GetDefaultMap(*table_info, column_map,
+                                              dynamic_cast<::hybridse::node::ExprListNode*>(value),
+                                              &str_length));
+        if (!default_maps->back()) {
+            status->msg = "get default value map of row[" + std::to_string(i) + "] failed";
```
Shouren: I think It would be better to contain the index info of row in the error message - e.g. "get default value map of row [xxx] ({index}/{total}) failed", but not sure the current format is good enough.

jingchen2222: Good idea! I will fix it. Thanks.

src/sdk/sql_cluster_router.cc
```diff
@@ -865,38 +944,58 @@ std::shared_ptr<hybridse::sdk::ResultSet> SQLClusterRouter::ExecuteSQLBatchReque
...
+
+    if (default_maps.empty() || str_lengths.empty() || default_maps.size() != str_lengths.size()) {
+        status->msg = "default maps is empty or str_lengths are empty or default_maps";
+        status->code = 1;
         LOG(WARNING) << status->msg;
         return false;
     }
-    if (!row->IsComplete()) {
-        status->msg = "insert value isn't complete";
+    if (default_maps.size() != str_lengths.size()) {
```
Shouren: Perhaps this check should be put inside the ```GetMultiRowInsertInfo``` function, since they both come from that funciton.

src/sdk/sql_cluster_router.cc
```diff
@@ -865,38 +944,58 @@ std::shared_ptr<hybridse::sdk::ResultSet> SQLClusterRouter::ExecuteSQLBatchReque
...
+            LOG(WARNING) << status->msg;
+            return false;
+        }
+        if (!row->IsComplete()) {
+            status->msg = "insert value isn't complete";
+            LOG(WARNING) << status->msg;
+            return false;
+        }
+        std::vector<std::shared_ptr<::openmldb::catalog::TabletAccessor>> tablets;
+        bool ret = cluster_sdk_->GetTablet(db, table_info->name(), &tablets);
```
Shouren: Perhaps ```GetTablet``` function should be outside the loop section, i don't know whether there is load-balance algorithm inside this function so we should call it before ```PutRow```.

Shouren (changes requested):
Having a strong concern on the [duplicate call](https://github.com/4paradigm/OpenMLDB/pull/399/files#r706080907) to ```GetTablelet``` if there is no other reason to avoid the change.

src/sdk/sql_cluster_router.cc
jingchen2222: Actually, there isn't a dynamic load-balance algorithm here. So we can simply put `GetTablet` out of the loop. Just a bad design. I will fix it.

jingchen2222:
> Having a strong concern on the [duplicate call](https://github.com/4paradigm/OpenMLDB/pull/399/files#r706080907) to `GetTablelet` if there is no other reason to avoid the change.

You're right. Fixed already. Pls check.

src/sdk/sql_cluster_router.cc
jingchen2222: Fixed

aceforeverd (approved):

imotai (approved):
LGTM

PR #414, opened 2021-09-13
feat: reconfiguration window skew optimization

zekai427:
- Delete some code implementation in window skew optimization with skew config
- Fix some unit tests

- In order to better understand the details of window skew optimization

> 1. Reverse order of column in window skew optimization's unionDf
> For example, “col3” is the time column. And “part” column is in the same order as the time column now.
> 
> > PREVIOUSLY：
> > wecom-temp-0719afcc896cbc9ec3860c31098ce11d
> > NOW：
> > wecom-temp-89ea01c49c83f17b848be7c9a14e67b8
> 
> 2. Rename "_EXPAND_" to "EXPANDED_ROW"
> 3. Rename "_PART_" to "PART_ID"
> 4. Rename "_PARTITION_" to "PARTITION_KEY"

- Now the type of  "EXPANDED_ROW" column is Boolean.The row which  "EXPANDED_ROW" is true means that it is expanded row.


tobegit3hub (approved):
LGTM

dl239 (approved):

PR #436, opened 2021-09-17
fix: fix count in some yaml cases

zekai427:
- Some cases use 'count' with window calculation, but don't add over window behind 'count'.
- Now, fix these cases.


tobegit3hub (approved):
LGTM

jingchen2222 (approved):
LGTM

PR #438, opened 2021-09-22
fix: fix a bug when turn on window parallelization and skew optimization 

### LINKED ISSUE #444: Fix enable WindowSkewOpt and WindowComputeParallelOpt at the same time
Now we can enable WindowSkewOpt or WindowComputeParallelOpt to accelerate the batch jobs. But enable them at the same time may trigger some bugs which need to be fixed.

zekai427:
- When open window parallelization and skew optimization, some cases will fail.
- Now, fix it.
- Resolve https://github.com/4paradigm/OpenMLDB/issues/444 .


tobegit3hub (approved):
LGTM

dl239 (approved):

PR #464, opened 2021-09-26
feat: read openmldb git properties and set in return string

### LINKED ISSUE #463: Return the openmldb version string for Batch
Now the openmldb version is hard-coded to print in STDOUT and we should get the string version and return.

The OpenMLDB Spark distribution is affected by this API and do not need to get Spark version again.

tobegit3hub:
* Resolve https://github.com/4paradigm/OpenMLDB/issues/463 .


zekai427 (approved):
LGTM

superzhaozheng (approved):
LGTM

PR #490, opened 2021-09-30
fix: add switch to open broadcast join when open window skew optimization

zekai427:
- Fix dataframe cache problem


java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/nodes/WindowAggPlan.scala
```diff
@@ -196,13 +192,18 @@ object WindowAggPlan {
       -1
     }
 
+    if (ctx.getConf.windowSkewOptCache) {
+      distributionDf = distributionDf.cache()
```
tobegit3hub: Do not need to return and set the `distributionDf` object.

java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/nodes/WindowAggPlan.scala
```diff
@@ -196,13 +192,18 @@ object WindowAggPlan {
...
+    }
+
     // 2. Add "part" column and "expand" column by joining the distribution table
-    val addColumnsDf = SkewDataFrameUtils.genAddColumnsDf(inputDf, distributionDf, quantile.intValue(),
+    var addColumnsDf = SkewDataFrameUtils.genAddColumnsDf(inputDf, distributionDf, quantile.intValue(),
```
tobegit3hub: Do not use `var` and set the `addColumnsDf` after calling `cache()`.

tobegit3hub (changes requested):
This does not fix the issue of `broadcast`, we can link to another issue like unpersisting cache dataframes.

java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/utils/SkewDataFrameUtils.scala
```diff
@@ -66,14 +66,20 @@ object SkewDataFrameUtils {
 
   def genAddColumnsDf(inputDf: DataFrame, distributionDf: DataFrame, quantile: Int,
                       repartitionColIndex: mutable.ArrayBuffer[Int], orderByColIndex: Int,
-                      partitionKeyColName: String, partIdColName: String, expandedRowColName: String): DataFrame = {
+                      partitionKeyColName: String, partIdColName: String, expandedRowColName: String,
+                      openBroadcastJoin: Boolean): DataFrame = {
```
tobegit3hub, line 70: You can pass the default value to set `openBroadcastJoin`.

tobegit3hub (approved):
LGTM

dl239 (approved):

PR #497, opened 2021-10-09
feat: add a config for print the physical plan

zekai427:
- We don't need to print physical plan each time when we execute plan.
- Now, if we want to get the physical plan, we should `set("openmldb.physical.plan.print", true)`.


jingchen2222 (approved):
LGTM

tobegit3hub (approved):
LGTM

PR #549, opened 2021-10-20
feat: fix the base dir of yaml cases

tobegit3hub:
Fix the base dir of yaml cases which should be `openmldb` instead of `rtidb`.

### DIFF
--- a/src/test/base_test.cc
+++ b/src/test/base_test.cc
@@ -39,7 +39,7 @@ std::string SQLCaseTest::GetYAMLBaseDir() {
             yaml_base_dir.append("/");
         }
     } else {
-        yaml_base_dir = "/rtidb";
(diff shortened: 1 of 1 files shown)


aceforeverd (approved):

jingchen2222 (approved):
LGTM

src/test/base_test.cc
```diff
@@ -39,7 +39,7 @@ std::string SQLCaseTest::GetYAMLBaseDir() {
             yaml_base_dir.append("/");
         }
     } else {
-        yaml_base_dir = "/rtidb";
+        yaml_base_dir = "/openmldb";
```
jingchen2222, line 42: Maybe we can simply remove the default the `yaml base dir`

superzhaozheng (approved):
LGTM

PR #554, opened 2021-10-20
feat: client supports `load data infile`

### LINKED ISSUE #443: SQL: support `LOAD DATA INFILE` Statement
~~desc:  batch import data into OpenMLDB~~
~~format: import table  file_path~~
~~ex: `import demo_table1 file:///OpenMLDB/cli/examples/taxi_tour_parquet_all/raw_data.csv;`~~

Desc: support `LOAD DATA INFILE 'file_name'` SQL Statement like MySQL
E.g: `LOAD DATA INFILE 'hello.csv' INTO TABLE t1;`

vagetablechicken:
Close #443 
Client supports to load local file to openmldb.
Only csv format now, can specify delimiter and nullValue, whether the file has header.
If has header, we will check if equal with table schema. The order of columns should be the same with table schema.


src/cmd/sql_cmd.h
```diff
@@ -532,6 +533,221 @@ void HandleCreateIndex(const hybridse::node::CreateIndexNode *create_index_node)
     }
 }
 
+bool SetOption(const std::shared_ptr<hybridse::node::OptionsMap> &options, const std::string &option_name,
+               hybridse::node::DataType option_type,
+               std::function<bool(const hybridse::node::ConstNode *node)> const &f) {
```
dl239, line 552: GetOption?

src/cmd/sql_cmd.h
```diff
@@ -532,6 +533,221 @@ void HandleCreateIndex(const hybridse::node::CreateIndexNode *create_index_node)
...
+    // scan all strings , calc the sum, to init SQLInsertRow's string length
+    std::string::size_type str_len_sum = 0;
+    for (int i = 0; i < cnt; ++i) {
+        if (schema->GetColumnType(i) == hybridse::sdk::kTypeString && cols[i] != nullValue) {
+            str_len_sum += cols[i].length();
+        }
+    }
+    row->Init(static_cast<int>(str_len_sum));
+
+    for (int i = 0; i < cnt; ++i) {
```
dl239, line 662: get and save string position from schema before write to avoid traverse all field when write every record

src/cmd/sql_cmd.h
vagetablechicken: ok, Get is better.

src/cmd/sql_cmd.h
vagetablechicken: ok

src/cmd/sql_cmd.h
```diff
@@ -532,6 +533,230 @@ void HandleCreateIndex(const hybridse::node::CreateIndexNode *create_index_node)
...
+        return false;
+    }
+    if (!f(node)) {
+        std::cout << "parse option " << option_name << " failed" << std::endl;
+        return false;
+    }
+    return true;
+}
+
+template <typename T>
```
dl239, line 571: why set row as template

vagetablechicken: For both InsertRow/RequestRow, referenced AppendJsonValue in api server.

src/cmd/split.h
```diff
@@ -0,0 +1,159 @@
...
+//     #include "util/csv/parser.h"
+//     for (Record rec : Parser(source)) {
+//       vector<string> fields = rec.fields();
+//     }
+//
+// See //util/csv/parser.h for more complete documentation.
+//
+// ----------------------------------------------------------------------
+void SplitCSVLineWithDelimiter(char* line, char delimiter, std::vector<char*>* cols) {
+    char* end_of_line = line + strlen(line);
```
dl239, line 93: Pass the string length with function arguments

src/cmd/split.h
```diff
@@ -0,0 +1,159 @@
...
+        assert(*line == '\0' || *line == delimiter);
+    }
+}
+
+void SplitCSVLine(char* line, std::vector<char*>* cols) { SplitCSVLineWithDelimiter(line, ',', cols); }
+
+void SplitCSVLineWithDelimiterForStrings(const std::string& line, char delimiter, std::vector<std::string>* cols) {
+    // Unfortunately, the interface requires char* instead of const char*
+    // which requires copying the string.
+    char* cline = strndup_with_new(line.c_str(), line.size());
```
dl239, line 144: need not copy raw string here if use slice/string vec as output argument

src/cmd/split.h
vagetablechicken: Maybe we can add a new version SplitCSVLineWithDelimiter for std::string later, not just get char* from the std::string

src/cmd/split.h
vagetablechicken: same solution https://github.com/4paradigm/OpenMLDB/pull/554#discussion_r733466001

dl239 (approved):

src/cmd/sql_cmd.h
```diff
@@ -532,6 +533,237 @@ void HandleCreateIndex(const hybridse::node::CreateIndexNode *create_index_node)
...
+
+// Only csv format
+bool HandleLoadDataInfile(const std::string &database, const std::string &table, const std::string &file_path,
+                          const std::shared_ptr<hybridse::node::OptionsMap> &options, std::string *error) {
+    DCHECK(error);
+
+    // options, value is ConstNode
+    char delimiter = ',';
+    bool header = true;
+    std::string nullValue{"null"}, format{"csv"};
```
vagetablechicken: null_value

aceforeverd (approved):

PR #557, opened 2021-10-20
feat: add procedure type

### LINKED ISSUE #556: Add procedure type to `ProcedureInfo`
Run `Deploy` cmd will create a procdure. So we need to distinguish which  procdure is deployment service

dl239:
Run Deploy cmd will create a procdure. So we need to distinguish which procdure is deployment service

### DIFF
--- a/hybridse/include/sdk/base.h
+++ b/hybridse/include/sdk/base.h
@@ -129,6 +129,12 @@ class TableSet {
     virtual int32_t Size() { return 0; }
 };
 
+enum ProcedureType {
+    kUnknow = -1,
+    kNormal = 0,
+    kDeployment,
+};
+
 class ProcedureInfo {
  public:
     ProcedureInfo() {}
@@ -141,6 +147,7 @@ class ProcedureInfo {
     virtual const std::vector<std::string>& GetTables() const = 0;
     virtual const std::string& GetMainTable() const = 0;
     virtual const std::string& GetMainDb() const = 0;
+    virtual ProcedureType GetType() const = 0;
 };
 
 }  // namespace sdk
--- a/hybridse/include/sdk/base.h
+++ b/hybridse/include/sdk/base.h
@@ -131,8 +131,8 @@ class TableSet {
 
 enum ProcedureType {
     kUnknow = -1,
-    kNormal = 0,
-    kDeployment,
+    kReqProcedure = 0,
+    kReqDeployment,
 };
 
 class ProcedureInfo {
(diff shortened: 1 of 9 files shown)


hybridse/include/sdk/base.h
```diff
@@ -129,6 +129,12 @@ class TableSet {
     virtual int32_t Size() { return 0; }
 };
 
+enum ProcedureType {
+    kUnknow = -1,
+    kNormal = 0,
+    kDeployment,
```
jingchen2222: Actually, the current Procedure isn't a normal one. It is designed in a special way only to work for `Request` and `BatchRequest` query.
Can we use type like kRequestSp for the current Procedure type. 
And we can add kNormal Procedure in the future which will process procedure in a normal way.

jingchen2222: What is the difference between creating a procedure with deploying a procedure?

steps/init_env.sh
```diff
@@ -73,6 +72,10 @@ if [[ ${HYBRIDSE_SOURCE} = "local" ]]; then
     popd
 
 else
+    if [ -d "$THIRDPARTY_PATH/hybridse" ]; then
+        echo "${GREEN}thirdparty/hybridse path: $THIRDPARTY_PATH/hybridse already exist, skip download/install deps${NC}"
+        exit 0
+    fi
```
jingchen2222, line 78: If so, we had better check hybridse version before.

dl239: yes, version check can doing in another issue/pr.

hybridse/include/sdk/base.h
dl239: we need to distinguish which procdure is a deployment service. If not doing this, we need store deployment like procedure

vagetablechicken (approved):
LGTM

hybridse/include/sdk/base.h
```diff
@@ -129,6 +129,12 @@ class TableSet {
     virtual int32_t Size() { return 0; }
 };
 
+enum ProcedureType {
```
tobegit3hub, line 132: Can we reuse the ProcedureType from protobuf?

aceforeverd (approved):

tobegit3hub (approved):
LGTM

hybridse/include/sdk/base.h
tobegit3hub: We may reuse the similar type definition in proto and hybridse later.

dl239: hybridse cannot use proto defined in src/proto

PR #593, opened 2021-10-26
fix: ttl set in IndexMap

vagetablechicken:
`IndexMapBuilder::ToMap()` forgot to set the ttl in ColumnKey.


dl239 (approved):

src/base/ddl_parser.cc
```diff
@@ -96,6 +92,8 @@ IndexMap IndexMapBuilder::ToMap() {
     IndexMap result;
     for (auto& pair : index_map_) {
         auto dec = Decode(pair.first);
+        // message owns the memory
+        dec.second.set_allocated_ttl(pair.second);
```
aceforeverd, line 96: clear vector content

vagetablechicken: done

tobegit3hub (approved):
LGTM

PR #596, opened 2021-10-27
feat: remove check db for some api in sql cmd

### LINKED ISSUE #591: SQL client support querying and other APIs without setting database
Now we have support multiple databases and can use SQL without setting database in advance. But `use $db` is required for command-line SQL client. We should remove this restriction by passing default database name.

tobegit3hub:
* Remove check if db is empty for some APIs
* Print status message when fail to run sql
* Resolve https://github.com/4paradigm/OpenMLDB/issues/591


zekai427 (approved):

jingchen2222 (approved):

tobegit3hub:
[CLI for offline jobs with unified metadata store](https://trello.com/c/rcSLPVcy/21-cli-for-offline-jobs-with-unified-metadata-store)

PR #598, opened 2021-10-27
feat: refactor option class

zekai427:

* **Other information**:


src/cmd/sql_cmd.h
```diff
@@ -72,148 +72,145 @@ ::openmldb::sdk::DBSDK *cs = nullptr;
...
-// TODO(zekai): use getoption
-class SaveFileOptions {
+class OptionsParse {
  public:
-    SaveFileOptions(const std::string &file_path, std::shared_ptr<hybridse::node::OptionsMap> options_map,
-        ::openmldb::base::ResultMsg* openmldb_base_status) {
-        // TODO(zekai): Resolved file path like (file:////usr/test.csv) or (hdfs:////usr/test.csv)
-        file_path_ = file_path;
-        for (auto iter = options_map->begin(); iter != options_map->end(); iter++) {
+    OptionsParse(std::shared_ptr<hybridse::node::OptionsMap> options_map, ::openmldb::base::ResultMsg *status) {
```
vagetablechicken: It's not good to do this long parsing in ctor.
it's better to do like this:
```
OptionsParser pasrer;
Status st = parser.parse(optionsMap);
```
Further more, you can OptionsParser parser(READ/WRITE MODE), or parser.setRead/serWrite to load different paser methods.
No need to use if(key==..) blabla, just store a methods map, method = map.find(key),  then call method.

src/cmd/sql_cmd.h
```diff
@@ -72,148 +72,145 @@ ::openmldb::sdk::DBSDK *cs = nullptr;
...
 void SaveResultSet(::hybridse::sdk::ResultSet *result_set, const std::string &file_path,
-    std::shared_ptr<hybridse::node::OptionsMap> options_map, ::openmldb::base::ResultMsg* openmldb_base_status) {
-    std::shared_ptr<openmldb::cmd::SaveFileOptions> options = std::make_shared<openmldb::cmd::SaveFileOptions>(
-        file_path, options_map, openmldb_base_status);
-    if (!result_set || !openmldb_base_status->OK()) {
+                   std::shared_ptr<hybridse::node::OptionsMap> options_map, ::openmldb::base::ResultMsg *status) {
+    if (!result_set) {
+        return;
+    }
+    std::shared_ptr<openmldb::cmd::OptionsParse> options_parse =
```
vagetablechicken: Why use shared ptr, local variable is not ok?

src/cmd/sql_cmd.h
```diff
@@ -72,148 +72,175 @@ ::openmldb::sdk::DBSDK *cs = nullptr;
...
+                status->msg = "ERROR: This option " + key + " is not currently supported";
+                status->code = openmldb::base::kSQLCmdRunError;
             }
-        }
-        // Check mode
-        if (mode_ == "error_if_exists") {
-            if (access(file_path_.c_str(), 0) == 0) {
-                openmldb_base_status->msg = "ERROR: File already exists";
-                openmldb_base_status->code = openmldb::base::kSQLCmdRunError;
+            if (!status->OK()) {
```
vagetablechicken: why not return immediately after line 124?

vagetablechicken: GetOption may failed too.

src/cmd/sql_cmd.h
```diff
@@ -72,148 +72,174 @@ ::openmldb::sdk::DBSDK *cs = nullptr;
...
-            if (access(file_path_.c_str(), 0) == 0) {
-                openmldb_base_status->msg = "ERROR: File already exists";
-                openmldb_base_status->code = openmldb::base::kSQLCmdRunError;
-                return;
-            } else {
-                fstream_.open(file_path_);
+            auto pair = check_map_.find(key);
+            if (pair == check_map_.end()) {
+                status_.msg = "ERROR: This option " + key + " is not currently supported";
+                status_.code = openmldb::base::kSQLCmdRunError;
```
vagetablechicken, line 95: If not found, why not return or break?

src/cmd/sql_cmd.h
```diff
@@ -72,148 +72,174 @@ ::openmldb::sdk::DBSDK *cs = nullptr;
...
-        } else if (mode_ == "overwrite") {
-            fstream_.open(file_path_, std::ios::out);
-        } else if (mode_ == "append") {
-            fstream_.open(file_path_, std::ios::app);
-            fstream_ << std::endl;
-        } else {
-            openmldb_base_status->msg = "ERROR: This mode (" + mode_ + ") is not currently supported";
-            openmldb_base_status->code = openmldb::base::kSQLCmdRunError;
-            return;
+            GetOption(iter->second, key, pair->second.first, pair->second.second);
```
vagetablechicken: `GetOption` should return bool, so it's easy to break the parse process by `if(!GetOption)`

src/cmd/sql_cmd.h
```diff
@@ -72,148 +72,174 @@ ::openmldb::sdk::DBSDK *cs = nullptr;
...
         }
-        // Check file
-        if (fstream_.is_open() == false) {
-            openmldb_base_status->msg = "ERROR: Fail to open file, please check file path";
-            openmldb_base_status->code = openmldb::base::kSQLCmdRunError;
+        return &status_;
```
vagetablechicken: It's not a good idea to return the ptr of status_, what if I delete parser then use this status ptr? Use Object

src/cmd/sql_cmd.h
```diff
@@ -72,148 +72,174 @@ ::openmldb::sdk::DBSDK *cs = nullptr;
...
+    }
+    void InsertCheckMap(
```
vagetablechicken: AddOption is better?

src/cmd/sql_cmd.h
```diff
@@ -72,148 +72,174 @@ ::openmldb::sdk::DBSDK *cs = nullptr;
...
+// TODO(zekai): refactor status
+class FileOptionsParser {
  public:
-    SaveFileOptions(const std::string &file_path, std::shared_ptr<hybridse::node::OptionsMap> options_map,
-        ::openmldb::base::ResultMsg* openmldb_base_status) {
-        // TODO(zekai): Resolved file path like (file:////usr/test.csv) or (hdfs:////usr/test.csv)
-        file_path_ = file_path;
-        for (auto iter = options_map->begin(); iter != options_map->end(); iter++) {
+    FileOptionsParser() {
+        check_map_.insert(std::make_pair(
```
vagetablechicken: how about use the method to insert? avoid too many make_pair

src/cmd/sql_cmd.h
```diff
@@ -76,148 +76,184 @@ ::openmldb::sdk::DBSDK *cs = nullptr;
...
-    }
-    std::string GetFilePath() {
-        return file_path_;
-    }
-    bool GetHeader() const {
-        return header_;
-    }
-    std::ofstream& GetOfstream() {
-        return fstream_;
+    std::pair<std::function<bool(const hybridse::node::ConstNode *node)>, hybridse::node::DataType> CheckFormat() {
```
tobegit3hub: This function is not only return the actual check function but also return the required type. The function name is not proper. You may return the actual check function here and `make_pair` when using.

Then the unit test for functions like`CheckFomart` would be easier to implement.

src/cmd/sql_cmd.h
```diff
@@ -76,148 +76,184 @@ ::openmldb::sdk::DBSDK *cs = nullptr;
...
+            status->msg = "ERROR: File already exists";
+            status->code = openmldb::base::kSQLCmdRunError;
+            return;
+        } else {
+            fstream.open(file_path);
+        }
+    } else if (options_parse.GetMode() == "overwrite") {
+        fstream.open(file_path, std::ios::out);
+    } else if (options_parse.GetMode() == "append") {
+        fstream.open(file_path, std::ios::app);
```
tobegit3hub, line 221: Add warning when using `mode=append` and `header=true` which may append the header in the middle of output file.

tobegit3hub (changes requested):

vagetablechicken (approved):
LGTM

tobegit3hub (approved):
LGTM

PR #610, opened 2021-10-28
fix: filter condition expression fix

### LINKED ISSUE #599: Bug: where statement return expected result when

When we query where statement, the result is unexpected. The filters condition doesn't work.

## Expected Behavior

Expect result rows filtered by `c2=2`;
## Current Behavior

wecom-temp-75948a2c673bc8beae6ada26a61de7ae

## Possible Solution

There might maybe something wrong with the online runner. We should:
1. Check the FilterPhysicalOp and the FilterRunner and figure out the problem.
2. Add related unit case

## Steps to Reproduce

wecom-temp-75948a2c673bc8beae6ada26a61de7ae

jingchen2222:

Bug fix

#599 

1. rebuild condition expression after key optimization

* **Other information**:

### DIFF
--- a/cases/function/select/test_where.yaml
+++ b/cases/function/select/test_where.yaml
@@ -39,8 +39,8 @@ cases:
       data: |
         0, 1, 5, 1.1, 11.1, 1, 1
         0, 2, 5, 2.2, 22.2, 2, 22
-  - id: 1
-    desc: Where部分条件命中索引
+  - id: 1-1
+    desc: Where部分条件命中索引, col1<3条件未命中
     mode: request-unsupport, offline-unsupport
     db: db1
     sql: |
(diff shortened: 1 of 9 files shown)


jingchen2222:
## Problem & solution

### Problem
We found a bug when querying where query statements in some specific situation. 

Before we dive into the problem and reason, I have to briefly introduce two physical optimization passes here: `kPassFilterOptimized` and `kPassGroupAndSortOptimized`.

##### kPassFilterOptimized
The pass `kPassFilterOptimized:` split the filter condition expression into two parts. One part is left/right keys extracted from the original condition. The other part is the rest condition.

Take condition `col1 = 10 and col2 = 20 and col3 < 30` as an **example**. 
After `kPassFilterOptimized`, the original condition can be split into:
- the left keys: `(10, 20)`, the right keys: `(col1, col2)` 
- the rest condition:`(col3 < 30)`.

###### kPassGroupAndSortOptimized
The pass `kPassGroupAndSortOptimized` is responsible to optimize the generated left_keys and right_keys with the given table's indexes. 
So, after applying for this pass, we can still keep some left_keys and right_keys if some keys **unmatch** with the index.

In OpenMLDB-batch, the executor makes use of left keys and right keys and performs filtering.

However, when it comes to OpenMLDB-online, the executor does not deal with left keys and right keys. That's why we can get unexpected results when querying where statement.


### Solution
1. modify kPassGroupAndSortOptimized, rebuild condition after keys optimization. new_condition = condition_expr and (left_key1 = right_key1) and (left_key2 = right_key2) and ....

dl239 (approved):

tobegit3hub (approved):
LGTM

PR #629, opened 2021-11-02
feat: support capital letters for bool like 'FALSE'

zekai427:

* **Other information**:

### DIFF
--- a/hybridse/src/planv2/ast_node_converter_test.cc
+++ b/hybridse/src/planv2/ast_node_converter_test.cc
@@ -185,7 +185,7 @@ TEST_F(ASTNodeConverterTest, InvalidASTIntervalLiteralTest) {
         ASSERT_EQ("Invalid interval literal: 1abds", status.msg);
     }
 }
-TEST_F(ASTNodeConverterTest, InvalidASTBollLiteralTest) {
+TEST_F(ASTNodeConverterTest, InvalidASTBoolLiteralTest) {
     node::NodeManager node_manager;
     {
         zetasql::ASTBooleanLiteral expression;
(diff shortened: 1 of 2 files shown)


dl239:
It's better to get the real value from parser instead of `bool` @jingchen2222

jingchen2222:
> It's better to get the real value from parser instead of `bool` @jingchen2222

@dl239 What does real value mean? 

After parsing this command, the parser will return a `SetPlanNode` with a `ConstNode` of bool type.
But we fail to convert some zetasql's literal to hybridse's ConstNode.

I think this pr can fix the well. It would be better if we can add some related unit tests. @Kanekanekane

zekai427:
> > It's better to get the real value from parser instead of `bool` @jingchen2222
> 
> @dl239 What does real value mean?
> 
> After parsing this command, the parser will return a `SetPlanNode` with a `ConstNode` of bool type. But we fail to convert some zetasql's literal to hybridse's ConstNode.
> 
> I think this pr can fix the well. It would be better if we can add some related unit tests. @Kanekanekane

Done.

dl239:
> > It's better to get the real value from parser instead of `bool` @jingchen2222
> 
> @dl239 What does real value mean?
> 
> After parsing this command, the parser will return a `SetPlanNode` with a `ConstNode` of bool type. But we fail to convert some zetasql's literal to hybridse's ConstNode.
> 
> I think this pr can fix the well. It would be better if we can add some related unit tests. @Kanekanekane

`set  key value` statement, parser need not concern the value is bool or not? User get string value and decide how to use

tobegit3hub (approved):
LGTM

dl239 (approved):

PR #630, opened 2021-11-02
fix: fix convert ttl

dl239:
* convert ttl from millisecond to minute
* disable cluster_optimized with standalone mode
* don't send load index request when there is no index to add


jingchen2222 (approved):
LGTM

vagetablechicken (approved):

PR #638, opened 2021-11-03
fix: tid in nameserver and show deployment print

### LINKED ISSUE #635: bug: tid in create table
When create table, the id of table(tid) is set by nameserver, and nameserver use `table_index_` to store the last tid.
But `table_index_` is uninitialized, and it's uint32.
Then we use this uint32 tid to `CreateTableOnTablet`.
https://github.com/4paradigm/OpenMLDB/blob/eaf73ad4541e4897d15e55d17b31dbe1dd5356be/src/nameserver/name_server_impl.cc#L2347
The tid in `::openmldb::api::TableMeta` is int32.
So if the random tid is too large, we will set a negative id to `::openmldb::api::TableMeta`. The tablet server's check will get `tid is zero`. Actually it's a bad msg too.
https://github.com/4paradigm/OpenMLDB/blob/eaf73ad4541e4897d15e55d17b31dbe1dd5356be/src/tablet/tablet_impl.cc#L684-L687

vagetablechicken:
Closes  #635

### DIFF
--- a/src/cmd/openmldb.cc
+++ b/src/cmd/openmldb.cc
@@ -326,7 +326,7 @@ int PutData(uint32_t tid, const std::map<uint32_t, std::vector<std::pair<std::st
         }
         if (ts_dimensions.empty()) {
             if (!clients[endpoint]->Put(tid, pid, ts, value, iter->second, format_version)) {
-                printf("put failed. tid %u pid %u endpoint %s ts %lu \n", tid, pid, endpoint.c_str(), ts);
+                printf("put failed. tid %u pid %u endpoint %s ts %llu \n", tid, pid, endpoint.c_str(), ts);
                 return -1;
             }
         } else {
--- a/src/cmd/openmldb.cc
+++ b/src/cmd/openmldb.cc
@@ -326,7 +326,7 @@ int PutData(uint32_t tid, const std::map<uint32_t, std::vector<std::pair<std::st
         }
         if (ts_dimensions.empty()) {
             if (!clients[endpoint]->Put(tid, pid, ts, value, iter->second, format_version)) {
-                printf("put failed. tid %u pid %u endpoint %s ts %llu \n", tid, pid, endpoint.c_str(), ts);
+                printf("put failed. tid %u pid %u endpoint %s ts %lu \n", tid, pid, endpoint.c_str(), ts);
                 return -1;
             }
         } else {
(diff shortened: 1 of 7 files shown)


zekai427 (approved):

src/cmd/openmldb.cc
```diff
@@ -326,7 +326,7 @@ int PutData(uint32_t tid, const std::map<uint32_t, std::vector<std::pair<std::st
         }
         if (ts_dimensions.empty()) {
             if (!clients[endpoint]->Put(tid, pid, ts, value, iter->second, format_version)) {
-                printf("put failed. tid %u pid %u endpoint %s ts %lu \n", tid, pid, endpoint.c_str(), ts);
+                printf("put failed. tid %u pid %u endpoint %s ts %llu \n", tid, pid, endpoint.c_str(), ts);
```
dl239: `llu` will generate compiling warning on linux

vagetablechicken: `lu` warn on mac. I'll revert this, but we may need a elegant solution about this warning later.

src/nameserver/name_server_impl.cc
```diff
@@ -2335,17 +2326,18 @@ int NameServerImpl::SetPartitionInfo(TableInfo& table_info) {
...
     ::openmldb::type::CompressType compress_type = ::openmldb::type::CompressType::kNoCompress;
     if (table_info->compress_type() == ::openmldb::type::kSnappy) {
         compress_type = ::openmldb::type::CompressType::kSnappy;
     }
     ::openmldb::api::TableMeta table_meta;
     table_meta.set_db(table_info->db());
     table_meta.set_name(table_info->name());
-    table_meta.set_tid(table_info->tid());
-    table_meta.set_seg_cnt(table_info->seg_cnt());
+    table_meta.set_tid(static_cast<::google::protobuf::int32>(table_info->tid()));
```
tobegit3hub, line 2339: I have seen tid is invalid(<=0) while printing the tid which is more than 0 in MacOS. Not sure if it is related to this.

tobegit3hub (approved):
LGTM

dl239 (approved):

src/cmd/openmldb.cc
dl239: `std::cout` may work well

src/nameserver/name_server_impl.cc
vagetablechicken: Probably yes, chenjing and zekai also met this error. Maybe error rate is high in MacOS.

PR #652, opened 2021-11-04
fix: ttl conversion and ddl parser bug

### LINKED ISSUE #623: bug: ttl convertion
1. sql to common::TTLSt
    
- [x] GetAbsTTL=0 means never gc
- [x] GetAbsTTL>0s(even just x ms, by set x without unit), we convert it to minutes, the smallest integer greater than the value.

1. DDLParser set common::TTLSt
    parsed sql may have 'window...rows_range between 0s preceding and ...', so `UpdateIndex` should support set abs ttl 0, actually it doesn't mean never gc, it means a short ttl, so we should convert it to 1min(the minimum ttl).
 and 'rows between 0 preceding....' should gc too.
 

- [x] but which means UNBOUNDED? check in `FrameExtent`. gethistory...start INT64_MIN means unbound.

- [x] fix this: https://github.com/4paradigm/OpenMLDB/blob/1b83bfc2763e7d5b1e0f2950263c17a0b3ca0ea2/src/base/ddl_parser.cc#L311
- [x] add ut

1. `desc t1;` shows the index info, if abs ttl 0, it's better to let user know that it means never gc.

vagetablechicken:
Closes #623 
`create table/index` to TTLSt: 0 means unbound
`extract index` to TTLSt: 0 means short ttl

And ddl parser convert ptr type fix, table check in load data.

### DIFF
--- a/src/base/ddl_parser.cc
+++ b/src/base/ddl_parser.cc
@@ -264,14 +264,6 @@ void DDLParser::AddTables(const T& schema, hybridse::type::Database* db) {
     }
 }
 
-uint64_t ConvertToMinute(uint64_t time_ms) {
-    if (time_ms == 0) {
-        // default Min minute is 1
-        return 1;
-    }
-    return time_ms / 60000 + (time_ms % 60000 ? 1 : 0);
-}
-
 bool IndexMapBuilder::CreateIndex(const std::string& table, const hybridse::node::ExprListNode* keys,
                                   const hybridse::node::OrderByNode* ts, const SchemasContext* ctx) {
     // we encode table, keys and ts to one string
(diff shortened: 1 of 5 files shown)


src/sdk/node_adapter.cc
```diff
@@ -224,6 +224,10 @@ bool NodeAdapter::TransformToColumnKey(hybridse::node::ColumnIndexNode* column_i
     if (column_index == nullptr) {
         return false;
     }
+
+    std::stringstream ss;
+    column_index->Print(ss, "");
+    DLOG(INFO) << ss.str();
```
dl239, line 230: call `Print `on debug mode.

src/base/ddl_parser.cc
```diff
@@ -289,11 +281,37 @@ bool IndexMapBuilder::CreateIndex(const std::string& table, const hybridse::node
...
 }
 
+int64_t AbsTTLConvert(int64_t time_ms, bool zero_eq_unbounded) {
+    if (zero_eq_unbounded && time_ms == 0) {
+        return 0;
+    }
+    return time_ms == 0 ? 1 : (time_ms / 60000 + (time_ms % 60000 ? 1 : 0));
+}
+
+int64_t LatTTLConvert(int64_t lat, bool zero_eq_unbounded) {
```
dl239, line 297: `if (zero_eq_unbounded) { return lat; }`

src/base/ddl_parser.cc
```diff
@@ -289,11 +281,37 @@ bool IndexMapBuilder::CreateIndex(const std::string& table, const hybridse::node
...
+    if (zero_eq_unbounded && lat == 0) {
+        return 0;
+    }
+
+    return lat == 0 ? 1 : lat;
+}
+
+// history_range_start == INT64_MIN: unbounded
```
dl239, line 305: When `history_range_start` is `INT64_MIN`

aceforeverd (approved):

src/sdk/node_adapter.cc
dl239: reslove later

dl239 (approved):

PR #659, opened 2021-11-05
refactor: delete client proto

### LINKED ISSUE #658: Delete `client.proto`
`client.proto` is unused currently

dl239:
* delete client proto
* add file_option_parser.h
* use GetAsString
* delete some unused code


src/cmd/sql_cmd.h
```diff
@@ -352,70 +170,7 @@ void PrintResultSet(std::ostream& stream, ::hybridse::sdk::ResultSet* result_set
...
-                    bool value = false;
-                    result_set->GetBool(i, &value);
-                    t.add(value ? "true" : "false");
-                    break;
-                }
-                default: {
-                    t.add("NA");
-                }
-            }
+            t.add(result_set->GetAsString(i));
```
dl239, line 173: bool true/false

tobegit3hub (approved):
LGTM

hybridse/include/sdk/result_set.h
```diff
@@ -79,7 +79,7 @@ class ResultSet {
                 return std::to_string(GetDoubleUnsafe(idx));
             }
             case kTypeBool: {
-                return std::to_string(GetBoolUnsafe(idx));
+                return GetBoolUnsafe(idx) ? "true" : "false";
```
dl239, line 82: @jingchen2222 there are any effects on hybridse if modify the bool value in `GetAsString`?

jingchen2222: `GetAsString` is used by the client. It's used only for viewing.
I think it's ok to modify the bool value's string format.

vagetablechicken (approved):

PR #664, opened 2021-11-09
refactor: define performance_sensitive in sql_cluster

zekai427:
- Now, it is not standard to define performance_sensitive in sql_cmd.h. And because of this realization, we should set additional options for setting performance_sensitive mode in openmldb JDBC.
- So, in order to solve these problems, I define performance_sensitive in sql_cluster and refactor some functions.
- Linked: https://github.com/4paradigm/OpenMLDB/issues/605

PR #685, opened 2021-11-12
fix: openmldb compilation

aceforeverd:
- fix `sql_standalone_sdk_test`,  rm `performance_sensitive` option, keep sync with #664

### DIFF
--- a/src/cmd/sql_cmd.h
+++ b/src/cmd/sql_cmd.h
@@ -818,8 +818,7 @@ base::Status HandleDeploy(const hybridse::node::DeployPlanNode* deploy_node) {
 }
 
 void SetVariable(const std::string& key, const hybridse::node::ConstNode* value) {
-    std::string lower_key = key;
-    boost::to_lower(lower_key);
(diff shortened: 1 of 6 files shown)


src/cmd/sql_cmd.h
```diff
@@ -818,8 +818,7 @@ base::Status HandleDeploy(const hybridse::node::DeployPlanNode* deploy_node) {
 }
 
 void SetVariable(const std::string& key, const hybridse::node::ConstNode* value) {
-    std::string lower_key = key;
-    boost::to_lower(lower_key);
+    auto lower_key = boost::to_lower_copy(key);
```
jingchen2222, line 821: `boost::to_lower_copy` is better?

jingchen2222 (approved):

src/cmd/sql_cmd.h
aceforeverd: Not a big difference.
Just thought two steps bit slower than the new one, intuitively.

tobegit3hub (approved):
LGTM

Issue #679, opened 2021-11-11
com._4paradigm.openmldb.sdk.Column has a field called sqlType, missing function to convert sqltype to openmldb.datatype


shawn-happy:
com._4paradigm.openmldb.sdk.Column has a field called sqlType, and provides setter method, maybe i will use `java.sql.Types`。but when i called `com._4paradigm.openmldb.DataType#swigToEnum` method, i get an error。
Because `java.sql.Types` is not `com._4paradigm.openmldb.DataType`。
i think maybe provides some convert method is better or provide some enum. 
for example:
```
public static DataType toDataType(int sqlType) throws SQLException {
        if (Types.BOOLEAN == sqlType) {
            return DataType.kTypeBool;
        } else if (Types.SMALLINT == sqlType) {
            return DataType.kTypeInt16;
        } else if (Types.INTEGER == sqlType) {
            return DataType.kTypeInt32;
        } else if (Types.BIGINT == sqlType) {
          return DataType.kTypeInt64;
        } else if (Types.FLOAT == sqlType) {
            return DataType.kTypeFloat;
        } else if (Types.DOUBLE == sqlType) {
            return DataType.kTypeDouble;
        } else if (Types.VARCHAR == sqlType) {
            return DataType.kTypeString;
        } else if (Types.DATE == sqlType) {
            return DataType.kTypeDate;
        } else if (Types.TIMESTAMP == sqlType) {
            return DataType.kTypeTimestamp;
        } else {
            throw new SQLException("Unexpected Values: " + sqlType);
        }
    }
```

or enum
```
public enum XXX{
      Bool(0),
      Int(2),
      Long(4),
      String(6),
   ....
}
```

PR #694, opened 2021-11-15
fix: sql type convert to data type

shawn-happy:
java.sql.Types convert to data type #679


java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/sdk/Common.java
```diff
@@ -63,4 +63,28 @@ public static int type2SqlType(DataType dataType) throws SQLException {
         }
         return new com._4paradigm.openmldb.sdk.Schema(columnList);
     }
+
+    public static DataType toDataType(int sqlType) throws SQLException {
```
vagetablechicken: ‘sqlTypeToDataType’ is more clear

PR #710, opened 2021-11-17
feat: add unsafe getasstring

### LINKED ISSUE #662: ResultSet refactor `GetAsString`

`ResultSet` is the interface used by the client to access result data set iterately.
The `const std::string GetAsString(uint32_t idx) ` method is responsible to Get value at specific index and convert it to string and return it. 
In fact, the `GetAsString` is unsafe, since it returns value unsafely. 
For instance, the method will return `"NA"` when the value is invalid or the index is out of range. 

We are going to do something to refactor this method:
- Rename the original `GetAsString` as `GetAsStringUnSafe` so that the uses will not feel confused with the unsafe behaviours.
- Add parameter default_na_value to `GetAsStringUnSafe` for specifying the default na string.  `default_na_value` is set to `NA` by default. `GetAsStringUnSafe(uint32_t index, const std::string& default_na = "NA")`
- Add `bool GetAsString(uint32_t idx, std::string* val)` for safely use.

**Additional context**

zekai427:
- Resolved: https://github.com/4paradigm/OpenMLDB/issues/662


tobegit3hub (approved):
LGTM

hybridse/include/sdk/result_set.h
```diff
@@ -95,17 +106,26 @@ class ResultSet {
...
             }
         }
+        return false;
+    }
 
-        return "NA";
+    const std::string GetAsStringUnsafe(uint32_t idx, const std::string& default_na_value = "NA") {
+        std::string val;
+        if (!GetAsString(idx, val)) {
+            return default_na_value;
```
jingchen2222, line 125: Add ut like : GetAsStringUnsafe(100)

jingchen2222 (approved):
LGTM

PR #761, opened 2021-11-22
feat: support spark conf in taskmanager

### LINKED ISSUE #724: Support flexible Spark parameters for TaskManager
Ad-hoc Spark parameters should be supported for jobs in TaskManager.

zekai427:
-Resolved: https://github.com/4paradigm/OpenMLDB/issues/724


src/proto/taskmanager.proto
```diff
@@ -13,6 +13,7 @@ message RunBatchSqlRequest {
     required string sql = 1;
     optional string db_name = 2;
     optional string output_table_name = 3;
+    map<string, string> conf = 4;
```
tobegit3hub, line 16: required or optional?

tobegit3hub (changes requested):

src/proto/taskmanager.proto
zekai427: In Protobuf, map can't add prefix

PR #783, opened 2021-11-26
feat: support physical LoadDataPlan in spark planner

### LINKED ISSUE #726: Support physical plan of "LOAD DATA" in Spark planner
Support importing data with Spark in cluster mode.

vagetablechicken:
Closes #726 
Not full functioned, need to support to get offline address from nameserver. And no unit test about hdfs.


java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/nodes/LoadDataPlan.scala
```diff
@@ -0,0 +1,87 @@
...
+    val inputFile = node.File()
+    val spark = ctx.getSparkSession
+
+    // read input file
+    val (format, options) = parseOptions(node)
+    logger.info("format {}, options {}", format, options: Any)
+    val df = spark.read.options(options).format(format).load(inputFile)
+
+    // write, offline address may contains some files, use append mode
+    df.write.mode("append").parquet(offlineAddress)
```
tobegit3hub: Choose `append` or `overwrite` from options. And default value should be `error_if_exists`.

java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/nodes/LoadDataPlan.scala
```diff
@@ -0,0 +1,87 @@
...
+
+object LoadDataPlan {
+  private val logger = LoggerFactory.getLogger(this.getClass)
+
+  def parseOptions(node: PhysicalLoadDataNode): (String, Map[String, String]) = {
+    var format = "csv"
+    var option = node.GetOption("format")
+    if (option != null) {
+      val f = option.GetStr()
+      if (f.equalsIgnoreCase("parquet")) {
```
tobegit3hub: We can use `match` syntax and switch-case in Scala.

vagetablechicken: ok

java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/nodes/LoadDataPlan.scala
vagetablechicken: So the offline address which we get from nameserver should be a new path?
I don't think this is good that `LOAD DATA` has the option `(write)mode`. Too confused.

java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/nodes/LoadDataPlan.scala
vagetablechicken: There's no need to match-case, options["format"] is a string, just need to get lowercase and assign it to return value.

tobegit3hub:
CICD failed and make sure ` com.4paradigm.hybridse:hybridse-sdk:jar:0.4.0-SNAPSHOT`  has been deployed.

java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/nodes/LoadDataPlan.scala
tobegit3hub: Users may `LOAD DATA` for multiple times and we may support different modes including `overwrite` and `append`.

vagetablechicken: but the offline address is a path, so we will create a new path for a file? e.g. load data infile f1.csv, we will get a offline address like 'f1_csv_path'?

tobegit3hub: Yes. For deep-copy mode(not symbolic mode), we will generate a new path(mostly HDFS path). Then load data and write into new path. Users may specify overwriting the old imported data or append to the old imported data.

vagetablechicken: ok, so `LOAD DATA` will have the option `mode`. I'll fix it.

aceforeverd (approved):

tobegit3hub (approved):
LGTM

vagetablechicken:

PR #795, opened 2021-11-29
feat: a new lru_cache that support upsert

### LINKED ISSUE #604: bug: SqlClusterRouter cache table name, but do not check table id
https://github.com/4paradigm/OpenMLDB/blob/ca8ce048423984347783580baf81a9a8188c8502/src/sdk/sql_cluster_router.cc#L1077-L1094
The cache logic is: if we use the same sql, get the same cache, and use `cache->name()` to  find tablet.
But if we cache the sql of table t1(tid=1), then recreate table t1(tid=2), use the same sql, we will get the cache of t1(tid=1), it's not ok.

vagetablechicken:
Closes #604 
Sql cache(by name) may be invalid cuz we may recreate the table(name is the same, but tid changed).
So when we met invalid sql cache, we shouldn't use it, and the sql cache should be update later(by SetCache).
boost::lru_cache doesn't support upsert, so we add a new lru_cache.

### DIFF
--- a/src/base/lru_cache.h
+++ b/src/base/lru_cache.h
@@ -1,0 +1,120 @@
+//---------------------------------------------------------------------------//
+// Copyright (c) 2013 Kyle Lutz <kyle.r.lutz@gmail.com>
+//
+// Distributed under the Boost Software License, Version 1.0
+// See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt
+//
+// See http://boostorg.github.com/compute for more information.
+//---------------------------------------------------------------------------//
+
+#ifndef SRC_BASE_LRU_CACHE_H_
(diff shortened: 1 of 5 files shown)


src/base/lru_cache.h
```diff
@@ -0,0 +1,121 @@
...
+        } else {
+            typename list_type::iterator j = i->second.second;
+            if (j != m_list.begin()) {
+                // move to the front
+                m_list.erase(j);
+                m_list.push_front(key);
+                j = m_list.begin();
+            }
+            // update the value too
+            m_map[key] = std::make_pair(value, j);
```
dl239, line 66: `m_map` need not update if j equal begin

dl239 (approved):

aceforeverd (approved):

vagetablechicken:

lumianph:

lumianph:

PR #798, opened 2021-11-29
feat: add catalog api for openmldb batch

### LINKED ISSUE #717: Add catalog API for NameServer to get table info
Now we have APIs to get databases and tables. It would be better to have APIs to get catalog info.

tobegit3hub:
* Add get tables api for client in cluster sql router 
* Add get databases and get tables api in openmldb catalog service
* Resolve https://github.com/4paradigm/OpenMLDB/issues/717

### DIFF
--- a/java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/catalog/OpenmldbCatalogService.scala
+++ b/java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/catalog/OpenmldbCatalogService.scala
@@ -17,7 +17,9 @@
 package com._4paradigm.openmldb.batch.catalog
 
 import com._4paradigm.hybridse.LibraryLoader
-import com._4paradigm.openmldb.{SQLRouterOptions, sql_router_sdk}
+import com._4paradigm.openmldb.{SQLRouterOptions, Status, VectorString, sql_router_sdk}
+
+import scala.collection.convert.ImplicitConversions.`collection AsScalaIterable`
 
 class OpenmldbCatalogService(val zkCluster: String, val zkPath: String) {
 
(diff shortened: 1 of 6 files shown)


dl239 (dismissed):

java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/catalog/OpenmldbCatalogService.scala
```diff
@@ -35,6 +37,29 @@ class OpenmldbCatalogService(val zkCluster: String, val zkPath: String) {
...
+  def getDatabases(): List[String] = {
+    val databases = new VectorString()
+    val status = new Status
+
+    sqlRouter.ShowDB(databases, status)
+    if (status.getCode == 0) {
+      databases.toList
+    } else {
+      null
+    }
```
dl239, line 49: should delete `databases` and `status` manually after used.  refer [here](https://github.com/4paradigm/OpenMLDB/blob/main/java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/sdk/impl/InsertPreparedStatementImpl.java#L57)

vagetablechicken: Maybe we should add the methods in SqlClusterExecutor, then openmldb-batch can use SqlClusterExecutor, avoid to load jsdk libs and do delete swig objects here(let's swig types hide in SqlClusterExecutor).

tobegit3hub: We are not extending `SqlClusterExecutor` in `openmldb-jdbc` since it is designed for JDBC API. But we have to handle the same issues as `SqlClusterExecutor`.

java/openmldb-batch/src/test/scala/com/_4paradigm/openmldb/batch/catalog/TestOpenmldbCatalogService.scala
```diff
@@ -0,0 +1,32 @@
...
+
+package com._4paradigm.openmldb.batch.catalog
+
+import org.scalatest.{FunSuite, Ignore}
+
+@Ignore
+class TestOpenmldbCatalogService extends FunSuite {
+
+  test("TestOpenmldbCatalogService") {
+    val zkCluster = "127.0.0.1:2181"
```
vagetablechicken, line 25: Why ignore this test? cicd java-sdk job can run it

tobegit3hub: It required external ZooKeeper cluster to run. Developers may compile this projects without setting up ZK cluster or using another config.

We may use in-process ZK mini cluster to test the functions which requires ZK.

vagetablechicken: mini cluster needs zk too, it'll only replace the openmldb servers.
We have a lot of java tests which need zk. Maybe we should apply one rule to them.

vagetablechicken (approved):

tobegit3hub:
Close this pr and we will add new APIs in `openmldb-jdbc` to reuse the java sdk methods.

PR #802, opened 2021-11-30
feat: option "MAC_TABLET_ENABLE"  on by default

vagetablechicken:
Now we'll release the macos version of openmldb, so `MAC_TABLET_ENABLE ` should be on by default.


Makefile
```diff
@@ -36,7 +36,9 @@ endif
 ifdef TESTING_ENABLE
     OPENMLDB_CMAKE_FLAGS += -DTESTING_ENABLE=$(TESTING_ENABLE)
 endif
-
```
jingchen2222, line 39: Do we need to support MAC_TABLET_ENABLE by default.
Otherwise, we should notify users/developers that the `make` command in MacOS won't build a tablet.

vagetablechicken: I don't know why `MAC_TABLET_ENABLE ` is false by default. Do you know it?

tobegit3hub: +1 for making MAC_TABLET_ENABLE as true by default.

It doesn't increase the output binary size and meaningless for MacOS users.

tobegit3hub (approved):
LGTM

Makefile
jingchen2222: > 

Because we didn't support building a mac tablet at the very beginning. Mac tablet is experimental for a long time.

vagetablechicken: But we released the macos version of openmldb(contains servers), that means we treat it official.
Let's set `MAC_TABLET_ENABLE ` as true by default.

dl239 (approved):

Issue #808, opened 2021-12-01
udf: refine `StringRef` with locale info


aceforeverd:
**Is your feature request related to a problem? Please describe.**

as described in https://github.com/4paradigm/OpenMLDB/pull/791#discussion_r758939927
the `ucase` may handle string with different locale

I don't fully understand locale, there might be a design in more aspects

PR #841, opened 2021-12-03
feat: like prediate and like udf

### LINKED ISSUE #224: sql: support like predicate

Currently, OpenMLDB doesn't support the some basic SQL syntaxes like `HAVING` clause and `LIKE` operator. There abilities are required by data scientis for data analysis.
> A **`HAVING`** clause in [SQL](https://en.wikipedia.org/wiki/SQL) specifies that an SQL `SELECT` statement must only return rows where aggregate values meet the specified conditions.

Generally, the `HAVING` clause is used in conjunction with aggregate functions and the GROUP BY clause to control which groups are selected.

## Support `LIKE` operator

The LIKE operator supports these wildcard characters in the quoted string.

#### Syntax

```
[ NOT ] { LIKE search_pattern [ ESCAPE esc_char ] }
```

#### Parameters

- **search_pattern**

  Specifies a string pattern to be searched by the `LIKE` clause. It can contain special pattern-matching characters:

  - `%` matches zero or more characters.
  - `_` matches exactly one character.

- **esc_char**

  Specifies the escape character. The default escape character is `\`.

- **regex_pattern**

  Specifies a regular expression search pattern to be searched by the `RLIKE` clause.

### Examples:

```SQL
CREATE TABLE person (id INT, name STRING, age INT);
INSERT INTO person VALUES
    (100, 'John', 30),
    (200, 'Mary', NULL),
    (300, 'Mike', 80),
    (400, 'Dan',  50),
    (500, 'Evan_w', 16);

SELECT * FROM person WHERE name LIKE 'M%';
+---+----+----+
| id|name| age|
+---+----+----+
|300|Mike|  80|
|200|Mary|null|
+---+----+----+

SELECT * FROM person WHERE name LIKE 'M_ry';
+---+----+----+
| id|name| age|
+---+----+----+
|200|Mary|null|
+---+----+----+

SELECT * FROM person WHERE name NOT LIKE 'M_ry';
+---+------+---+
| id|  name|age|
+---+------+---+
|500|Evan_W| 16|
|300|  Mike| 80|
|100|  John| 30|
|400|   Dan| 50|
+---+------+---+


SELECT * FROM person WHERE name LIKE '%\_%';
+---+------+---+
| id|  name|age|
+---+------+---+
|500|Evan_W| 16|
+---+------+---+

SELECT * FROM person WHERE name LIKE '%$_%' ESCAPE '$';
+---+------+---+
| id|  name|age|
+---+------+---+
|500|Evan_W| 16|
+---+------+---+
```

**Additional context**

### LINKED ISSUE #769: Support new SQL syntax for job management
Since we want to support job management in OpenMLDB CLI. The SQL parser needs to be extends and add the following commands.

```
SHOW JOBS

SHOW JOB $id

STOP JOB $id

DELETE JOB $id
```

More detailed in https://github.com/4paradigm/OpenMLDB/discussions/732 .

### LINKED ISSUE #799: Add database name for PhysicalDataProviderNode
Now we need to know the database name when resolving SQL and get PhysicalDataProviderNode. It would be better to add the database name for PhysicalDataProviderNode.

```
class PhysicalDataProviderNode : public PhysicalOpNode {
 public:
    PhysicalDataProviderNode(const std::shared_ptr<TableHandler> &table_handler,
                             DataProviderType provider_type)
        : PhysicalOpNode(kPhysicalOpDataProvider, true),
          provider_type_(provider_type),
          table_handler_(table_handler) {}
    ~PhysicalDataProviderNode() {}

    base::Status InitSchema(PhysicalPlanContext *) override;

    static PhysicalDataProviderNode *CastFrom(PhysicalOpNode *node);
    const std::string &GetName() const;
    const DataProviderType provider_type_;
    const std::shared_ptr<TableHandler> table_handler_;
};
```

aceforeverd:
### Features

resolve #224 
detail sql rules found in #224 and #686 

### What not implemented yet:

- convert <match character> and <pattern> into string if possible, currently only support string or null
- data exception: invalid escape sequence is not checked

### Further work

- Implicit conversion for like's `<target expression>` & `<pattern expression>`
- support `RLIKE`/`SIMILAR TO` predicate (regexp_match)


hybridse/src/udf/udf.cc
```diff
@@ -250,6 +251,138 @@ void timestamp_to_date(codec::Timestamp *timestamp,
...
+*
+* rules:
+* - default case insensitive
+* - escape
+*   - exception(invalid escape character) if escape size >= 2
+*   - exception(invalid pattern) if pattern is escape character ended
+*   - exception check should check outside, it is not checked here
+*   - empty string means disable escape
+* - nullable
+*   - any of (name, pattern, escape) is null, return null
```
aceforeverd: not escape

hybridse/src/udf/udf.cc
```diff
@@ -250,6 +251,138 @@ void timestamp_to_date(codec::Timestamp *timestamp,
...
+    return n_it == n_end;
+}
+
+/*
+* SQL style glob match, it use
+* - percent sign (%) as zero or more characters
+* - underscore (_) as exactly one.
+* - backslash (\) as escape character by default
+*
+* rules:
```
jingchen2222: The rules are quite different from what issue #224 described. I think that is because #224 borrows rules from T-SQL but the current implemented rules follow MySQL's rules to some extend.
- A little concern about the rule that the pattern is default case insensitive. What if the users try to match a string starting with 'A' exactly? I have noticed the `regexp_match ` function but didn't find how to invoke it?
- In my opinion, T-SQL's rules are much easier to implement.
   - pattern is case sensitive by default.
   - using LIKE '[Aa]%' to match a string starting with 'a' or 'A'
   - using LIKE 'A%' to match a string starting with 'A' exactly.

However, since there is no standard for the LIKE pattern for SQL, so both kinds of rules are ok. 

Please update #224 with our new rules to keep consistency. 
Update related SQL documentation after the pr is merged.

hybridse/src/udf/udf_test.cc
```diff
@@ -664,6 +664,86 @@ TEST_F(ExternUdfTest, TestCompoundTypedExternalCall) {
         &library, "make_tuple", TupleResT(1, nullptr, 3), 1, nullptr, 3);
 }
 
+TEST_F(ExternUdfTest, LikeMatchTest) {
```
jingchen2222, line 667: Can we perform `LIKE` on Chinese characters or other Unicode chars?
If so, add some related unit tests and YAML cases.
If non-support currently, add some comments and documentation. And throw an error during converting ASTNode or expression evaluation.

hybridse/src/udf/default_udf_library.cc
```diff
@@ -666,6 +666,35 @@ void DefaultUdfLibrary::InitStringUdf() {
...
                 --output "2020-05-22"
             @endcode)");
+    RegisterExternal("like_match")
+        .args<StringRef, StringRef, StringRef>(reinterpret_cast<void *>(
+            static_cast<void (*)(codec::StringRef *, codec::StringRef *, codec::StringRef *, bool *, bool *)>(
+                udf::v1::like_match)))
+        .return_by_arg(true)
+        .returns<Nullable<bool>>()
+        .doc(R"r(
+                @brief perform SQL LIKE match
```
jingchen2222: Please add enough rules and examples for "like_match" in `.doc()`, since it is really important and there are no standard rules for that.

hybridse/src/udf/udf.cc
aceforeverd: I follow the ansi sql and postgre sql's guide as well: https://www.postgresql.org/docs/current/functions-matching.html

That detail of like predicate may different in many ways, but should easy to tweak, e.g for case sensitive just change the equal function.

bracket character is not implemented, it seems not the standard of ansi sql.

aceforeverd: BTW, there does has a general problem:

It may not appropriate to use backslash(`\`) as default escape character, since zetasql parser backslash as escape char as well. As a result, in string `'abc\n'`, `'\n'` is treated as special character, and in a like predicate, we must write `col1 LIKE 'abc\\n' ESCAPE '\\` to represent that backslash. Also, in yaml, `"abc\\n"` seems have different meanings.

So I'd wondering if we should tweak a lit of escape default:
1. if `ESCAPE` clause not exist, disable escape feature
2. documented: in sql, use `\\` to represent backslash
3. There is no default escape character and users are encourage to escape with another character(prefer not backslash):
    `c1 LIKE 'a$%' ESCAPE '$'`

hybridse/src/udf/udf_test.cc
aceforeverd: I'm afraid not, see #808 , the string going to compare don't have locale info, and the simple u8 char compare is performed

jingchen2222: It's okay. Inform users in function documentation and our language guide.

aceforeverd:
[sql support for LIKE](https://trello.com/c/ZgAcOGCw/26-sql-support-for-like)

hybridse/src/udf/udf.cc
jingchen2222: 1. if `ESCAPE` clause not exist, using '\\' as default escape as described in https://github.com/4paradigm/OpenMLDB/discussions/686.

hybridse/src/udf/udf.cc
```diff
@@ -252,6 +253,151 @@ void timestamp_to_date(codec::Timestamp *timestamp,
...
+            if (!equal(c, *n_it)) {
+                return false;
+            }
+
+            std::advance(it, 1);
+        } else if (c == *escape) {
+            // exact character match
+            if (std::next(it) == end) {
+                // the pattern is terminated with escape character, just return false
+                return false;
```
aceforeverd, line 290: add a case

hybridse/src/udf/udf.cc
```diff
@@ -252,6 +253,151 @@ void timestamp_to_date(codec::Timestamp *timestamp,
...
+        out, is_null);
+}
+
+void ilike(codec::StringRef* name, codec::StringRef* pattern, bool* out, bool* is_null) {
+    static codec::StringRef default_esc(1, "\\");
+    ilike(name, pattern, &default_esc, out, is_null);
+}
+
+/* case insensitive regex match */
+void regexp_match(codec::StringRef* name, codec::StringRef* pattern,
```
aceforeverd: rm unused code

hybridse/src/udf/default_udf_library.cc
```diff
@@ -666,9 +666,152 @@ void DefaultUdfLibrary::InitStringUdf() {
...
+                   - if <escape character> is empty or null value, escape feautre is disabled
+                3. case sensitive
+                4. backslash: sql string literal use backslash(\) for escape sequences, write '\\' as backslash itself
+
+                Example:
+                @code{.sql}
+                    select like_match('Mike', 'Mi_k', '\\')
+                    -- output: true
+
+                    select like_match('Mike', 'Mi_k', '')
```
aceforeverd: TODO: run with postgre

'Mi_e', 'Mi\_e', '\' -> true  
'Mi_e', 'Mi\_e', '' -> false

hybridse/src/udf/udf.cc
```diff
@@ -252,6 +253,151 @@ void timestamp_to_date(codec::Timestamp *timestamp,
...
+
+    while (it != end) {
+        if (n_it == n_end) {
+            return false;
+        }
+
+        char c = *it;
+        if (escape == nullptr) {
+            // escape feature is disabled
+            if (!equal(c, *n_it)) {
```
aceforeverd: fix this

hybridse/src/udf/udf_test.cc
```diff
@@ -664,6 +664,183 @@ TEST_F(ExternUdfTest, TestCompoundTypedExternalCall) {
...
+
+    check_like(true, false, "Mary", "M_ry", "\\");
+    check_like(false, false, "Mary", "m_ry", "\\");
+
+    // match empty
+    check_like(true, false, "", "", "\\");
+    check_like(false, false, "abc", "", "\\");
+    check_like(false, false, "", "abc", "\\");
+
+    // esacpe character
```
zhanghaohit: escape

hybridse/src/udf/udf_test.cc
```diff
@@ -664,6 +664,183 @@ TEST_F(ExternUdfTest, TestCompoundTypedExternalCall) {
...
+    check_like(false, false, "Mike", "mary", "\\");
+    check_like(false, false, "Evan_W", "%_x", "\\");
+    // it is exact match
+    check_like(false, false, "Evan_W", "Evan", "\\");
+    check_like(false, false, "Super", "SupberGlob", "\\");
+
+    // disable escape by provide empty string
+    check_like(true, false, R"r(Evan\%w)r", R"r(Evan\%w)r", "");
+    check_like(false, false, R"r(Evan%w)r", R"r(Evan\%w)r", "");
+    check_like(true, false, R"r(Evan%w)r", R"r(Evan\%w)r", "\\");
```
aceforeverd, line 721: add test for meaning of % and _ when escape char disable

hybridse/src/udf/udf_test.cc
```diff
@@ -664,6 +664,183 @@ TEST_F(ExternUdfTest, TestCompoundTypedExternalCall) {
...
+    check_ilike(true, false, "", "", "\\");
+    check_ilike(false, false, "abc", "", "\\");
+    check_ilike(false, false, "", "abc", "\\");
+
+    // esacpe character
+    check_ilike(true, false, "Evan_W", R"r(%\_%)r", "\\");
+    check_ilike(false, false, "EvansW", R"r(%\_%)r", "\\");
+    check_ilike(true, false, "Evan_W", "%$_%", "$");
+    check_ilike(false, false, "EvansW", "%$_%", "$");
+    check_ilike(true, false, "Evan%W", "eva_p%w", "p");
```
aceforeverd, line 801: ignore case of escape character 'p' ?

aceforeverd: no. add a case

cases/query/simple_query.yaml
```diff
@@ -463,3 +463,297 @@ cases:
...
+    desc: disable escape
+    inputs:
+      - name: t1
+        columns: ["col1:int32", "std_ts:timestamp", "col2:string"]
+        indexs: ["index1:col1:std_ts"]
+        rows:
+          - [1, 1590115420001, "Mike"]
+    dataProvider:
+      - ["like_match", "ilike_match"]
+      - ["''" ]
```
aceforeverd: string(null)

cases/query/simple_query.yaml
```diff
@@ -463,3 +463,297 @@ cases:
...
+  - id: like_predicate_7
+    desc: nullable2, PC:null -> null
+    inputs:
+      - name: t1
+        columns: ["col1:int32", "std_ts:timestamp", "col2:string", "col3:string"]
+        indexs: ["index1:col1:std_ts"]
+        rows:
+          - [1, 1590115420001, Mike, null]
+    sql: |
+      select col2 LIKE col3 as col2 from t1;
```
aceforeverd, line 560: add a case if pattern is columnref

hybridse/include/vm/physical_op.h
```diff
@@ -559,6 +562,9 @@ class PhysicalDataProviderNode : public PhysicalOpNode {
 
     static PhysicalDataProviderNode *CastFrom(PhysicalOpNode *node);
     const std::string &GetName() const;
+    const std::string& GetDb() const {
```
aceforeverd, line 565: add a case

cases/query/simple_query.yaml
jingchen2222: There might be a crush for `string(null)`. https://github.com/4paradigm/OpenMLDB/issues/854.

jingchen2222:
>>> 
> Implicit conversion for like's <target expression> & <pattern expression>
support RLIKE/SIMILAR TO predicate (regexp_match)

Create related issues and link to them.

cases/query/simple_query.yaml
jingchen2222: Will support when #855 is merged

jingchen2222 (approved):
Add like string(null) cases and Merge when  #855  has been merged

jingchen2222 (dismissed):
Add like string(null) cases when #855 has been merged

aceforeverd:
`string(NULL)` crash, depends #855

aceforeverd:
> Add like string(null) cases when #855 has been merged

BTW, I'd like update behavior of `(i)like_match(___, string(null))`, it default to return `NULL`, instead treat like match as escape mechanism disabled. Which also consistent with ansi sql

aceforeverd:
@jingchen2222  looks like `select like_match(__, string(NULL)) ` still failed, but only on linux( a address boundry error)

aceforeverd:
@jingchen2222  looks like `select like_match(__, string(NULL)) ` still failed, but only on linux( a address boundry error)

jingchen2222:
> @jingchen2222 looks like `select like_match(__, string(NULL)) ` still failed, but only on linux( a address boundry error)
We need to figure out the root cause.  
BTW, `string(NULL)` seems to work on Linux since #855 passed the cicd.

aceforeverd:
> > @jingchen2222 looks like `select like_match(__, string(NULL)) ` still failed, but only on linux( a address boundry error)
> > We need to figure out the root cause.
> > BTW, `string(NULL)` seems to work on Linux since #855 passed the cicd.

Yes,  disable that case temporally to pass cicd. I'll check this problem later

aceforeverd:
@jingchen2222 @dl239 @tobegit3hub this is pr is ready to merge after cicd passed

tobegit3hub (approved):
LGTM

dl239 (approved):

PR #864, opened 2021-12-08
feat: add global variables and replace the config in hybridse flags

tobegit3hub:
* Remove the gflags parameters in hybridse flags
* Add global variables for `g_enable_spark_unsaferow_format`


tobegit3hub:
We will discuss later if we want to remove the hybridse flags.

PR #870, opened 2021-12-09
fix: missing nproc when compiling OpenMLDB on macOS

Shouren:

A bug fix for missing nproc when compiling

See the screenshot below
Screen Shot 2021-12-09 at 1 20 24 PM

No more error of ```nproc: command not found```

No, 

* **Other information**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,15 @@
 
 MAKEFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 MAKEFILE_DIR  := $(dir $(MAKEFILE_PATH))
-NPROC ?= $(shell (nproc))
+
+# Processor count with OS detection
+ifeq ($(shell (uname -s)), Linux)
+	NPROC ?= $(shell (nproc))
+endif
+ifeq ($(shell (uname -s)), Darwin)
+	NPROC ?= $(shell (sysctl -n hw.ncpu))
+endif
+
 
 CMAKE_PRG ?= $(shell (command -v cmake3 || echo cmake))
 CMAKE_BUILD_TYPE ?= RelWithDebInfo


Makefile
```diff
@@ -14,7 +14,15 @@
...
 MAKEFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 MAKEFILE_DIR  := $(dir $(MAKEFILE_PATH))
-NPROC ?= $(shell (nproc))
+
+# Process count with OS detection
+ifeq ($(shell (uname -s)), Linux)
+	NPROC ?= $(shell (nproc))
+endif
+ifeq ($(shell (uname -s)), Darwin)
+	NPROC ?= $(shell (sysctl -n hw.physicalcpu))
```
aceforeverd: Logicialcpu

Shouren: According to https://developer.apple.com/documentation/foundation/processinfo/1415622-processorcount, fixed it with hw.ncpu

aceforeverd (approved):

tobegit3hub (approved):
LGTM

Makefile
```diff
@@ -14,7 +14,15 @@
 
 MAKEFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 MAKEFILE_DIR  := $(dir $(MAKEFILE_PATH))
-NPROC ?= $(shell (nproc))
```
aceforeverd, line 17: as described in this thread https://github.com/4paradigm/OpenMLDB/pull/874#issuecomment-989707092, using the max cpu number may kill the build. We'd wondering if it is still necessary  to keep NPROC default to $(nproc)
@zhanghaohit

zhanghaohit: we may change the default `NPROC` to a lower number, say 4? in case the building process freezes the whole system.

superzhaozheng (approved):
LGTM

Makefile
Shouren: Monitoring the CPU and memory usage during a compile on my VPS with 2 cores and 4GB memory.
### When ``` NPROC=1 ```
Screen Shot 2021-12-10 at 12 35 38 PM
Screen Shot 2021-12-10 at 12 36 32 PM
It takes nearly 2GB mem when compiling ``` avg ``` or ``` count ``` function.

### When ``` NPROC=2 ```
Screen Shot 2021-12-10 at 12 38 15 PM
The screen shot shows the last output of ```top``` before the building process freezes the whole system. And we can see the cpu usage of ```[kswapd0]``` which indicates the whole system is running out of memory.

### Conclusion
So it is not how many building process that matters, even a VPS with 4cores and 4GB memory may be frozen by ```NPROC=2```.
**What should we do is to find out why compiling those built-in functions takes so much memory.**

aceforeverd: Nice experiment. Yet I don't know too much about the reason or how to improve it. Maybe with another build tool or dynamic link?
`avg_by_category_def.cc` compile is a known problem, I'll create a issue if not exists. #882 

BTW can u try build with ninja instead of make, by `cmake ... -G Ninja`, it is said to have better memory usage.

Shouren: Tried with Ninja in my VPS, no luck when ```NPROC=2```

aceforeverd:
close in favor of #889 
we'll set nproc = 1 by default,

PR #873, opened 2021-12-09
feat: add new static hybridse jsdk and link to openmldb jsdk

tobegit3hub:
* Add new static library for hybridse jsdk
* Link openmldb jsdk with hybridse jsdk
* openmldb-batch will load openmldb jsdk instead of hybridse jsdk


java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/sdk/impl/SqlClusterExecutor.java
```diff
@@ -72,6 +83,22 @@ public SqlClusterExecutor(SdkOption option) throws SqlException {
         }
     }
 
+    public static String getSqlJsdkLibraryPath() {
```
vagetablechicken: Should this function move to LibraryLoader class? It's better to put `find library` functions(load/extract/getLibPath) together.

java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/OpenmldbBatchConfig.scala
```diff
@@ -120,7 +120,7 @@ class OpenmldbBatchConfig extends Serializable {
   @ConfigOption(name="openmldb.disable", doc="Disable OpenMLDB optimization or not")
   var disableOpenmldb = false
 
-  // HybridSE dynamic library path
+  // OpenMLDB Java SDK dynamic library path, notice that this should not be set hybridse jsdk so
```
vagetablechicken, line 123: It should be noted that, we set `hybridseJsdkLibraryPath` to openmldb jsdk lib, because we can't load openmldb jsdk and hybridse jsdk at the same time. And openmldb jsdk has all symbols of hybridse jsdk, so it works.

src/sdk/CMakeLists.txt
```diff
@@ -134,7 +134,9 @@ if(SQL_JAVASDK_ENABLE)
             SOURCES sql_router_sdk.i)
     target_include_directories(sql_jsdk PRIVATE ${JNI_INCLUDE_DIRS})
     target_compile_options(sql_jsdk PRIVATE -w)
-    target_link_libraries(sql_jsdk PRIVATE ${SDK_LIBS} boost_filesystem)
+
+    add_dependencies(sql_jsdk hybridse_jsdk_core_static)
+    target_link_libraries(sql_jsdk PRIVATE "-Wl,-force_load $<TARGET_FILE:hybridse_jsdk_core_static>" ${SDK_LIBS} boost_filesystem)
```
vagetablechicken: `-force_load` is only for macos, linux will use `--whole-archive` and `--no-whole-archive`

tobegit3hub:
[End-to-end offline workflow features](https://trello.com/c/0AErqe3E/32-end-to-end-offline-workflow-features)

java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/OpenmldbBatchConfig.scala
tobegit3hub: We will remove this config later since it has to be the same so file of openmldb-jdbc.

java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/sdk/impl/SqlClusterExecutor.java
tobegit3hub: Since we will alway find the path of "sql_jsdk", it is not suitable to add in `LibraryLoader`.

vagetablechicken (approved):

aceforeverd (approved):

tobegit3hub:
[End-to-end offline workflow features](https://trello.com/c/0AErqe3E/32-end-to-end-offline-workflow-features)

PR #908, opened 2021-12-15
fix: method GetTableInfo may get nullptr

vagetablechicken:
If `cluster_sdk_->GetTableInfo` get nullptr, we can't do `*table_infos`, it'll core.


tobegit3hub (approved):
LGTM

aceforeverd (approved):

PR #982, opened 2021-12-28
fix(udf): like_match

### LINKED ISSUE #898: There is a bug in the like expression

c1 like 'a%b%'

a_b -> false

## Expected Behavior
a_b -> true

## Current Behavior
a_b -> false

aceforeverd:
resolve #898, if pattern ends with `<percent>s`, and name reaches end
while matching pattern, it need extra check before return


jingchen2222 (approved):
LGTM

vagetablechicken (approved):

lumianph:

PR #986, opened 2021-12-29
fix: after each sql sentence is forced to add a semicolon to be deleted

### LINKED ISSUE #978: Remove the code of concating SQL with semicolon in openmldb-batch
Now we will add the semicolon in the end of SQL text which is not required by internal SQL engine now. We should remove the code in `OpenmldbSession.scala`.

```
    var sql: String = sqlText
    if (!sql.trim.endsWith(";")) {
      sql = sql.trim + ";"
    }
```

yabg-shuai666:
close #978

### DIFF
--- a/java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/api/OpenmldbSession.scala
+++ b/java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/api/OpenmldbSession.scala
@@ -155,7 +155,6 @@ class OpenmldbSession {
       return OpenmldbDataframe(this, sparksql(sqlText))
     }
 
-
     val planner = new SparkPlanner(this, config)
     this.planner = planner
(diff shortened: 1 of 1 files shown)


jingchen2222 (approved):
LGTM

java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/api/OpenmldbSession.scala
```diff
@@ -156,9 +156,7 @@ class OpenmldbSession {
     }
 
     var sql: String = sqlText
```
tobegit3hub: Do not need to use `sql` now. Remove this and use `sqlText` directly.

yabg-shuai666: already edited

tobegit3hub (approved):
LGTM

PR #988, opened 2021-12-29
feat: build wheel file for specific platform (Mac + Linux)

### LINKED ISSUE #957: openmldb-sqlalchemy in pypi does not support MacOS
Maybe Do not allow to install in MacOS by setting supported OS in `setup.py`.

HuilinWu2:
Add platform tag while building wheel file. Currently, we may maintain wheel files for specific operating system.


tobegit3hub (approved):
LGTM

tobegit3hub:
Resolve https://github.com/4paradigm/OpenMLDB/issues/957 .

src/sdk/CMakeLists.txt
```diff
@@ -110,7 +110,7 @@ if(SQL_PYSDK_ENABLE)
                 # Do not strip library for MacOS, refer to https://github.com/4paradigm/OpenMLDB/issues/905
                 COMMAND ${CMAKE_COMMAND} -E copy ${PROJECT_SOURCE_DIR}/python/sqlalchemy-openmldb/setup.py ${CMAKE_BINARY_DIR}/python/
                 COMMAND ${CMAKE_COMMAND} -E remove_directory dist
-                COMMAND cd ${CMAKE_BINARY_DIR}/python/ && ${Python3_EXECUTABLE} setup.py bdist_wheel
+                COMMAND cd ${CMAKE_BINARY_DIR}/python/ && ${Python3_EXECUTABLE} setup.py bdist_wheel --plat-name macosx_10_9_x86_64
```
vagetablechicken: 10_9? Our pysdk lib is built on 10.15, so 10_15 is right?

HuilinWu2: `--plat-name macosx_10_15_x86_64`  is also right. Here we want to set the lowest version of Mac OS(after testing, is 10_9) that supports `pip install openmldb` from PyPi instead of a specific version 10_15.

vagetablechicken: Do we test pysdk on 10.9 OS?

HuilinWu2: We've tested pysdk on 10.15 OS and wheel file with newer version name than 10_15 such as `openmldb_py3_none_macosx_10_16_x86_64.whl` or older that 10_9 such as `openmldb_py3_none_macosx_10_8_x86_64.whl` won't be installed by `pip` method.

aceforeverd: I think it is no need to maintain compatibility for macos < 10.15.

HuilinWu2: updated, currently we will provide `openmldb_py3_none_macosx_10_15_x86_64.whl` for users

vagetablechicken (approved):

PR #996, opened 2021-12-30
feat: add batchjob in taskmanager and use the default batchjob jar

tobegit3hub:
* Include batchjob in taskmanager binary.
* Set `batchjob.jar.path` to find the default jar so that users do not need to set config before running taskmanager.
* Remove batchjob when installing the project which is included in taskmanager libraries

### DIFF
--- a/java/openmldb-taskmanager/pom.xml
+++ b/java/openmldb-taskmanager/pom.xml
@@ -98,13 +98,6 @@
             <scope>provided</scope>
         </dependency>
 
-        <!-- OpenMLDB -->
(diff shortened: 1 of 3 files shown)


vagetablechicken (approved):

aceforeverd (approved):

zhanghaohit (approved):

tobegit3hub:
[Implementation of TaskManager](https://trello.com/c/EkxZMO7i/28-implementation-of-taskmanager)

PR #998, opened 2021-12-30
fix: improve load data test and coverage fix

vagetablechicken:
Load data test can't work well, cuz droping table always fails(taskmanager is unalive so can't get in drop table function).
So ignore the offline load test.
Use test.properties to set openmldb zk in test env.
Update test src csv files.
Update logic of load validation.


aceforeverd (approved):

tobegit3hub (approved):
LGTM

vagetablechicken:

PR #1005, opened 2021-12-31
fix: timestamp type and catch all in DataWriter

vagetablechicken:
use normal timestamp in offline storage
df use sql timestamp(s), df write InternalRow use spark sql TimestampType(us), openmldb use long(int64, ms)


aceforeverd (approved):

tobegit3hub (approved):
LGTM

vagetablechicken:

PR #1008, opened 2022-01-04
fix: fix deploy index

### LINKED ISSUE #1006: When creating an index during deploy, you need to judge whether it is the same index according to column and ts
When creating an index during deploy, you need to judge whether it is the same index according to column and ts

## Expected Behavior
Determine whether the index is duplicate through column and ts

## Current Behavior
Judge whether the index is duplicate only according to the column

dl239:
fix #1006


vagetablechicken (approved):

tobegit3hub (approved):
LGTM

lumianph:

PR #1038, opened 2022-01-06
fix: fix incorrect job type of select into job

tobegit3hub:
* The job type of offline `select into` is incorrect and we should fix by invoking the correct api.


vagetablechicken (approved):

aceforeverd (approved):

tobegit3hub:
[Implementation of TaskManager](https://trello.com/c/EkxZMO7i/28-implementation-of-taskmanager)

PR #1074, opened 2022-01-12
fix: transform test column resolve fail test

aceforeverd:
two major changes:

- fix `transform_test` 
- add a cache step for macos job, that cache thirdparty components

PR #1078, opened 2022-01-12
feat: refresh catalog in cmd when importing offline table

tobegit3hub:
* Add refresh catalog so that we can describe table to see offline storage after importing in the same client


vagetablechicken (approved):

aceforeverd:
wait the `transform_test` fix #1074

tobegit3hub:
Do not work for expected result. Will address the issue later.

PR #1106, opened 2022-01-17
feat: replace PROCEDURE with DEPLOYMENT in the result of show deployment

### LINKED ISSUE #1063: Replace `PROCEDURE` with `DEPLOYMENT` in the result of  show deployment
When execute `show deployment xxx` or curl "http:://xxx/deployment/xxx"， the result will display `CREATE PROCEDURE` is  inaccurate. It should show the deploy statement

keyu813:
close #1063


src/cmd/sql_cmd.h
```diff
@@ -349,8 +349,13 @@ void PrintProcedureInfo(const hybridse::sdk::ProcedureInfo& sp_info) {
     if (sp_info.GetType() == hybridse::sdk::kReqDeployment) {
         type_name = "Deployment";
     }
+    std::string sql = sp_info.GetSql();
+    if (sp_info.GetType() == hybridse::sdk::kReqDeployment) {
+        // replace the sql "CREATE PROCEDURE" with "DEPLOY"
+        sql.replace(0u, 16u, "DEPLOY");
```
tobegit3hub: Will users use "CREATE PROCEDURE"? Replace string directly is strange if we support standard "CREATE PROCEDURE" in the future.

keyu813: Actually, We hope user use Deployment instead of Procedure, and we should decouple the code between them. But right now sql haven't equal Deployment and Procedure, so I just replace string directly

keyu813: > Will users use "CREATE PROCEDURE"? Replace string directly is strange if we support standard "CREATE PROCEDURE" in the future.

Actually, We hope user use Deployment instead of Procedure, and we should decouple the code between them. But right now sql haven't equal Deployment and Procedure, so I just replace string directly

src/cmd/sql_cmd.h
```diff
@@ -349,8 +350,15 @@ void PrintProcedureInfo(const hybridse::sdk::ProcedureInfo& sp_info) {
     if (sp_info.GetType() == hybridse::sdk::kReqDeployment) {
         type_name = "Deployment";
     }
+    std::string sql = sp_info.GetSql();
+    if (sp_info.GetType() == hybridse::sdk::kReqDeployment) {
```
dl239, line 353: move to the line350 if statement

dl239:
merge the upstream

vagetablechicken (approved):

dl239 (approved):

PR #1110, opened 2022-01-18
feat: update taskmanager unit tests

tobegit3hub:
* Add `getJobLog` and `close` methods in taskmanager client
* Add license for java files
* Enable taskmanager client unit tests which use TaskManager in onebox later


tobegit3hub:
We need to `mvn package` before starting onebox with TaskManager which may takes more time for cicd process.

onebox/start_onebox.sh
```diff
@@ -89,30 +89,6 @@ test -d recycle_bin3 && rm -rf recycle_bin3
                    --zk_cluster=${ZK_CLUSTER}\
                    --zk_root_path=/onebox > ns3.log 2>&1 &
 
-# start TaskManager if exists
```
vagetablechicken, line 92: why delete? Don't we need to start task manager when starting onebox?

tobegit3hub: The `start_onebox.sh` is executed before `mvn package`. The java target path does not exist when starting onebox and fail to start TaskManager(confirmed by cicd log). Now we change to start TaskManager in java unit test file which may use the latest code of TaskManager.

yabg-shuai666 (approved):

onebox/start_onebox.sh
vagetablechicken: But the normal user can't start task manager by `start_onebox.sh`, it's not good.

tobegit3hub: Normal users may use `start-all.sh` in release to setup the all-in-one cluster. IMO, `onebox` is used for unit tests currently. If you want to keep this in `onebox`, I can revert this but make sure everyone knows the TaskManager may not start in CICD environment.

vagetablechicken: I see, so `start_onebox.sh` is only for ut(I've forgot start-all.sh is for cluster, it was for standalone before). It's fine, but this script should better be renamed.

vagetablechicken (approved):

PR #1111, opened 2022-01-18
feat: download openmldb spark distribution in java cicd

tobegit3hub:
* Add script to download openmldb spark distribution
* Download and extract spark files in cicd of java

This is useful for unit tests of TaskManager which requires TaskManager with Spark to run test cases.

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -116,6 +116,7 @@ jobs:
       SQL_JAVASDK_ENABLE: ON
       OPENMLDB_BUILD_TARGET: 'cp_native_so openmldb'
       MAVEN_OPTS: -Duser.home=/github/home
+      SPARK_HOME: /tmp/spark/
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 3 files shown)


tobegit3hub:
The time of CICD may increase about 8 minutes to download Spark file. Compare with previous result.

image

vagetablechicken (approved):

.github/workflows/cicd.yaml
```diff
@@ -166,6 +167,7 @@ jobs:
       - name: start services
         run: |
           sh steps/ut_zookeeper.sh start
+          sh steps/download_openmldb_spark.sh $SPARK_HOME
```
aceforeverd, line 170: pls update coverage.yml as well

tobegit3hub: Okay.

vagetablechicken:
image
cicd error

.github/workflows/cicd.yaml
aceforeverd, line 170: should be ${{ env.SPARK_HOME }}
in a workflow runner, the shell don't know the $SPARK_HOME variable

tobegit3hub: It is the error of `coverage` which lacks of `SPARK_HOME`. This `$SPARK_HOME` works for `cicd.yaml` workflow.

aceforeverd: I' wrong

aceforeverd:
well, let's wait ⌚

vagetablechicken (approved):

aceforeverd (approved):

PR #1115, opened 2022-01-18
fix: fix start script

### LINKED ISSUE #1108: start failed if pid file is exist
WX20220117-183718@2x

dl239:
fix #1108


release/bin/start.sh
```diff
@@ -83,7 +83,10 @@ case $OP in
         then
              echo "no $COMPONENT to stop (could not find file $OPENMLDB_PID_FILE)"
         else
-            tr -d '\0' < "$OPENMLDB_PID_FILE" | xargs kill
+            PID="$(tr -d '\0' < "$OPENMLDB_PID_FILE")"
+            if kill -0 "$PID" > /dev/null 2>&1; then
+                kill "$PID"
```
tobegit3hub, line 88: Try `kill -9` to make sure it is killed?

tobegit3hub (approved):
LGTM

release/bin/start.sh
dl239: `mon` will not kill its subprocess if use `kill -9`

aceforeverd (approved):
It just work.

But not robust, if reboot happens the old pid file exists in fs, and another process lives just with that pid number.

dl239:
> It just work.
> 
> But not robust, if reboot happens the old pid file exists in fs, and another process lives just with that pid number.

yes, do you have some advice about this?  There may be another OpenMLDB process with the same pid

PR #1119, opened 2022-01-19
build: fix link error if system already has icu and boost libs

### LINKED ISSUE #1118: build(link system thirdparty ): use result variables from `find_package`
somethings compile error might happen if he installed e.g boost in `/lib64` already in there system. It may necessary to use variables set by `find_package`, like `Boost_LIBRARIES`, in case cmake accidentally link boost  to that system installed boost ( a non-standard system)

In detail, e.g  https://github.com/4paradigm/OpenMLDB/blob/72a052400996b80a5c9ae8408802e23e04370ce7/hybridse/CMakeLists.txt#L98, should replace `boost_*` with `${Boost_LIBRARIES}`

It might one of the things to support link openmldb with system libraries

zhanghaohit:
bug fix

If system has icu and boost libraries, will link system libs rather than third-party libs

link third-party static libs

No

* **Other information**:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,7 +128,8 @@ endif()
 
 add_subdirectory(hybridse)
 
-set(VM_LIBS hybridse_core hybridse_sdk hybridse_flags boost_filesystem boost_date_time boost_regex)
+find_package(Boost COMPONENTS filesystem date_time regex REQUIRED)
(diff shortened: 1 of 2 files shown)


aceforeverd (approved):

tobegit3hub (approved):
LGTM

PR #1122, opened 2022-01-19
fix: importing CSV file in offline reate files with correct schema

### LINKED ISSUE #1100: Importing CSV file in offline may create files with incorrect schema without warning
Now importing CSV file in offline may generate parquet files whose schema is not expected, maybe all string columns.

We should raise exception if the schema does not match.

yabg-shuai666:
close #1100

### DIFF
--- a/java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/nodes/LoadDataPlan.scala
+++ b/java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/nodes/LoadDataPlan.scala
@@ -117,7 +117,12 @@ object LoadDataPlan {
 
         // do deep copy
         require(inputFile != writePath, "read and write paths shouldn't be the same, it may clean data in the path")
-        val df = spark.read.options(options).format(format).load(inputFile)
+        var struct = new StructType
+        info.getColumnDescList.forEach(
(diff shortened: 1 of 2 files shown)


lumianph:
@tobegit3hub let's merge this PR ASAP, and then let the user in wechat know when it is fixed.

java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/nodes/LoadDataPlan.scala
```diff
@@ -117,7 +117,12 @@ object LoadDataPlan {
 
         // do deep copy
         require(inputFile != writePath, "read and write paths shouldn't be the same, it may clean data in the path")
-        val df = spark.read.options(options).format(format).load(inputFile)
+        var struct = new StructType
```
tobegit3hub, line 120: Please add an unit test for this added function.

tobegit3hub (changes requested):

tobegit3hub (approved):
LGTM

java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/nodes/LoadDataPlan.scala
tobegit3hub: We will add tests later.

vagetablechicken (approved):

PR #1125, opened 2022-01-20
feat: change testing framework of Python SDK and add new unittest for sqlalchemy API

HuilinWu2:
1. Change the original `nose` test framework to `pytest` framework for its convenient in both coding and implementing.
2. Add new UT for sqlalchemy table-object-based api in pytest style. the `openmldb_client_test.py` will be further rewritten into pytest style in a new PR.

### DIFF
--- a/python/test/openmldb_client_test.py
+++ b/python/test/openmldb_client_test.py
@@ -315,7 +315,7 @@ class TestOpenMLDBClient(unittest.TestCase):
       ]
     self.check_result(rs, expectRows)
 
-# test sqlalchemy Table-object-based API of pytest style
+# test sqlalchemy Table-object-based API in pytest style
 class TestSqlalchemyAPI:
 
     def setup_module(self):
--- a/python/test/openmldb_client_test.py
+++ b/python/test/openmldb_client_test.py
@@ -330,7 +330,7 @@ class TestSqlalchemyAPI:
         self.metadata.create_all(self.engine)
         
     def test_create_table(self):
-        assert self.connection.dialect.has_table('test_table')
+        assert self.connection.dialect.has_table(self.connection,'test_table')
 
     def test_insert(self):
         try:
--- a/python/test/openmldb_client_test.py
+++ b/python/test/openmldb_client_test.py
@@ -341,10 +341,8 @@ class TestSqlalchemyAPI:
 
     def test_select(self):
           for row in self.connection.execute(select([self.test_table])):
-             if 'first' in list(row):
-                assert 100 in list(row)
-            else:
-                assert False
+             assert 'first' in list(row)
+             assert 100 in list(row)
 
     def teardown_class(self):
         self.connection.execute(self.test_table.delete())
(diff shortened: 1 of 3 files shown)


python/test/openmldb_client_test.py
```diff
@@ -314,10 +316,48 @@ def test_parameterized_query(self):
       (1008, '2021-01-02', 'province3', 'city9', 9, 1590738998000),
       ]
     self.check_result(rs, expectRows)
- 
 
+# test sqlalchemy Table-object-based API in pytest style
+class TestSqlalchemyAPI:
+
```
tobegit3hub, line 322: define object members in `__init__()`.

python/test/openmldb_client_test.py
```diff
@@ -314,10 +316,48 @@ def test_parameterized_query(self):
...
+    def setup_class(self):
+        self.engine = db.create_engine('openmldb:///db_test?zk=127.0.0.1:6181&zkPath=/onebox')
```
tobegit3hub, line 324: The engine is not used in other methods. We could make it as local variable instead of object member.

python/test/openmldb_client_test.py
```diff
@@ -314,10 +316,48 @@ def test_parameterized_query(self):
...
+                if 'first' not in list(row):
+                    logging.warning("Insert failed, 'first' not in the row")
+                    assert False
+                elif 100 not in list(row):
+                    logging.warning("Insert failed, 100 not in the row")
+                    assert False
+        except Exception as e:
+            pass
+
+    def teardown_class(self):
```
tobegit3hub, line 347: We may delete the created table which is for testing.

python/test/openmldb_client_test.py
```diff
@@ -314,10 +316,48 @@ def test_parameterized_query(self):
...
+        try:
+            self.connection.execute(self.test_table.insert().values(x='first', y=100))
+        except Exception as e:
+            logging.warning("error occured {}".format(e))
+            assert False
+
+    def test_select(self):
+        try:
+            for row in self.connection.execute(select([self.test_table])):
+                if 'first' not in list(row):
```
tobegit3hub: assert the result directly instead of add warning log and assert.

tobegit3hub (changes requested):

python/test/openmldb_client_test.py
HuilinWu2: In Pytest's conventions, `test prefixed test functions or methods inside Test prefixed test classes (without an __init__ method)`.
it does not support the `Test` class with a `__init__` function, otherwise, it raise error like: ` cannot collect test class 'TestSqlalchemyAPI' because it has a __init__ constructor.`

python/test/openmldb_client_test.py
HuilinWu2: engine is used in `metadata.create(self.engine)`

python/test/openmldb_client_test.py
HuilinWu2: updated.

python/test/openmldb_client_test.py
HuilinWu2: updated.

tobegit3hub (approved):
LGTM

dl239 (approved):

vagetablechicken (approved):

PR #1139, opened 2022-01-24
feat: refactor to help users directly use OpenMLDB DB API

HuilinWu2:
[Python SDK] Now users are able to use OpenMLDB DB API through:

```
import openmldb as db

# get openmldb DB API
dbapi = db.dbapi

# Open database connection
connection = dbapi.connect("db_test", "127.0.0.1:6181", "/onebox")

# prepare a cursor object using cursor() method
cursor = connection.cursor()

# execute sql
cursor.execute("create database db_test;")

...

```


python/openmldb/sdk/__init__.py
```diff
@@ -11,3 +11,5 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from . import sdk
```
tobegit3hub: We may change `from sdk import *` later and update the usage in `dbapi.py` or others.

tobegit3hub (approved):
LGTM

python/openmldb/dbapi/dbapi.py
```diff
@@ -18,8 +18,8 @@
 sys.path.append("../")
 
 import logging
-from openmldb_sdk import sdk as sdk_module
-from openmldb_native import sql_router_sdk
+from ..sdk import sdk as sdk_module
+from ..native import sql_router_sdk
```
tobegit3hub: Make sure this works in linux and macos environments.

tobegit3hub:
We may update the docs, just like https://www.runoob.com/python3/python3-mysql.html and https://www.runoob.com/python/python-mysql.html .

```
import openmldb.dbapi

db = openmldb.dbapi.connect("db_test", "127.0.0.1:6181", "/onebox")

cursor = db.cursor()

cursor.execute("create database db_test;")

...
```

tobegit3hub:
Add an unit test if it is possible to use dbapi directly.

python/openmldb/sdk/__init__.py
HuilinWu2: updated.

python/openmldb/dbapi/dbapi.py
HuilinWu2: This is relative import and should work in Linux and Macos:
* If the import module in the same dir, use e.g: from . import core
* If the import module in the top dir, use e.g: from .. import core
* If the import module in the other subdir, use e.g: from ..other import core

dl239 (approved):

python/openmldb/__init__.py
```diff
@@ -13,3 +13,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from . import openmldb_dbapi
+from . import openmldb_sdk
+from . import openmldb_native
+from . import sqlalchemy_openmldb
```
vagetablechicken: add newline

python/openmldb/dbapi/__init__.py
```diff
@@ -11,3 +11,5 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from .dbapi import *
```
vagetablechicken: add newline

python/openmldb/test/sql_smoke_test.py
```diff
@@ -14,7 +14,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from openmldb_sdk import sdk as sdk_module
+import sys
```
vagetablechicken, line 18: Will it be reformat? You'd better add fmt:off

vagetablechicken (changes requested):

python/openmldb/test/sql_smoke_test.py
vagetablechicken: And `../` is not a good way, `sys.path.append(os.path.dirname(__file__) + "/..")` is better?

tobegit3hub: We don't need add "../" now for relative import. @HuilinWu2 please try to remove these.

vagetablechicken (approved):

Issue #1156, opened 2022-01-28
cppunit not found


zoyopei:
Could NOT find cppunit when build third-party from source. My cppunit library is in `${CPPUNIT_ROOT_DIR}/lib` .
`make BUILD_BUNDLED=ON THIRD_PARTY_CMAKE_FLAGS="-DCPPUNIT_ROOT_DIR=${CPPUNIT_ROOT_DIR}"`
```
-- Could NOT find cppunit (missing: CPPUNIT_LIBRARY CPPUNIT_INCLUDE_DIR)
CMake Error at cmake/FetchZookeeper.cmake:28 (message):
  zookeeper require cppunit
Call Stack (most recent call first):
  CMakeLists.txt:194 (include)
```

aceforeverd:
affected environment ? os & cppunit path

zoyopei:
> affected environment ? os & cppunit path

ubuntu18.04  CPPUNIT_ROOT_DIR=~/.linuxbrew/Cellar/cppunit/1.15.1

PR #1157, opened 2022-01-28
fix: add cppunit library search path

zoyopei:

Bug fix.

#1156 

Add subdirectories lib/ and lib64/ to search libraries.

* **Other information**:

### DIFF
--- a/third-party/cmake/Findcppunit.cmake
+++ b/third-party/cmake/Findcppunit.cmake
@@ -34,13 +34,19 @@ find_library(CPPUNIT_LIBRARY_RELEASE
 	NAMES
 	cppunit
 	HINTS
-	"${CPPUNIT_ROOT_DIR}")
+	"${CPPUNIT_ROOT_DIR}"
+    PATH_SUFFIXES
+    lib/
+    lib64/)
 
 find_library(CPPUNIT_LIBRARY_DEBUG
 	NAMES
 	cppunitd
 	HINTS
-	"${CPPUNIT_ROOT_DIR}")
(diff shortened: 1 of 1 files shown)


third-party/cmake/Findcppunit.cmake
```diff
@@ -34,13 +34,19 @@ find_library(CPPUNIT_LIBRARY_RELEASE
 	NAMES
 	cppunit
 	HINTS
-	"${CPPUNIT_ROOT_DIR}")
+	"${CPPUNIT_ROOT_DIR}"
```
aceforeverd, line 37: https://cmake.org/cmake/help/latest/command/find_library.html
> Search the paths specified by the HINTS option. These should be paths computed by system introspection, such as a hint provided by the location of another item already found. Hard-coded guesses should be specified with the PATHS option.

or should use `PATHS` instead ?

Cmake manual always hard to understand 😢

zoyopei: [https://cmake.org/pipermail/cmake/2010-October/040460.html](url)

> Long story short: Hints get searched before the system paths, and
should only be set by some source of knowledge (location of other
files, etc) and not populated with "guesses" or default locations.
Paths get searched after system locations, and is the place to put
default locations.

I have tested both HINTS and PATHS are ok.

aceforeverd:
Thanks you~. I'll confirm it for other OSs later

tobegit3hub (approved):
LGTM

aceforeverd (approved):

PR #1212, opened 2022-02-14
fix: python SDK magic register function bug fix

HuilinWu2:
Lack of an input param.


python/setup.py
```diff
@@ -28,7 +28,8 @@
         'Programming Language :: Python :: 3',
         ],
     install_requires=[
-        "sqlalchemy < 1.4.0"
+        "sqlalchemy < 1.4.0",
+        "IPython == 8.0.1"
```
tobegit3hub: It would be better to support more version. Maybe change to `IPython >= 8.0.1`.

HuilinWu2: updated. currently IPython without specific version

tobegit3hub (approved):
LGTM

vagetablechicken (approved):

PR #1215, opened 2022-02-14
feat: support function fetchmany() and fetchall() in python SDK and UT

### LINKED ISSUE #1158: Support fetchmany and fetchall in Python SDK
Just like what mysql dbapi has been supported.

image

### LINKED ISSUE #1159: Throw exception if fetchone fail in Python SDK
Now if `fetchone()` in `openmldb.dbapi` fail, it return the string which is useless and we should throw exception for this error case.

```
    def fetchone(self):
        if self._resultSet is None: return "call fetchone"
```

HuilinWu2:
Reference: [Standard DBAPI](https://www.python.org/dev/peps/pep-0249/#fetchone)
*.fetchone()
Fetch the next row of a query result set, returning a single sequence, None when no more data is available.

*.fetchmany(size=None)
Fetch the next set of rows of a query result, returning a sequence of sequences (e.g. a list of tuples). An empty sequence is returned when no more rows are available.
The number of rows to fetch per call is specified by the parameter. If it is not given, the cursor's arraysize determines the number of rows to be fetched. 

*.fetchall()
Fetch all (remaining) rows of a query result, returning them as a sequence of sequences (e.g. a list of tuples).

### DIFF
--- a/python/openmldb/dbapi/dbapi.py
+++ b/python/openmldb/dbapi/dbapi.py
@@ -343,7 +343,7 @@ class Cursor(object):
         return self.connection._sdk.getDatabases()
 
     def fetchone(self):
-        if self._resultSet is None: raise Exception("query data failed")
+        if self._resultSet is None: raise DatabaseError("query data failed")
         ok = self._resultSet.Next()
         if not ok:
             return None
@@ -357,7 +357,7 @@ class Cursor(object):
 
     @connected
     def fetchmany(self, size=None):
-        if self._resultSet is None: raise Exception("query data failed")
+        if self._resultSet is None: raise DatabaseError("query data failed")
         if size is None:
             size = self.arraysize
         elif size < 0:
--- a/python/openmldb/dbapi/dbapi.py
+++ b/python/openmldb/dbapi/dbapi.py
@@ -356,11 +356,9 @@ class Cursor(object):
         return tuple(values)
 
     @connected
-    def fetchmany(self, size=None):
+    def fetchmany(self, size=self.arraysize):
         if self._resultSet is None: raise Exception("query data failed")
-        if size is None:
-            size = self.arraysize
-        elif size < 0:
+        if size < 0:
             raise Exception(f"Given size should greater than zero")
         values = []
         for k in range(size):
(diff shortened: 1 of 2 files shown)


python/openmldb/dbapi/dbapi.py
```diff
@@ -343,14 +343,9 @@ def get_databases(self):
         return self.connection._sdk.getDatabases()
 
     def fetchone(self):
-
-        if self._resultSet is None: return "call fetchone"
+        if self._resultSet is None: raise Exception("query data failed")
```
tobegit3hub: Throw the exception which is defined in DB API. Here may be `DataError` or `DatabaseError`.

python/openmldb/dbapi/dbapi.py
```diff
@@ -375,7 +387,20 @@ def setoutputsize(self, size, columns=()):
         
     @connected
     def fetchall(self):
-        raise NotSupportedError("Unsupported in OpenMLDB")
+        if self._resultSet is None: raise Exception("query data failed")
+        values = []
+        for k in range(self.rowcount):
```
tobegit3hub: We may re-use the logic in `fetchmany`.

python/openmldb/dbapi/dbapi.py
```diff
@@ -362,7 +357,24 @@ def fetchone(self):
 
     @connected
     def fetchmany(self, size=None):
-        raise NotSupportedError("Unsupported in OpenMLDB")
+        if self._resultSet is None: raise Exception("query data failed")
+        if size is None:
```
tobegit3hub, line 361: `size` is not initialized and will be None all the time. Maybe just init with `size = self.arraysize`?

tobegit3hub (changes requested):

python/openmldb/dbapi/dbapi.py
HuilinWu2: the default args of class function does not support the class attribute. `def fetchmany(self, size = self.arraysize):`

python/openmldb/dbapi/dbapi.py
HuilinWu2: updated.

python/openmldb/dbapi/dbapi.py
HuilinWu2: updated.

python/openmldb/dbapi/dbapi.py
vagetablechicken: `arraysize` means all rows? But I only see `rowcount` updated, where to update arraysize?

HuilinWu2: for `cursor. arraysize ` is  an invariant attribute of `Cursor` object, it specifies the number of rows to fetch at a time with [.fetchmany()](https://www.python.org/dev/peps/pep-0249/#id22). If `size` of `fetchmany()` is not given, then we will initialize `size` with `cursor.arraysize`.

tobegit3hub (approved):
LGTM

vagetablechicken (approved):

Issue #1089, opened 2022-01-13
test: split_test SIGABRT when build profile is DEBUG 


aceforeverd:
## reproduce

```sh
cmake -B build -DCMAKE_BUILD_TYPE=Debug
cd build
make split_test && ./bin/split_test
```

dump log
```log
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from SplitTest
[ RUN      ] SplitTest.split_line
WARNING: Logging before InitGoogleLogging() is written to STDERR
F0113 16:32:23.367864 630285824 split.h:141] Check failed: memcmp(line, delimiter, delimiter_len) == 0 (-45 vs. 0)
*** Check failure stack trace: ***
fish: Job 1, './bin/split_test' terminated by signal SIGABRT (Abort)
```

PR #1216, opened 2022-02-14
fix: ci workflow & split_test (debug)

aceforeverd:
1. fix failed workflow
    because some tests not support parallel test via ctest currently, left to #1223
2. fix split_test, #1089 
    it turns out https://github.com/4paradigm/OpenMLDB/blob/a607e959effc809a3931aeeef1b14fad5e37ffa0/src/cmd/split.h#L141 check always fail, because e.g when `line = "abc,"`, the `line[3]` is override with `\0`
3. `api_server_test` & `tablet_impl_test` are skipped in coverage, left to #1222 
4. other minor tweaks

## Cons
slow


zhanghaohit (approved):

src/cmd/split.h
```diff
@@ -138,7 +138,6 @@ void SplitLineWithDelimiter(char* line, const char* delimiter, std::vector<char*
         // incorrect.
         const bool need_another_column = (line + delimiter_len == end_of_line);
```
vagetablechicken, line 139: `const bool need_another_column = (line + delimiter_len == end_of_line) && (memcmp(line, delimiter, delimiter_len) == 0);`

dl239 (approved):

tobegit3hub (approved):
LGTM

PR #1231, opened 2022-02-15
feat: add statement class

### LINKED ISSUE #1230: Add `execute` method to java sdk
we can use `execute` method to run the sql as below:
- create database/table
- set variables
- load data/select into
- deploy

dl239:
Add `ExecuteSQL` interface to `SQLRouter` class. Java/Python SDK can call this method to execute sql.
- If execute success status.IsOK() will be true.
- If there are some data returned, it will be packed in the ResultSet.

`std::shared_ptr<hybridse::sdk::ResultSet> ExecuteSQL(const std::string& sql,  ::hybridse::sdk::Status* status)`


dl239:
[Java sdk with new API getStatement](https://trello.com/c/oc18z9L0/26-java-sdk-with-new-api-getstatement)

src/sdk/sql_cluster_router.h
```diff
@@ -152,6 +154,10 @@ class SQLClusterRouter : public SQLRouter {
     std::shared_ptr<hybridse::sdk::ResultSet> ExecuteSQLRequest(const std::string& db, const std::string& sql,
                                                                 std::shared_ptr<SQLRequestRow> row,
                                                                 hybridse::sdk::Status* status) override;
+
+    std::shared_ptr<hybridse::sdk::ResultSet> HandleSQL(const std::string& sql,
```
tobegit3hub: Is it better to be `ExecuteSql`? We may rename `ExecuteSQL` as `QuerySql` later.

java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/Statement.java
```diff
@@ -0,0 +1,314 @@
...
+    private SQLRouter sqlRouter;
+    private boolean closed;
+
+    public Statement(SQLRouter router) {
+        this.sqlRouter = router;
+        this.closed = false;
+    }
+
+    @Override
+    public boolean execute(String sql) throws SQLException {
```
tobegit3hub, line 18: Some sql like `use db` may not return `resultSet`. Will it return to false here even if executed successfully?

java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/Statement.java
```diff
@@ -0,0 +1,314 @@
...
+        if (!status.IsOK()) {
+            String msg = status.getMsg();
+            status.delete();
+            throw new SQLException("executeSQL fail: " + msg);
+        }
+        status.delete();
+        if (rs != null) {
+            rs.delete();
+        }
+        return 0;
```
tobegit3hub, line 73: Maybe return void if the return code is not used.

java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/Statement.java
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    @Override
+    public void close() throws SQLException {
+        sqlRouter = null;
+        if (resultSet != null) {
+            resultSet.delete();
```
tobegit3hub, line 80: Should we set `resultSet = null` here?

java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/sdk/SqlExecutor.java
```diff
@@ -21,15 +21,19 @@
...
+    @Deprecated
     boolean createDB(String db);
 
+    @Deprecated
     boolean dropDB(String db);
 
+    @Deprecated
     boolean executeDDL(String db, String sql);
 
     boolean executeInsert(String db, String sql);
```
tobegit3hub, line 40: Make this as `Deprecated` as well?

tobegit3hub (changes requested):

java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/Statement.java
dl239: - if execute failed, it will throw an exception.
- true if the first result is a ResultSet object; false if there are no results

https://docs.oracle.com/javase/7/docs/api/java/sql/Statement.html#execute(java.lang.String)

java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/Statement.java
dl239: function return type is `int`,  we cannot return null. And what the mean of `return void`?

vagetablechicken: It's a defination of `java.sql.Statement.executeUpdate`, just return rowcount.

src/cmd/sql_cmd.h
```diff
@@ -1463,18 +122,19 @@ void Shell() {
...
+    std::string display_prefix = ns_endpoint + "/" + sr->GetDatabase() + "> ";
     std::string multi_line_perfix = std::string(display_prefix.length() - 3, ' ') + "-> ";
     std::string sql;
     bool multi_line = false;
     while (true) {
         std::string buffer;
         if (!FLAGS_interactive) {
             buffer = FLAGS_cmd;
-            db = FLAGS_database;
+            std::string db = FLAGS_database;
```
vagetablechicken, line 133: How about `if(!FLAGS_cmd.empty){buffer=FLAGS_cmd;}`? `FLAG_interactive` is only  a config.
So we can exec multi sql, e.g. `echo "drop table xx;\n drop table xxx;" | openmldb --sql_client --interactive=false`.

java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/Statement.java
dl239: > either (1) the row count for SQL Data Manipulation Language (DML) statements or (2) 0 for SQL statements that return nothing

`ExecuteInsert` method not support return row count now. I post an issue #1324 and will implement later

java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/Statement.java
tobegit3hub: Okay.

tobegit3hub (approved):
LGTM

vagetablechicken (approved):

PR #1243, opened 2022-02-16
feat: add print table in python sdk and show in notebook magic function

tobegit3hub:
* Add print table api in python sdk
* Print the table when using openmldb sql magic function

image


HuilinWu2 (approved):
LGTM

aceforeverd (approved):

PR #1250, opened 2022-02-17
feat: taskmanager will find the local batchjob jar if not configured

tobegit3hub:
* Remove the config of `batchjob.jar.path` with specified version in TaskManager config file
 * Support automatically finding batchjob jar with added util function
 
After merged, we do not need to update the version in TaskManager config file.

### DIFF
--- a/java/openmldb-taskmanager/src/main/java/com/_4paradigm/openmldb/taskmanager/config/TaskManagerConfig.java
+++ b/java/openmldb-taskmanager/src/main/java/com/_4paradigm/openmldb/taskmanager/config/TaskManagerConfig.java
@@ -16,6 +16,7 @@
 
 package com._4paradigm.openmldb.taskmanager.config;
 
+import com._4paradigm.openmldb.taskmanager.util.BatchJobUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import java.io.File;
@@ -68,7 +69,10 @@ public class TaskManagerConfig {
             ZK_MAX_CONNECT_WAIT_TIME = Integer.parseInt(prop.getProperty("zookeeper.max_connect_waitTime", "30000"));
             OFFLINE_DATA_PREFIX = prop.getProperty("offline.data.prefix");
             SPARK_MASTER = prop.getProperty("spark.master", "yarn");
-            BATCHJOB_JAR_PATH = prop.getProperty("batchjob.jar.path");
+            BATCHJOB_JAR_PATH = prop.getProperty("batchjob.jar.path", "");
+            if (BATCHJOB_JAR_PATH.isEmpty()) {
+                BATCHJOB_JAR_PATH = BatchJobUtil.findLocalBatchJobJar();
+            }
             SPARK_YARN_JARS = prop.getProperty("spark.yarn.jars");
             SPARK_HOME = prop.getProperty("spark.home");
             PREFETCH_JOBID_NUM = Integer.parseInt(prop.getProperty("prefetch.jobid.num", "10"));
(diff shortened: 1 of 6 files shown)


java/openmldb-taskmanager/src/main/scala/com/_4paradigm/openmldb/taskmanager/util/BatchJobUtil.scala
```diff
@@ -0,0 +1,46 @@
...
+import java.io.File
+
+object BatchJobUtil {
+
+  /**
+   * Get the default batch jor from presupposed directories.
+   *
+   * @return the path of openmldb-batchjob jar
+   */
+  def findLocalBatchJobJar(): String = {
```
aceforeverd, line 13: what's the working directory for those funs ?

java/openmldb-taskmanager/src/main/scala/com/_4paradigm/openmldb/taskmanager/util/BatchJobUtil.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  /**
+   * Find the openmldb-batchjob jar from specified directory.
+   *
+   * @param libDirectory the directory to check
+   * @return
+   */
+  def findBatchJobJar(libDirectory: String): String = {
+    val libDirectoryFile = new File(libDirectory)
```
aceforeverd, line 32: will there any Exceptions ? e.g `FileNotFoundException`, `IOException`

java/openmldb-taskmanager/src/main/scala/com/_4paradigm/openmldb/taskmanager/util/BatchJobUtil.scala
tobegit3hub: It would be the directory of shell script if we use the pre-built package. Then we will check for `../jars/` by default.

It would be the directory of maven parent-module  if we run in IDE. Then we will check for `./openmldb-batchjob/target/` by default.

java/openmldb-taskmanager/src/main/scala/com/_4paradigm/openmldb/taskmanager/util/BatchJobUtil.scala
tobegit3hub: `IOException` could be better. But we may unified all the exceptions later.

java/openmldb-taskmanager/src/main/java/com/_4paradigm/openmldb/taskmanager/config/TaskManagerConfig.java
```diff
@@ -68,7 +69,10 @@
             ZK_MAX_CONNECT_WAIT_TIME = Integer.parseInt(prop.getProperty("zookeeper.max_connect_waitTime", "30000"));
             OFFLINE_DATA_PREFIX = prop.getProperty("offline.data.prefix");
             SPARK_MASTER = prop.getProperty("spark.master", "yarn");
-            BATCHJOB_JAR_PATH = prop.getProperty("batchjob.jar.path");
+            BATCHJOB_JAR_PATH = prop.getProperty("batchjob.jar.path", "");
+            if (BATCHJOB_JAR_PATH.isEmpty()) {
+                BATCHJOB_JAR_PATH = BatchJobUtil.findLocalBatchJobJar();
```
aceforeverd, line 74: need test coverage

tobegit3hub: We will add check for config of TaskManager in this issue https://github.com/4paradigm/OpenMLDB/issues/1252 .

aceforeverd (approved):

HuilinWu2 (approved):
LGTM

PR #1253, opened 2022-02-18
feat: unittest for sql magic in Python SDK

HuilinWu2:


python/openmldb/test/openmldb_client_test.py
```diff
@@ -395,6 +395,55 @@ def test_drop_table(self):
...
+    def setup_class(self):
+        self.db = openmldb.dbapi.connect('db_test', '127.0.0.1:6181', '/onebox')
+        self.ip = openmldb.sql_magic.register(self.db)
+
+    def execute(self,magic_name,line):
+        try:
+            self.ip.run_line_magic(magic_name,line)
+            return 'ok'
+        except Exception as e:
+            raise Exception(e)
```
tobegit3hub: If we just catch exception and raise, you can remove these lines and do not catch the exception.

python/openmldb/test/openmldb_client_test.py
```diff
@@ -395,6 +395,55 @@ def test_drop_table(self):
...
     def teardown_class(self):
         self.cursor.close()
+        
+class TestSQLMagicOpenMLDB:
+
+    def setup_class(self):
+        self.db = openmldb.dbapi.connect('db_test', '127.0.0.1:6181', '/onebox')
+        self.ip = openmldb.sql_magic.register(self.db)
+
+    def execute(self,magic_name,line):
```
tobegit3hub: Since our magic function is always "sql", you can remove the this parameter and only pass the `line`.

The parameter `line` would change to be `sql` or `code` which is more meaningful.

python/openmldb/test/openmldb_client_test.py
```diff
@@ -395,6 +395,55 @@ def test_drop_table(self):
...
+        try:
+            self.ip.run_cell_magic('sql','', "create table magic_table (x string, y int);")
+        except Exception as e:
+            assert False
+        assert "magic_table" in self.db.cursor().get_all_tables()
+
+        with pytest.raises(Exception):
+            assert self.execute('sql', "create table magic_table;")
+
+    def test_magic_insert(self):
```
tobegit3hub: Since the name of the test class is `TestSQLMagicOpenMLDB`. The names of method could be simpler, like `test_insert`.

tobegit3hub (changes requested):

python/openmldb/sql_magic/sql_magic.py
```diff
@@ -16,6 +16,7 @@
 from __future__ import print_function
 from IPython.core.magic import (Magics, magics_class, line_magic,
                                 cell_magic, line_cell_magic)
+from IPython.testing.globalipapp import get_ipython
```
tobegit3hub: This may be only used in test environment. Please check in non-testing cases.

python/openmldb/test/openmldb_client_test.py
```diff
@@ -395,6 +395,55 @@ def test_drop_table(self):
...
+class TestSQLMagicOpenMLDB:
+
+    def setup_class(self):
+        self.db = openmldb.dbapi.connect('db_test', '127.0.0.1:6181', '/onebox')
+        self.ip = openmldb.sql_magic.register(self.db)
+
+    def execute(self,magic_name,sql):
+        try:
+            self.ip.run_line_magic(magic_name,sql)
+            return 'ok'
```
tobegit3hub: Returning bool may be better than string. And you use `assert self.execute()` later which requires boolean.

python/openmldb/test/openmldb_client_test.py
```diff
@@ -395,6 +395,55 @@ def test_drop_table(self):
...
+        except Exception as e:
+            assert False
+    
+    def test_drop(self):
+        try:
+            self.ip.run_line_magic('sql', "drop table magic_table;")
+        except Exception as e:
+            assert False
+            
+        assert "magic_table" not in self.db.cursor().get_all_tables()
```
tobegit3hub, line 446: We may add function `has_table()` later instead of test like this.

tobegit3hub (changes requested):

tobegit3hub (approved):
LGTM

dl239 (approved):

PR #1267, opened 2022-02-18
feat: openmldb batch do not crash when connect openmldb fail

### LINKED ISSUE #1264: ZK is required to set even if running simple SQL without reading and write openmldb cluster
If ZK is not set correctly and run `select 1` in offline mode, it fails with the following log. Maybe we can support these SQL without reading or write openmldb cluster.

image

tobegit3hub:
* Resolve https://github.com/4paradigm/OpenMLDB/issues/1264


tobegit3hub:
Now we can run `select 1` after `SET @@SESSION.execute_mode = "offline"` without connecting openmldb cluster in yarn.

dl239 (approved):

vagetablechicken (approved):

PR #1268, opened 2022-02-19
fix: cli command will not be executed if we add space after semicolon

### LINKED ISSUE #1121: CLI command will not be executed if we add space after semicolon
image
In sql_client, if we add space after semicolon and ENTER, command will not be executed

keyu813:
close #1121
image
only delete spaces after last semicolon if exist

### DIFF
--- a/src/cmd/sql_cmd.h
+++ b/src/cmd/sql_cmd.h
@@ -158,10 +158,13 @@ void Shell() {
             std::string::size_type buffer_size = buffer.size();
             std::string after_semicolon_sql = buffer.substr(last_semicolon + 1, buffer_size);
             absl::string_view input(after_semicolon_sql);
-            while (!absl::ConsumePrefix(&input, " ")) {
-                buffer = buffer.substr(0, last_semicolon + 1);
-                buffer = buffer + after_semicolon_sql;
+            while (true) {
+                if (!absl::ConsumePrefix(&input, " ")) {
+                    break;
+                }
             }
+            buffer = buffer.substr(0, last_semicolon + 1);
+            buffer = buffer + std::string(input);
         }
         sql.append(buffer);
         if (sql.length() == 4 || sql.length() == 5) {
--- a/src/cmd/sql_cmd.h
+++ b/src/cmd/sql_cmd.h
@@ -152,7 +152,7 @@ void Shell() {
                 continue;
             }
         }
-        // todo: should support multiple sql
+        // todo: should support multiple sql.
         // trim space after last semicolon in sql
         auto last_semicolon_pos = buffer.find_last_of(';');
         if (last_semicolon_pos != std::string::npos && buffer.back() != ';') {
--- a/src/cmd/sql_cmd.h
+++ b/src/cmd/sql_cmd.h
@@ -151,6 +151,18 @@ void Shell() {
                 continue;
             }
         }
+        if (buffer.find_last_of(';') != std::string::npos && buffer.back() != ';') {
+            std::string::size_type last_semicolon = buffer.find_last_of(';');
+            std::string::size_type buffer_size = buffer.size();
+            std::string after_semicolon_sql = buffer.substr(last_semicolon + 1, buffer_size);
+            boost::trim(after_semicolon_sql);
+            if (after_semicolon_sql.empty()) {
+                buffer = buffer.substr(0, last_semicolon + 1);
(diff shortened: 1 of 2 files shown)


tobegit3hub:
Is this compatible with other databases? MySQL command client may execute event if it ends with spaces;

image

keyu813:
> Is this compatible with other databases? MySQL command client may execute event if it ends with spaces;
> 
> image

Mysql support execute multiple commands (split by semicolon) at once, so "show databases; ;" will be explained as "show databases;" and " ;". Execution will get the result of "show databases" as well as Error(result of ";"). Openmldb's sql is a little different from Mysql, openmldb support execute a command at once temporarily. So I consider exit when there are multiple semicolons.

src/cmd/sql_cmd.h
```diff
@@ -1469,6 +1468,18 @@ void Shell() {
                 continue;
             }
         }
+        if (buffer.find_last_of(';') != std::string::npos && buffer.back() != ';') {
+            std::string::size_type last_semicolon = buffer.find_last_of(';');
+            std::string::size_type buffer_size = buffer.size();
+            std::string after_semicolon_sql = buffer.substr(last_semicolon + 1, buffer_size);
+            boost::trim(after_semicolon_sql);
+            if (after_semicolon_sql.empty()) {
```
aceforeverd: write tests to cover diff codes

tobegit3hub:
@keyu813  We may support multiple commands in OpenMLDB soon. So "exit when there are multiple semicolons" seems not an good idea if we support multiple commands later.

src/cmd/sql_cmd.h
```diff
@@ -1469,6 +1468,18 @@ void Shell() {
                 continue;
             }
         }
+        if (buffer.find_last_of(';') != std::string::npos && buffer.back() != ';') {
+            std::string::size_type last_semicolon = buffer.find_last_of(';');
+            std::string::size_type buffer_size = buffer.size();
+            std::string after_semicolon_sql = buffer.substr(last_semicolon + 1, buffer_size);
+            boost::trim(after_semicolon_sql);
```
aceforeverd: avoid use boost, have a look at abseil https://github.com/abseil/abseil-cpp/blob/master/absl/strings/strip.h

keyu813:
> @keyu813 We may support multiple commands in OpenMLDB soon. So "exit when there are multiple semicolons" seems not an good idea if we support multiple commands later.

ok, I will consider other solutions

src/cmd/sql_cmd.h
```diff
@@ -1469,6 +1468,18 @@ void Shell() {
                 continue;
             }
         }
+        if (buffer.find_last_of(';') != std::string::npos && buffer.back() != ';') {
```
aceforeverd: what's the use of semicolon ? 

so you'd like to cover the case like `show databases; show ` ? I checked mysql but they treat the second `show` as unterminated.

src/cmd/sql_cmd.h
keyu813: ok I'll add case

src/cmd/sql_cmd.h
keyu813: ok

src/cmd/sql_cmd.h
keyu813: I think mysql execute "show databases" and buffer "show" and wait for command, mysql support multiple commands. Openmldb support single command temporarily. So I will treat "show databases; show" as an illegal command. "Show databases; show databases;" will be consider as single command and parse failed. "Show database;      " will be executed.

vagetablechicken (changes requested):
test `create procedure`

keyu813:
> test `create procedure`

```
create procedure sp_auto_pro5(c1 string,c3 int,c7 timestamp)
        BEGIN
        SELECT * FROM t1 ; 
        END;
```
will failed (sql client can't read the whole sql because it will terminate when reading the line that ends with semicolon), but 
```
create procedure sp_auto_pro5(c1 string,c3 int,c7 timestamp)
        BEGIN
        SELECT * FROM t1 
        ;END;
```
can execute.

aceforeverd:
How about the status now? @keyu813

dl239 (approved):

vagetablechicken (approved):

PR #1276, opened 2022-02-21
fix: detect incorrect sql like `select 10s` or `select 10s + 1`

### LINKED ISSUE #1208: Incorrect SQL like `SELECT 100ss` or `select 100s` can be run
Now we can run SQL like `SELECT 100ss` in CLI which may output unexcepted result. Here is the execution and result.

Screen Shot 2022-02-12 at 7 56 42 PM

aceforeverd:
close #1208

reason: `ConstNode` 's `ShadowCopy` not cover for time interval. The patch fix `ShadowCopy` method for time interval.
After patch the original infer attr flow for `select 10s` or `select 10s + 1` work as expected.

Other changes:

- rm default parameter for `PhysicalOpNode::SchemaToString`, not recommended for virtual method

### DIFF
--- a/cases/query/const_query.yaml
+++ b/cases/query/const_query.yaml
@@ -133,4 +133,4 @@ cases:
     expect:
       columns: ['c1 bool', 'c2 int16', 'c3 int', 'c4 double', 'c5 string', 'c6 date', 'c7 timestamp' ]
       rows:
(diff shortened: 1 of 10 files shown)


tobegit3hub (approved):
LGTM

jingchen2222 (approved):
LGTM

Issue #1309, opened 2022-02-24
Fail to register notebook magic function with Python SDK


tobegit3hub:
With latest code, we will fail when register the notebook magic function. Here is the error message.

![企业微信截图_6946022b-d027-43bd-8d51-e9df9de5cb7e](https://user-images.githubusercontent.com/2715000/155507015-fb1e82a1-b5fa-4cf8-9d87-264dadfa8133.png)

Comment out the test code will work again.

<img width="895" alt="wecom-temp-9e9772f701c9c8b2c904c8d10664a881" src="https://user-images.githubusercontent.com/2715000/155507082-421114aa-f0c9-450a-baa8-ed4cba0d6e86.png">

HuilinWu2:
![](https://github.trello.services/images/mini-trello-icon.png) [Improve functionality of Python SDK](https://trello.com/c/xlXZPfTf/20-improve-functionality-of-python-sdk)

PR #1311, opened 2022-02-24
fix: bugfix fail to register notebook magic with python SDK

HuilinWu2:
see issue https://github.com/4paradigm/OpenMLDB/issues/1309


tobegit3hub (commented):
LGTM and tested in MacOS.

PR #1340, opened 2022-02-28
fix: reset the df schema after read to fix the "not null" issue when loading data

### LINKED ISSUE #1271: online load: reading data of "not null" columns in Spark causes data load failed
related to #1170 
Why the error is "write null to non-null column `device_id` failed"? Not the column `std_time`?
**We can set any nullable column to `null`, it will got the same error.** code version ef2e6e826
e.g.
```
device_id,event_type,ability_id,category_first_id,std_time,std_date,testno
aaa,a,a,100,2022-02-07 20:54:20,2022-02-01,null
```

We should check about spark read df. `openmldb-spark-connector` has ut about insert null, it's ok, so the source df may have some error.

vagetablechicken:
Closes #1271


tobegit3hub (approved):
LGTM

dl239 (approved):

vagetablechicken:
[improvement for CSV import](https://trello.com/c/eSTnTgIY/13-improvement-for-csv-import)

PR #1354, opened 2022-03-01
feat: update the taskmanager config file in release

tobegit3hub:
* We have new taskmanager config which updates the default one in release.


dl239:
if sometime modify the conf in java/openmldb-taskmanager, it may be  inconsistent with the conf in release dir.
It's better to copy the file from java/openmldb-taskmanager automatically.

dl239 (approved):

vagetablechicken (approved):

PR #1380, opened 2022-03-04
feat: impl show components

### LINKED ISSUE #1232: feat: support sql `SHOW COMPONENTS`
`SHOW COMPONENTS` prints all components registered in the cluster, which include 
- endpoint
- role: tablet/name server/api server/task manager
- connect_time
- status : healthy or offline

for tablet there is healthy/offline status, for name server /api server, there is jut healthy, offline components does not exists in the list

aceforeverd:
close #1232

A correct example:
Screen Shot 2022-03-11 at 10 59 46 AM

### DIFF
--- a/src/sdk/sql_cluster_router.cc
+++ b/src/sdk/sql_cluster_router.cc
@@ -3012,7 +3012,7 @@ static const std::initializer_list<std::string> GetComponetSchema() {
 // where
 // - ENDPOINT: IP:PORT or DOMAIN:PORT
 // - ROLE can be 'tablet', 'nameserver', 'taskmanager'
-// - CONNECT_TIME uptime time in milliseconds from epoch
+// - CONNECT_TIME last conncted timestamp from epoch
 // - STATUS can be 'online', 'offline' or 'NULL' (otherwise)
 // - NS_ROLE can be 'master', 'standby', or 'NULL' (for non-namespace component)
 std::shared_ptr<hybridse::sdk::ResultSet> SQLClusterRouter::ExecuteShowComponents(hybridse::sdk::Status* status) {
@@ -3109,7 +3109,7 @@ std::shared_ptr<hybridse::sdk::ResultSet> SQLClusterRouter::ExecuteShowNameServe
         data[i][1] = "nameserver";
 
         // connect time
-        data[i][2] = std::to_string(::baidu::common::timer::get_micros() / 1000 - it->second);
+        data[i][2] = std::to_string(it->second);
 
         // status
         // offlined nameserver won't register in zookeeper, so there is only online
@@ -3147,8 +3147,8 @@ std::shared_ptr<hybridse::sdk::ResultSet> SQLClusterRouter::ExecuteShowTablets(h
         data[i][0] = tablets[i].endpoint;  // endpoint
         data[i][1] = "tablet";             // role
 
-        // connect time in milliseconds
-        data[i][2] = std::to_string(tablets[i].age);
+        // connecct time
+        data[i][2] = std::to_string(::baidu::common::timer::get_micros() / 1000 - tablets[i].age);
 
         // state
         if (tablets[i].state == "kHealthy") {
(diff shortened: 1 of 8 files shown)


src/sdk/sql_cluster_router.cc
```diff
@@ -2960,5 +2990,189 @@ bool SQLClusterRouter::CheckPreAggrTableExist(const std::string& base_table, con
...
+        data[i][3] = std::to_string(it->second);
+
+        // status
+        // offlined nameserver won't register in zookeeper, so there is only online
+        data[i][4] = "online";
+
+        // ns_role
+        // NSs runs as mater/standby mode
+        if (i == 0) {
+            data[i][5] = "leader";
```
aceforeverd: master

src/sdk/sql_cluster_router.cc
```diff
@@ -2960,5 +2990,189 @@ bool SQLClusterRouter::CheckPreAggrTableExist(const std::string& base_table, con
...
+        Stat stat;
+        if (!zk_client->GetNodeValueAndStat(real_path.c_str(), &endpoint, &stat)) {
+            status->code = hybridse::common::kRunError;
+            status->msg = absl::StrCat("get endpoint failed for path: ", real_path);
+            return {};
+        }
+        if (endpoint_set.find(endpoint) == endpoint_set.end()) {
+            endpoint_set[endpoint] = stat.ctime;
+        } else {
+            endpoint_set[endpoint] = std::min(stat.ctime, endpoint_set[endpoint]);
```
aceforeverd: `ctime` seems inaccurate.
Capture

aceforeverd:
a offline node's connect time do not reset to zero:
Capture

src/sdk/sql_cluster_router.cc
```diff
@@ -1405,7 +1441,8 @@ std::shared_ptr<hybridse::sdk::ResultSet> SQLClusterRouter::HandleSQLCmd(
                 return {};
             }
             auto tables = cluster_sdk_->GetTables(db);
-            std::vector<std::vector<std::string>> values;;
+            std::vector<std::vector<std::string>> values;
+            ;
```
vagetablechicken: only one semicolon?

vagetablechicken:
> a offline node's connect time do not reset to zero: Capture

Can we make the `connect time` more readable? `std::gmtime` or other way?

aceforeverd:
> > a offline node's connect time do not reset to zero: Capture
> 
> Can we make the `connect time` more readable? `std::gmtime` or other way?

the prometheus will going to parse the time, so I guess we should do some hack in the cli ?

vagetablechicken:
> > > a offline node's connect time do not reset to zero: Capture
> > 
> > 
> > Can we make the `connect time` more readable? `std::gmtime` or other way?
> 
> the prometheus will going to parse the time, so I guess we should do some hack in the cli ?

Yes, just print, no need to change the type.

aceforeverd:
> > > > a offline node's connect time do not reset to zero: Capture
> > > 
> > > 
> > > Can we make the `connect time` more readable? `std::gmtime` or other way?
> > 
> > 
> > the prometheus will going to parse the time, so I guess we should do some hack in the cli ?
> 
> Yes, just print, no need to change the type.

Well I may misunderstood. `CONNECT_TIME` here is actually impled as `connected_time` (time from now to first  connect time), which is a time interval. I think it might better to impl it as `CONNECT_TIME` , which is the timestamp. For timestamp, we already pretty-printed in the console

vagetablechicken:
So the `connected_time` is just xx microseconds interval?  Shall we convert it to xx min or hour

aceforeverd:
> So the `connected_time` is just xx microseconds interval? Shall we convert it to xx min or hour

Yes, now I fixed it and it is now regarded as timestamp. The `ResultSetSQL::MakeResultSet` do not support put a timestamp or int into it directly, so I'll fix pretty-print issue later

vagetablechicken (approved):

dl239 (approved):

PR #1392, opened 2022-03-07
chore: enhance like predicate

aceforeverd:
- build: resolve compile on darwin by move `CMAKE_PREFIX_PATH` before `find_package`s
- add a explicit constructor for `Nullable<StringRef>(const char*)`, so it is possialb to write `Nullable<StringRef>` as 
 `"abc"` instead `StringRef("abc")`, which may help reduce key strokes
- chore(codegen): fix a arg name for like expr & add a few ExprIR test for LIKE predicate

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,10 @@ if (MAC_TABLET_ENABLE)
     add_compile_definitions(__mac_tablet__=1)
 endif ()
 
+if (NOT DEFINED CMAKE_PREFIX_PATH)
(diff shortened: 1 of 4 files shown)


zhanghaohit (approved):
LGTM

tobegit3hub (commented):
LGTM

aceforeverd:
[bugfix-liguo](https://trello.com/c/Ut8UamBY/37-bugfix-liguo)

vagetablechicken (approved):

PR #1429, opened 2022-03-11
feat: check batch insert result in spark connector

### LINKED ISSUE #1329: spark connector: should handle the returns after write
https://github.com/4paradigm/OpenMLDB/blob/1aecdbc7f3bffe62373fecdab158f9f6611d9775/java/openmldb-spark-connector/src/main/java/com/_4paradigm/openmldb/spark/write/OpenmldbDataWriter.java#L130

`executeBatch` will return a list of int, 0 means ok, -1 means insert failed.
Read the list, if not all 0, throw an exception.

cswjd:
feat(1329): add preparedStatement.executeBatch() to throw an insert failure exception when the execution result is abnormal

### DIFF
--- a/java/openmldb-spark-connector/src/main/java/com/_4paradigm/openmldb/spark/write/OpenmldbDataWriter.java
+++ b/java/openmldb-spark-connector/src/main/java/com/_4paradigm/openmldb/spark/write/OpenmldbDataWriter.java
@@ -131,7 +131,7 @@ public class OpenmldbDataWriter implements DataWriter<InternalRow> {
             for(int i = 0; i < rc.length; i++) {
                 int code = rc[i];
                 if (code < 0) {
-                    throw new SQLException("List of " + i + " insert failed. Response code: " + code);
+                    throw new SQLException("insert failed. result code: " + code);
                 } 
             }
             preparedStatement.close();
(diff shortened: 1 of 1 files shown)


java/openmldb-spark-connector/src/main/java/com/_4paradigm/openmldb/spark/write/OpenmldbDataWriter.java
```diff
@@ -127,7 +127,13 @@ private void addRow(InternalRow record, PreparedStatement preparedStatement) thr
     @Override
     public WriterCommitMessage commit() throws IOException {
         try {
-            preparedStatement.executeBatch();
+            int[] rc = preparedStatement.executeBatch();
+            for(int i = 0; i < rc.length; i++) {
+                int code = rc[i];
+                if (code < 0) {
+                    throw new SQLException("List of " + i + " insert failed. Response code: " + code);
```
vagetablechicken: About the execption message, `i` is useless, we can't find which row is the i th row. And `rc[i]` is just a int, not a code.
Count how many elem in rc is -1, report `insert X rows, Y rows failed`.

cswjd: ok

dl239 (approved):

vagetablechicken (approved):

PR #1437, opened 2022-03-14
fix(#1410): log error and return at the end of switch statement in `ShadowCopy` function

### LINKED ISSUE #1410: Add `return` after the `switch` statement
WX20220310-093747

It needs to add `return` after the `switch` statement in `ShadowCopy` function. Please log error message such as 'Unsupported Data type' and return nullptr;

TheFu527:

Bug fix

https://github.com/4paradigm/OpenMLDB/issues/1410

Add default to the switch statement in ShadowCopy function. Log error message 'Unsupported Data type' and return nullptr.

No.

* **Other information**:

### DIFF
--- a/hybridse/src/node/expr_node.cc
+++ b/hybridse/src/node/expr_node.cc
@@ -918,6 +918,10 @@ ConstNode* ConstNode::ShadowCopy(NodeManager* nm) const {
             LOG(WARNING) << "Fail to copy primary expr of type " << node::DataTypeName(GetDataType());
             return nm->MakeConstNode(GetDataType());
         }
+        
+        default:
+            LOG(ERROR) << "Unsupported Data type";
+            return nullptr;
     }
 }
 
--- a/hybridse/src/node/expr_node.cc
+++ b/hybridse/src/node/expr_node.cc
@@ -918,9 +918,9 @@ ConstNode* ConstNode::ShadowCopy(NodeManager* nm) const {
             LOG(WARNING) << "Fail to copy primary expr of type " << node::DataTypeName(GetDataType());
             return nm->MakeConstNode(GetDataType());
         }
-        
+
         default:
-            LOG(ERROR) << "Unsupported Data type";
+            LOG(ERROR) << "Unsupported Data type " << node::DataTypeName(GetDataType());
             return nullptr;
     }
 }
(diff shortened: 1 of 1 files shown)


hybridse/src/node/expr_node.cc
```diff
@@ -918,6 +918,10 @@ ConstNode* ConstNode::ShadowCopy(NodeManager* nm) const {
             LOG(WARNING) << "Fail to copy primary expr of type " << node::DataTypeName(GetDataType());
             return nm->MakeConstNode(GetDataType());
         }
+        
+        default:
+            LOG(ERROR) << "Unsupported Data type";
```
dl239: Append the current data type to log info
`LOG(ERROR) << "Unsupported Data type " << node::DataTypeName(GetDataType())`

hybridse/src/node/expr_node.cc
```diff
@@ -918,6 +918,10 @@ ConstNode* ConstNode::ShadowCopy(NodeManager* nm) const {
             LOG(WARNING) << "Fail to copy primary expr of type " << node::DataTypeName(GetDataType());
             return nm->MakeConstNode(GetDataType());
         }
+
```
dl239: please resolve the cpplint warning

hybridse/src/node/expr_node.cc
```diff
@@ -918,6 +918,10 @@ ConstNode* ConstNode::ShadowCopy(NodeManager* nm) const {
...
+        default:
```
aceforeverd: `default` prevent static analyzer generate error if any new Datatype added. Instead, put those simply after switch block

TheFu527: So nothing should be done in the default case?

dl239: hi @TheFu527 ,  the advices of @aceforeverd is to remove the `default` statement and add logging and return at the end of switch block as follows:
switch: {
     case xxx
}
LOG xxxx;
return nullptr;

hybridse/src/node/expr_node.cc
```diff
@@ -918,7 +918,12 @@ ConstNode* ConstNode::ShadowCopy(NodeManager* nm) const {
             LOG(WARNING) << "Fail to copy primary expr of type " << node::DataTypeName(GetDataType());
             return nm->MakeConstNode(GetDataType());
         }
+
+        default: {
```
aceforeverd: just rm default, line 925-926 is enough

dl239 (approved):

aceforeverd (approved):

PR #1471, opened 2022-03-17
feat: support init global table and support recover table when tablet restart

keyu813:
1. support init global system table when nameserver init
2. support recover system table when tablet restart
3. add ut

### DIFF
--- a/src/nameserver/name_server_impl.cc
+++ b/src/nameserver/name_server_impl.cc
@@ -34,6 +34,7 @@
 #include "gflags/gflags.h"
 #include "schema/index_util.h"
 #include "schema/schema_adapter.h"
+#include "codec/row_codec.h"
 
 DECLARE_string(endpoint);
 DECLARE_string(zk_cluster);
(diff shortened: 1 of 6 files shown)


src/nameserver/name_server_impl.cc
```diff
@@ -10202,6 +10219,58 @@ base::Status NameServerImpl::CreateSystemTable(const std::string& table_name, Sy
...
+        dimensions.push_back(std::make_pair(iter->first, 0));
+        rows_dimensions.push_back(dimensions);
+    }
+    // insert value
+    uint32_t tid = table_info->tid();
+    uint32_t pid_num = table_info->table_partition_size();
+    for (int i = 0; i < default_value.size(); i++) {
+        std::string row = rows[i];
+        std::vector<std::pair<std::string, uint32_t>> dimensions = rows_dimensions[i];
+        uint32_t pid;
```
dl239: `uint32_t pid = 0;`

src/nameserver/name_server_impl.cc
```diff
@@ -1278,6 +1279,19 @@ void NameServerImpl::RecoverEndpointInternal(const std::string& endpoint, bool n
     for (const auto& kv : db_table_info_) {
         RecoverEndpointDBInternal(endpoint, need_restore, concurrency, kv.second);
     }
+    // recover global variable after tablet restart
+    NotifyGlobalVarChanged();
```
dl239: if global_var table not reside in this endpoint, notify here. otherwise notify in `UpdateZkTableNode`  and judge global_var table

PR #1472, opened 2022-03-17
feat: pre aggregator

### LINKED ISSUE #1202: Basic Pre-aggregator Design and Implementation
sub-task of long-time window feature

- Upon `Put` request, we check if there are associated pre-aggregators.
- If Yes, we update these pre-aggregators

Pre-aggregator update steps:
- update its internal states
- if the current mini-window is full/complete, close the current mini-window and insert the new pre-aggregated mini-window result into the pre-aggregation table
- if the new tuple is out of the boundary of the current mini-window, re-read the corresponding entry, update and put back

*As first attempt, we only consider one op: sum*

**Additional context**
long-time window feature

This issue depends on Issue #1197 and #1217

For this issue, we only consider the normal data insert.
For the scenario where there are already some data in the table when issuing deploy, we will consider in another issue #1204

nautaa:
feature.

close #1202 

**aggregator design**
The aggregator will be created when we specify `long_windows=xx` in the deploy statement. 
We will create the same number of aggregators as the base table partitions for each aggregation column. That is to say, the number of aggregators is bound to the number of base table partitions.
When a base table partition receives a put request, it determines whether to write to the aggregator by scanning all aggregators bound to the table partition for the same index as the put request.
Since the base table partition is in a one-to-one correspondence with the aggregator and the aggregation table partition. Therefore, under normal circumstances, the aggregated value after aggregation by the aggregator will not generate cross-tablet writes.

Aggregator support two window type: `rows` and `rows_range`.

When out-of-order writes occur, the aggregator seeks pre_aggregator table to update.


nautaa:

src/sdk/sql_cluster_test.cc
```diff
@@ -504,6 +504,74 @@ TEST_F(SQLClusterTest, CreatePreAggrTable) {
...
+    for (int i = 5; i >= 1; i--) {
+        ASSERT_TRUE(rs->Next());
+        ASSERT_EQ("str1|str2", rs->GetStringUnsafe(0));
+        ASSERT_EQ(i * 2 - 1, rs->GetInt64Unsafe(1));
+        ASSERT_EQ(i * 2, rs->GetInt64Unsafe(2));
+        ASSERT_EQ(2, rs->GetInt32Unsafe(3));
+        std::string aggr_val_str = rs->GetStringUnsafe(4);
+        int64_t aggr_val = *reinterpret_cast<int64_t*>(&aggr_val_str[0]);
+        ASSERT_EQ(i * 4 - 1, aggr_val);
+        ASSERT_EQ(i * 2, rs->GetInt64Unsafe(5));
```
zhanghaohit, line 562: binlog_offset is starting from 2?

src/storage/aggregator.cc
```diff
@@ -0,0 +1,319 @@
...
+    auto& aggr_buffer = aggr_buffer_map_.at(key);
+    if (window_type_ == WindowType::kRowsRange && cur_ts > aggr_buffer.ts_end_) {
+        FlushAggrBuffer(key, aggr_buffer);
+    }
+    if (aggr_buffer.aggr_cnt_ == 0) {
+        aggr_buffer.ts_begin_ = cur_ts;
+        if (window_type_ == WindowType::kRowsRange) {
+            aggr_buffer.ts_end_ = cur_ts + window_size_ - 1;
+        }
+    }
```
zhanghaohit, line 97: why not put this `init` steps and the `init` steps in the `FlushAggrBuffer` together? 

One issue is `ts_begin_` shouldn't be set multiple times. It's safe to set `ts_begin_` to last `ts_end_` + 1. `cur_ts` may be much larger than last `ts_end_` + 1 and there shouldn't be a gap between neighbor mini-windows. 
Otherwise you have to update the `ts_begin_` or `ts_end_` if the out-of-out case's ts is not in the range of [`ts_begin_`, `ts_end_`]

src/storage/aggregator.cc
```diff
@@ -0,0 +1,319 @@
...
+        }
+        default: {
+            PDLOG(WARNING, "unsupported data type");
+            return false;
+        }
+    }
+    return true;
+}
+
+bool SumAggregator::UpdateAggrVal(codec::RowView* row_view, AggrBuffer* aggr_buffer) {
```
zhanghaohit: can merge the two `UpdateAggrVal`? at least for the calculation logics?

src/storage/aggregator.cc
```diff
@@ -0,0 +1,319 @@
...
+    if (window_type_ == WindowType::kRowsNum && aggr_buffer.aggr_cnt_ >= window_size_) {
+        FlushAggrBuffer(key, aggr_buffer);
+    }
+
+    return true;
+}
+
+void Aggregator::FlushAggrBuffer(const std::string& key, const AggrBuffer& buffer) {
+    std::string encoded_row;
+    codec::RowBuilder row_builder(aggr_table_schema_);
```
nautaa: reuse

src/storage/aggregator.cc
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    // handle the case that the current timestamp is smaller than the begin timestamp in aggregate buffer
+    if (cur_ts < aggr_buffer.ts_begin_) {
+        auto it = aggr_table_->NewTraverseIterator(0);
+        it->Seek(key, cur_ts);
+        if (it->Valid()) {
+            auto val = it->GetValue();
+            std::string origin_data = val.ToString();
+            codec::RowView origin_row_view(aggr_table_schema_,
```
nautaa: reuse

src/storage/aggregator.cc
```diff
@@ -0,0 +1,319 @@
...
+    }
+    if (aggr_buffer_map_.find(key) == aggr_buffer_map_.end()) {
+        aggr_buffer_map_.emplace(key, AggrBuffer{});
+    }
+    auto& aggr_buffer = aggr_buffer_map_.at(key);
+    if (window_type_ == WindowType::kRowsRange && cur_ts > aggr_buffer.ts_end_) {
+        FlushAggrBuffer(key, aggr_buffer);
+    }
+    if (aggr_buffer.aggr_cnt_ == 0) {
+        aggr_buffer.ts_begin_ = cur_ts;
```
nautaa, line 93: time  init

src/storage/aggregator.cc
```diff
@@ -0,0 +1,319 @@
...
+#include "base/glog_wapper.h"
+#include "base/strings.h"
+#include "common/timer.h"
+#include "storage/aggregator.h"
+#include "storage/table.h"
+
+namespace openmldb {
+namespace storage {
+
+Aggregator::Aggregator(const ::openmldb::api::TableMeta& base_meta, const ::openmldb::api::TableMeta& aggr_meta,
```
nautaa, line 28: multi put concurrent

src/sdk/sql_cluster_router.cc
```diff
@@ -2968,6 +2970,54 @@ hybridse::sdk::Status SQLClusterRouter::HandleLongWindows(
             if (!create_status.OK()) {
                 return {base::ReturnCode::kError, "create pre-aggr table failed"};
             }
+
+            // create aggregator
+            std::vector<std::shared_ptr<::openmldb::catalog::TabletAccessor>> tablets;
+            bool ret = cluster_sdk_->GetTablet(base_db, base_table, &tablets);
+            if (!ret || tablets.empty()) {
```
nautaa, line 3009: base table partition and pre aggr table partition in same tablet

zhanghaohit (changes requested):

dl239:
PR review meeting note:
- reuse `RowBuilder` and `RowView`
- guarantee  the thread safe when put in muti-thread
- the problem about distribution of base table and agg table
- (TODO) automatic  migrate agg table when migrating the base table.

src/sdk/sql_cluster_test.cc
nautaa: it started from 1. in this case, `mini_window_size = 2`, so the first of `binlog_offset` is `2`, and so on.

nautaa:
Major modifications include: 
1. base table and pre-aggr table use the same `table_partition`, so they have the same partition layout.
2. Add a mutex variable for each `AggrBuffer` to ensure thread safety when `AggrBuffer` is modified.
3. Timestamp ranges for pre-aggregated tables are contiguous.
4. reuse `RowBuilder` and `RowView` in `aggregator`.

src/storage/aggregator.cc
```diff
@@ -0,0 +1,327 @@
...
+            std::string origin_data = val.ToString();
+            aggr_row_view_.Reset(reinterpret_cast<int8_t*>(const_cast<char*>(origin_data.c_str())), origin_data.size());
+
+            AggrBuffer tmp_buffer;
+            bool ok = GetAggrBufferFromRowView(&aggr_row_view_, &tmp_buffer);
+            if (!ok) {
+                LOG(ERROR) << "GetAggrBufferFromRowView failed";
+                return false;
+            }
+            if (cur_ts > tmp_buffer.ts_end_ || cur_ts < tmp_buffer.ts_begin_) {
```
zhanghaohit: Are there any cases this condition is true?

src/storage/aggregator.cc
```diff
@@ -0,0 +1,327 @@
...
+    }
+    if (window_type_ == WindowType::kRowsRange && cur_ts > aggr_buffer.ts_end_) {
+        FlushAggrBuffer(key, aggr_buffer);
+    }
+
+    // handle the case that the current timestamp is smaller than the begin timestamp in aggregate buffer
+    if (cur_ts < aggr_buffer.ts_begin_) {
+        auto it = aggr_table_->NewTraverseIterator(0);
+        it->Seek(key, cur_ts);
+        if (it->Valid()) {
```
zhanghaohit: what if `Valid()` is false? Insert a new row?

src/storage/aggregator.cc
```diff
@@ -0,0 +1,327 @@
...
+        }
+    }
+    aggr_col_type_ = base_meta.column_desc(aggr_col_idx_).data_type();
+    ts_col_type_ = base_meta.column_desc(ts_col_idx_).data_type();
+    auto dimension = dimensions_.Add();
+    dimension->set_idx(0);
+}
+
+bool Aggregator::Update(const std::string& key, const std::string& row, const uint64_t& offset) {
+    base_row_view_.Reset(reinterpret_cast<int8_t*>(const_cast<char*>(row.c_str())), row.size());
```
zhanghaohit: For thread-safe, row_view cannot be re-used. What if `Update` is being called by multiple threads at the same time?

src/storage/aggregator.cc
```diff
@@ -0,0 +1,327 @@
...
+            base_row_view_.GetTimestamp(ts_col_idx_, &cur_ts);
+            break;
+        }
+        default: {
+            LOG(ERROR) << "Unsupported timestamp type: " << ts_col_type_;
+            return false;
+        }
+    }
+    if (aggr_buffer_map_.find(key) == aggr_buffer_map_.end()) {
+        aggr_buffer_map_.emplace(key, AggrBuffer{});
```
zhanghaohit: aggr_buffer_map_ is not thread-safe

src/storage/aggregator.cc
```diff
@@ -0,0 +1,327 @@
...
+    }
+    auto& aggr_buffer = aggr_buffer_map_.at(key);
+    std::lock_guard<std::mutex> lock(*aggr_buffer.mu_);
```
zhanghaohit: the lock is too broad. Only protect the aggr_buffer update/init/clear parts. Exclude the aggr_table read/update and codec parts.

src/storage/aggregator.cc
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    // init the next window range
+    auto& latest_val = aggr_buffer_map_.at(key);
+    int64_t latest_ts = latest_val.ts_end_ + 1;
+    latest_val.clear();
+    latest_val.ts_begin_ = latest_ts;
+    if (window_type_ == WindowType::kRowsRange) {
+        latest_val.ts_end_ = latest_ts + window_size_ - 1;
+    }
```
zhanghaohit: for efficiency and thread-safe, the AggrBuffer can be copied and passed in. The init/clear phases can be done outside and protected.

src/storage/aggregator.cc
```diff
@@ -0,0 +1,327 @@
...
+    return true;
+}
+
+SumAggregator::SumAggregator(const ::openmldb::api::TableMeta& base_meta, const ::openmldb::api::TableMeta& aggr_meta,
+                             std::shared_ptr<Table> aggr_table, const uint32_t& index_pos, const std::string& aggr_col,
+                             const AggrType& aggr_type, const std::string& ts_col, WindowType window_tpye,
+                             uint32_t window_size)
+    : Aggregator(base_meta, aggr_meta, aggr_table, index_pos, aggr_col, aggr_type, ts_col, window_tpye, window_size) {}
+
+bool SumAggregator::UpdateAggrVal(codec::RowView* row_view, AggrBuffer* aggr_buffer) {
```
zhanghaohit: row_view can be `const`?

src/tablet/tablet_impl.cc
```diff
@@ -5096,5 +5134,33 @@ void TabletImpl::BulkLoad(RpcController* controller, const ::openmldb::api::Bulk
...
+    if (!aggregator) {
+        response->set_code(::openmldb::base::ReturnCode::kError);
+        response->set_msg("create aggregator failed");
+        return;
+    }
+    uint64_t uid = (uint64_t) base_meta->tid() << 32 | base_meta->pid();
+    if (aggregators_.find(uid) == aggregators_.end()) {
+        aggregators_.emplace(uid, std::make_shared<Aggrs>());
+    }
+    aggregators_[uid]->push_back(aggregator);
```
zhanghaohit: lock protect the aggregators_ map?

src/tablet/tablet_impl_test.cc
```diff
@@ -5714,6 +5742,72 @@ TEST_F(TabletImplTest, AddIndex) {
...
+        request.set_aggr_table_pid(1);
+        request.set_aggr_col("col4");
+        request.set_aggr_func("sum");
+        request.set_index_pos(0);
+        request.set_order_by_col("ts_col");
+        request.set_bucket_size("1d");
+        ::openmldb::api::CreateAggregatorResponse response;
+        MockClosure closure;
+        tablet.CreateAggregator(NULL, &request, &response, &closure);
+        ASSERT_EQ(0, response.code());
```
zhanghaohit, line 5809: can double-check the `aggregators_` map?

zhanghaohit (changes requested):

src/storage/aggregator.cc
zhanghaohit: One way to re-use the row_view is to use the `GetValue` counterpart by passing the `row` pointer every time.

src/storage/aggregator.h
```diff
@@ -0,0 +1,154 @@
...
+ protected:
+    WindowType window_type_;
+
+    // for kRowsNum, window_size_ is the rows num in mini window
+    // for kRowsRange, window size is the time interval in mini window
+    int32_t window_size_;
+
+    codec::RowView base_row_view_;
+    codec::RowView aggr_row_view_;
+    codec::RowBuilder row_builder_;
```
zhanghaohit, line 137: thread-safe issue for row_builder_

src/storage/aggregator.cc
nautaa: > row_view can be `const`?

call a non-const member function on a const object is not allowed.

src/storage/aggregator.cc
nautaa: > what if `Valid()` is false? Insert a new row?

Under normal circumstances, it only occurs when cur_ts is smaller than all the previous ts, in which case a pre-aggr row is written for the out-of-order row.

src/storage/aggregator.cc
nautaa: > Are there any cases this condition is true?

It shouldn't happen, logging error messages have been added.

src/storage/aggregator.cc
zhanghaohit: can mark these `GetValue` and other as `const`. They should be const functions.

src/storage/aggregator.cc
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    {
+        std::lock_guard<std::mutex> lock(mu_);
+        if (aggr_buffer_map_.find(key) == aggr_buffer_map_.end()) {
+            aggr_buffer_map_.emplace(key, AggrBufferLocked{});
+        }
+    }
+
+    auto& aggr_buffer_lock = aggr_buffer_map_.at(key);
```
zhanghaohit: the `at()` has to be protected too.

src/storage/aggregator.cc
```diff
@@ -0,0 +1,364 @@
...
+        bool ok = UpdateAggrVal(base_row_view_, row_ptr, &aggr_buffer);
+        if (!ok) {
+            LOG(ERROR) << "Update aggr value failed";
+            return false;
+        }
+    }
+    return true;
+}
+
+bool Aggregator::GetAggrBufferFromRowView(codec::RowView* row_view, int8_t* row_ptr, AggrBuffer* buffer) {
```
zhanghaohit: `const` for `row_view` and `row_ptr`

src/storage/aggregator.cc
```diff
@@ -0,0 +1,364 @@
...
+        }
+        default: {
+            PDLOG(WARNING, "unsupported data type");
+            return false;
+        }
+    }
+    return true;
+}
+
+bool Aggregator::FlushAggrBuffer(const std::string& key, AggrBuffer buffer) {
```
zhanghaohit: `const AggrBuffer& buffer`

src/storage/aggregator.cc
```diff
@@ -0,0 +1,364 @@
...
+        }
+    }
+
+    int str_length = key.size() + aggr_val.size();
+    uint32_t row_size = row_builder_.CalTotalLength(str_length);
+    encoded_row.resize(row_size);
+    {
+        std::lock_guard<std::mutex> lock(rb_mu_);
+        row_builder_.SetBuffer(reinterpret_cast<int8_t*>(&(encoded_row[0])), row_size);
+        row_builder_.AppendString(key.c_str(), key.size());
```
zhanghaohit, line 213: can add an issue to add a thread-safe and re-usable `RowBuilder` as what `RowView` does.
e.g., `AppendString(int8_t* ptr, ...)`

src/storage/aggregator.cc
```diff
@@ -0,0 +1,364 @@
...
+    return false;
+}
+
+SumAggregator::SumAggregator(const ::openmldb::api::TableMeta& base_meta, const ::openmldb::api::TableMeta& aggr_meta,
+                             std::shared_ptr<Table> aggr_table, const uint32_t& index_pos, const std::string& aggr_col,
+                             const AggrType& aggr_type, const std::string& ts_col, WindowType window_tpye,
+                             uint32_t window_size)
+    : Aggregator(base_meta, aggr_meta, aggr_table, index_pos, aggr_col, aggr_type, ts_col, window_tpye, window_size) {}
+
+bool SumAggregator::UpdateAggrVal(const codec::RowView& row_view, int8_t* row_ptr, AggrBuffer* aggr_buffer) {
```
zhanghaohit: `const int8_t* row_ptr`

src/storage/aggregator.cc
```diff
@@ -0,0 +1,364 @@
...
+        }
+        case DataType::kFloat:
+        case DataType::kDouble: {
+            double val;
+            row_view.GetValue(row_ptr, aggr_col_idx_, aggr_col_type_, &val);
+            aggr_buffer->aggr_val_.vdouble += val;
+            break;
+        }
+        default: {
+            PDLOG(WARNING, "unsupported data type");
```
zhanghaohit: change all the `upsupported` to `PDLOG(ERROR)`

src/storage/aggregator.cc
```diff
@@ -0,0 +1,364 @@
...
+        window_type = WindowType::kRowsNum;
+        window_size = std::stoi(bucket_size);
+    } else {
+        window_type = WindowType::kRowsRange;
+        if (bucket_size.empty()) {
+            PDLOG(WARNING, "bucket size is empty");
+            return std::shared_ptr<Aggregator>();
+        }
+        char time_unit = bucket_size.back();
+        std::string time_size = bucket_size.substr(0, bucket_size.size() - 1);
```
zhanghaohit, line 347: do strip/trim in case there are spaces between the number and the unit

src/storage/aggregator.cc
```diff
@@ -0,0 +1,364 @@
...
+            }
+        }
+    }
+
+    // TODO(yuhang): support more aggr_type
+    if (aggr_func == "sum") {
+        return std::make_shared<SumAggregator>(base_meta, aggr_meta, aggr_table, index_pos, aggr_col, AggrType::kSum,
+                                               ts_col, window_type, window_size);
+    } else {
+        PDLOG(WARNING, "Unsupported aggregate function type");
```
zhanghaohit: change to `ERROR`

src/storage/aggregator_test.cc
```diff
@@ -0,0 +1,363 @@
...
+            row_builder.AppendInt64(i);
+            row_builder.AppendFloat(static_cast<float>(i));
+            row_builder.AppendDouble(static_cast<double>(i));
+            bool ok = aggr->Update(key, encoded_row, i);
+            ASSERT_TRUE(ok);
+        }
+        ASSERT_EQ(aggr_table->GetRecordCnt(), 50);
+        auto it = aggr_table->NewTraverseIterator(0);
+        it->SeekToFirst();
+        for (int i = 50 - 1; i >= 0; --i) {
```
zhanghaohit, line 277: also check the last un-committed aggrbuffer?

src/storage/aggregator_test.cc
```diff
@@ -0,0 +1,363 @@
...
+            row_builder.AppendTimestamp(i * 1000);
+            row_builder.AppendInt32(i);
+            row_builder.AppendInt16(i);
+            row_builder.AppendInt64(i);
+            row_builder.AppendFloat(static_cast<float>(i));
+            row_builder.AppendDouble(static_cast<double>(i));
+            bool ok = aggr->Update(key, encoded_row, i);
+            ASSERT_TRUE(ok);
+        }
+        ASSERT_EQ(aggr_table->GetRecordCnt(), 100);
```
zhanghaohit: also check the aggregated values?

src/storage/aggregator_test.cc
```diff
@@ -0,0 +1,363 @@
...
+        row_builder.AppendInt32(i);
+        row_builder.AppendInt16(i);
+        row_builder.AppendInt64(i);
+        row_builder.AppendFloat(static_cast<float>(i));
+        row_builder.AppendDouble(static_cast<double>(i));
+        bool ok = aggr->Update(key, encoded_row, i);
+        if (!ok) {
+            return false;
+        }
+    }
```
zhanghaohit, line 159: also check the aggregated values?

src/storage/aggregator_test.cc
```diff
@@ -0,0 +1,363 @@
...
+        counter += 2;
+        CheckAggregatorUpdate(counter, "col4", "sum");
+        counter += 2;
+        CheckAggregatorUpdate(counter, "col5", "sum");
+        counter += 2;
+        CheckAggregatorUpdate(counter, "col6", "sum");
+        counter += 2;
+        CheckAggregatorUpdate(counter, "col7", "sum");
+    }
+}
```
zhanghaohit, line 369: after each test, delete the base table and aggr table?

zhanghaohit (changes requested):

src/storage/aggregator_test.cc
nautaa: > after each test, delete the base table and aggr table?

Not needed in `aggregator test`, have dropped the table in the `tablet test`.

zhanghaohit (approved):
LGTM

src/storage/aggregator.cc
```diff
@@ -0,0 +1,383 @@
...
+        }
+        char time_unit = bucket_size.back();
+        std::string time_size = bucket_size.substr(0, bucket_size.size() - 1);
+        boost::trim(time_size);
+        if (!::openmldb::base::IsNumber(time_size)) {
+            PDLOG(ERROR, "Bucket size is not a number");
+            return std::shared_ptr<Aggregator>();
+        }
+        switch (time_unit) {
+            case 's':
```
zhanghaohit, line 354: we also have "ms" unit.

nautaa: > we also have "ms" unit.

however, when parsing SQL, there is no implementation for `ms`. see `ASTIntervalLIteralToNum`

zhanghaohit: oh. ic

src/tablet/tablet_impl.cc
```diff
@@ -3761,6 +3763,42 @@ std::shared_ptr<Table> TabletImpl::GetTableUnLock(uint32_t tid, uint32_t pid) {
...
+    std::lock_guard<SpinMutex> spin_lock(spin_mutex_);
+    return GetAggregatorsUnLock(tid, pid);
+}
+
+std::shared_ptr<Aggrs> TabletImpl::GetAggregatorsUnLock(uint32_t tid, uint32_t pid) {
+    uint64_t uid = (uint64_t) tid << 32 | pid;
+    if (aggregators_.find(uid) == aggregators_.end()) {
+        return std::shared_ptr<Aggrs>();
+    }
+    return aggregators_.at(uid);
```
dl239: it's no need to find again. get `Aggrs` from iterator directly

src/tablet/tablet_impl.cc
```diff
@@ -644,6 +639,14 @@ void TabletImpl::Put(RpcController* controller, const ::openmldb::api::PutReques
         replicator->AppendEntry(entry);
     } while (false);
 
+    ok = UpdateAggrs(request->tid(), request->pid(), request->value(),
+                     request->dimensions(), replicator->GetLogOffset());
```
dl239: the offset obtained  by `replicator->GetLogOffset()` may be not referenced this recored. if you want to use the offset of current record, `log_index` in entry is recommend

src/storage/aggregator.cc
```diff
@@ -0,0 +1,383 @@
...
+    if (CheckBufferFilled(cur_ts, aggr_buffer.ts_end_, aggr_buffer.aggr_cnt_)) {
+        AggrBuffer flush_buffer = aggr_buffer;
+        int64_t latest_ts = aggr_buffer.ts_end_ + 1;
+        aggr_buffer.clear();
+        aggr_buffer.ts_begin_ = latest_ts;
+        if (window_type_ == WindowType::kRowsRange) {
+            aggr_buffer.ts_end_ = latest_ts + window_size_ - 1;
+        }
+        lock.unlock();
+        FlushAggrBuffer(key, flush_buffer);
```
dl239, line 108: `aggr_buffer.ts_begin_` may be modified here in another thread, whether you want to keep the same value in next if block

nautaa: > `aggr_buffer.ts_begin_` may be modified here in another thread, whether you want to keep the same value in next if block

@dl239 This is not required? If `ts_begin` changes in another thread, treat `aggr update` as out of order to update.

src/storage/aggregator.cc
```diff
@@ -0,0 +1,383 @@
...
+            PDLOG(ERROR, "Unsupported data type");
+            return false;
+        }
+    }
+
+    int str_length = key.size() + aggr_val.size();
+    uint32_t row_size = row_builder_.CalTotalLength(str_length);
+    encoded_row.resize(row_size);
+    {
+        std::lock_guard<std::mutex> lock(rb_mu_);
```
dl239, line 211: Remove the lock. reuse `RowBuilder` if #1525 has been done

nautaa: yep, it will be done in the next pr.

src/storage/aggregator.cc
```diff
@@ -0,0 +1,383 @@
...
+        }
+    }
+
+    AggrBufferLocked* aggr_buffer_lock;
+    {
+        std::lock_guard<std::mutex> lock(mu_);
+        if (aggr_buffer_map_.find(key) == aggr_buffer_map_.end()) {
+            aggr_buffer_map_.emplace(key, AggrBufferLocked{});
+        }
+        aggr_buffer_lock = &aggr_buffer_map_.at(key);
```
dl239: `aggr_buffer_map_` may be very large, do not find again. use the iterator emplace function returned.

src/storage/aggregator.cc
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    std::unique_lock<std::mutex> lock(*aggr_buffer_lock->mu_);
```
dl239, line 88: One key own one mutex, this will cost much more memory. It's better to design efficient strategy to achieve thread safe. or post an issue  to optimize later @zhanghaohit @nautaa

src/storage/aggregator.h
```diff
@@ -0,0 +1,163 @@
...
+    int64_t vlong;
+    float vfloat;
+    double vdouble;
+};
+
+struct AggrBuffer {
+    AggrVal aggr_val_;
+    int64_t ts_begin_;
+    int64_t ts_end_;
+    int32_t aggr_cnt_;
```
dl239, line 61: move `aggr_cnt_` to the first position.

src/storage/aggregator.cc
```diff
@@ -0,0 +1,383 @@
...
+
+bool Aggregator::UpdateFlushedBuffer(const std::string& key, const int8_t* base_row_ptr, int64_t cur_ts,
+                                     uint64_t offset) {
+    auto it = aggr_table_->NewTraverseIterator(0);
+    // If there is no repetition of ts, `seek` will locate to the position that less than ts.
+    it->Seek(key, cur_ts + 1);
+    AggrBuffer tmp_buffer;
+    if (it->Valid()) {
+        auto val = it->GetValue();
+        std::string origin_data = val.ToString();
```
dl239: Do not assign to another string. Get data pointer from Slice

src/storage/aggregator.cc
```diff
@@ -0,0 +1,383 @@
...
+    }
+    row_view.GetValue(row_ptr, 1, DataType::kTimestamp, &buffer->ts_begin_);
+    row_view.GetValue(row_ptr, 2, DataType::kTimestamp, &buffer->ts_end_);
+    row_view.GetValue(row_ptr, 3, DataType::kInt, &buffer->aggr_cnt_);
+    char* ch = NULL;
+    uint32_t ch_length = 0;
+    row_view.GetValue(row_ptr, 4, &ch, &ch_length);
+    switch (aggr_col_type_) {
+        case DataType::kSmallInt:
+        case DataType::kInt:
```
dl239, line 156: why not use `vint` and `vsmallint`

src/storage/aggregator.h
```diff
@@ -0,0 +1,163 @@
...
+        memset(&aggr_val_, 0, sizeof(aggr_val_));
+        ts_begin_ = -1;
+        ts_end_ = 0;
+        aggr_cnt_ = 0;
+        binlog_offset_ = 0;
+    }
+};
+struct AggrBufferLocked {
+    std::unique_ptr<std::mutex> mu_;
+    AggrBuffer buffer_;
```
dl239, line 74: it's better to use consecutive buffer to store these field in order to reduce memory usage

src/storage/aggregator.cc
nautaa: for possible overflow of `aggr_val` in `SumAggregator`

src/storage/aggregator.cc
nautaa: > One key own one mutex, this will cost much more memory. It's better to design efficient strategy to achieve thread safe. or post an issue to optimize later @zhanghaohit @nautaa

This is an implementation problem of concurrent hashmap, maybe we can introduce the implementation of concurrent hashmap from third-party libraries in the furture.

src/storage/aggregator.h
nautaa: > it's better to use consecutive buffer to store these field in order to reduce memory usage

the union of `aggr_val` extra memory usage is small compared to the table space. and also need a pointer to local buffer space will increase memory usage.

dl239: with [flexible array member](https://en.wikipedia.org/wiki/Flexible_array_member)

dl239 (approved):

src/storage/aggregator.cc
zhanghaohit: https://github.com/4paradigm/OpenMLDB/issues/1559

We'll optimize in the next step.

src/storage/aggregator.h
zhanghaohit: #1578

PR #1498, opened 2022-03-22
feat: update Python SDK smoke test 

HuilinWu2:
1. We are using `ExecuteSQL` instead of `createDB`, `DropDB` methods, and we may further don't consider compiling them into the dylib/so for Python.


aceforeverd (approved):

tobegit3hub (approved):
LGTM

PR #1539, opened 2022-03-30
fix: autoLoad format

vagetablechicken:
`autoLoad` should use the arg `format`, not the reader's default format.
fix ut.


dl239 (approved):

tobegit3hub (approved):
LGTM

PR #1540, opened 2022-03-30
fix: execute udf function on string field will core dump if length greater than 2048

### LINKED ISSUE #1537: Execute udf function on string field will core dump if length greater than 2048
WX20220330-135841

dl239:
expand `MAX_ALLOC_SIZE` to 2M


tobegit3hub (approved):
LGTM

zhanghaohit (approved):

aceforeverd (approved):

PR #1551, opened 2022-03-31
fix: removes an unused variable and its statement

### LINKED ISSUE #1536: delete unused variable and do right close
https://github.com/4paradigm/OpenMLDB/blob/230f99b9b5744f2663b0253a87a2496c9f5ce96b/java/openmldb-taskmanager/src/main/scala/com/_4paradigm/openmldb/taskmanager/JobInfoManager.scala#L141-L170

`pstmt ` is useless, delete it. And delete `statement ` in `finally` block.

johndinh391:
It removes an unused variable and its related statement.

It should resolve #1536.

There should be no new behavior.


java/openmldb-taskmanager/src/main/scala/com/_4paradigm/openmldb/taskmanager/JobInfoManager.scala
```diff
@@ -161,9 +160,8 @@ object JobInfoManager {
     } catch {
       case e: SQLException =>
         e.printStackTrace()
-    } finally if (pstmt != null) try {
-      pstmt.close()
-    } catch {
+    } finally { 
+      catch {
```
tobegit3hub: Remove this as well.

java/openmldb-taskmanager/src/main/scala/com/_4paradigm/openmldb/taskmanager/JobInfoManager.scala
```diff
@@ -149,7 +149,6 @@ object JobInfoManager {
     statement.setString(8, job.getApplicationId)
     statement.setString(9, job.getError)
 
-    var pstmt: PreparedStatement = null
     try {
```
aceforeverd, line 152: looks wired, maybe u can use `try with resource`  alternative in scala: https://www.baeldung.com/scala/try-with-resources

aceforeverd (changes requested):

java/openmldb-taskmanager/src/main/scala/com/_4paradigm/openmldb/taskmanager/JobInfoManager.scala
```diff
@@ -161,9 +161,13 @@ object JobInfoManager {
     } catch {
       case e: SQLException =>
         e.printStackTrace()
-    } finally if (pstmt != null) try {
-      pstmt.close()
-    } catch {
+    } Using.Manager { use =>
```
aceforeverd: I'm  afraid it is not the correct way to with `Using.Manager`, here we want to automatically close the `pstmt` object in the try block (that in line 155), so u should replace the first try block with `Using.Manager`. In my mind, the second try ... catch block is unnecessary

aceforeverd:
Hi @johndinh391, the PR won't pass the tests. So I guess you should first try to run the related test locally before submit changes.
You can mainly focus on the java/scala tests,  have a look at the workflow job: https://github.com/4paradigm/OpenMLDB/blob/f5fc988a38d0d9b8952155cdfe9d782133502e46/.github/workflows/sdk.yml#L79-L101

In summary, 
1. compile: run in top directory as `make build SQL_JAVASDK_ENABLE=ON`
2. Setup test requirements, as:
     ```bash
          sh steps/ut_zookeeper.sh start
          # replace $SPARK_HOME with the real directory path, give whatever u want
          sh steps/download_openmldb_spark.sh $SPARK_HOME
          cd onebox && sh start_onebox.sh && sh start_onebox.sh standalone && cd - || exit
     ```
3. test: `cd java/ && ./mvnw test`
    it will run all tests, but if u are in IDE, u probably find way to run single tests as well

tobegit3hub (approved):
LGTM

vagetablechicken (approved):

aceforeverd (approved):

PR #1556, opened 2022-03-31
ci: improve & update workflow files

aceforeverd:

tow fixes introduced:
1. docker tagging support push to man and workflow_dispatch event
2. fix `paths-ignore` option in compile workflows, they target on file path instead directory

### DIFF
--- a/.github/workflows/hybridsql-docker.yml
+++ b/.github/workflows/hybridsql-docker.yml
@@ -50,10 +50,18 @@ jobs:
           images: |
             ghcr.io/${{ github.repository_owner }}/${{ env.IMAGE_NAME }}
             ${{ env.DOCKERHUB_REPO }}/${{ env.IMAGE_NAME }}
+          # tagging ruels, means in order(event -> result tag):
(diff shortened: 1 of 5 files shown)


dl239 (approved):

.github/workflows/hybridsql-docker.yml
```diff
@@ -50,10 +50,18 @@ jobs:
           images: |
             ghcr.io/${{ github.repository_owner }}/${{ env.IMAGE_NAME }}
             ${{ env.DOCKERHUB_REPO }}/${{ env.IMAGE_NAME }}
+          # tagging ruels, means in order(event -> result tag):
```
vagetablechicken, line 53: rules.
and 1&2 is the same event, but result is not the same?

tobegit3hub (approved):
LGTM

.github/workflows/hybridsql-docker.yml
vagetablechicken: 0.x will be the latest 0.x.y

PR #1573, opened 2022-04-02
fix: support pre-aggregate in standalone

### LINKED ISSUE #1560: Create PreAggrTables in standalone mode
**Bug Description**
Only in cluster mode, it will create pre-aggr tables. We should do it also in standalone mode.

nautaa:
fix #1560

### DIFF
--- a/src/catalog/base.cc
+++ b/src/catalog/base.cc
@@ -44,7 +44,6 @@ ProcedureInfoImpl::ProcedureInfoImpl(const ::openmldb::api::ProcedureInfo& proce
 
     for (const auto& op : procedure.options()) {
         options_[op.name()] = op.value().value();
-        DLOG(INFO) << "[procedure option] " << op.name() << " : " << options_[op.name()];
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/catalog/base.cc
```diff
@@ -44,7 +44,6 @@ ProcedureInfoImpl::ProcedureInfoImpl(const ::openmldb::api::ProcedureInfo& proce
 
     for (const auto& op : procedure.options()) {
         options_[op.name()] = op.value().value();
-        DLOG(INFO) << "[procedure option] " << op.name() << " : " << options_[op.name()];
```
nautaa, line 47: in standalone mode, this will generate periodic logs on the command line.

zhanghaohit (approved):
LGTM

src/catalog/base.cc
zhanghaohit: ok to remove.

Actually it will only print when compiling with `Debug` mode.

zhanghaohit:

dl239 (approved):

PR #1616, opened 2022-04-12
feat: full iterator

### LINKED ISSUE #1419: DistributedIterator
`DistributedIterator` can be used to iterate over the table that is partitioned into multiple tablet servers.

This can be used to:
- replace the SQLRouter in the tablet server.
- support online batch query

An iterator interface to scan all the data in a distributed table.

dl239:


src/flags.cc
```diff
@@ -95,6 +95,7 @@ DEFINE_int32(request_timeout_ms, 20000, "request timeout");
 DEFINE_int32(request_sleep_time, 1000, "the sleep time when request error");
 
 DEFINE_uint32(max_traverse_cnt, 50000, "max traverse iter loop cnt");
+DEFINE_uint32(traverse_cnt, 1000, "traverse iter loop cnt");
```
vagetablechicken: `traverse_cnt` is not a good name? How about `traverse_cnt_limit`?

vagetablechicken: And we'd better to tell the difference between  `max_traverse_cnt ` and `traverse_cnt `

src/catalog/distribute_iterator.cc
```diff
@@ -15,53 +15,153 @@
...
+    }
+    if (kv_it_) {
+        kv_it_->Next();
+        if (kv_it_->Valid()) {
+            key_ = kv_it_->GetKey();
+            last_pk_ = kv_it_->GetPK();
+            return true;
+        }
     }
+    auto iter = tablet_clients_.end();;
```
zhanghaohit: `tablet_clients.begin()`?

src/catalog/distribute_iterator.h
```diff
@@ -43,10 +46,21 @@ class FullTableIterator : public ::hybridse::codec::ConstIterator<uint64_t, ::hy
...
+ private:
+    uint32_t tid_;
     std::shared_ptr<Tables> tables_;
+    std::map<uint32_t, std::shared_ptr<openmldb::client::TabletClient>> tablet_clients_;
+    bool in_local_;
     uint32_t cur_pid_;
     std::unique_ptr<::openmldb::storage::TableIterator> it_;
+    std::unique_ptr<::openmldb::base::KvIterator> kv_it_;
     uint64_t key_;
+    std::string last_pk_;
```
zhanghaohit, line 65: also add the `GetPK()` API?

src/catalog/tablet_catalog.cc
```diff
@@ -150,10 +147,17 @@ const ::hybridse::codec::Row TabletTableHandler::Get(int32_t pos) {
...
+        if (tables->count(pid) == 0) {
+            auto accessor = table_client_manager_->GetTablet(pid);
+            if (accessor) {
+                tablet_clients.emplace(pid, accessor->GetClient());
+            }
+        }
     }
-    return nullptr;
+    DLOG(INFO) << "table size " << tables->size() << " tablet_clients size " << tablet_clients.size();
+    return new catalog::FullTableIterator(GetTid(), tables, tablet_clients);
```
zhanghaohit, line 172: here it is a bit dangerous as previously it is assumed to be local tables only. `select *` may return duplicate results.

have to check

zhanghaohit (changes requested):

dl239:
it need to modify `DistributeWindowIterator` to reslove `select * from t1 where xxx=xxx;`

src/catalog/distribute_iterator.h
dl239: `last_pk_` is assigned only in `NextFromRemote`. if provide `GetPK` API, we should assign `last_pk_` in `NextFromLocal` which need to copy string content. of course, if we indeed need later, we can add it  in  the future

src/catalog/tablet_catalog.cc
dl239: there is a test case to cover this scene
https://github.com/4paradigm/OpenMLDB/blob/e9acd54366616e9e7211dd17db7dd01bbf07bbd2/src/catalog/distribute_iterator_test.cc#L98-L123

zhanghaohit: so `FullTableIterator` is never used before in the sql engine?

zhanghaohit: I mean, in the cluster mode, to handle the `select  * from t1`:
- send the query to all the tablets
- each tablet uses the `FullTableIterator` to traverse the `local` table (before this PR)
- collect the response from all tablets and merge the results

However, if changing the `FullTableIterator` to iterating over a distributed table (i.e., collecting both local and remote partitions), the result from each `tablet` will be a full table. After merging, results would be duplicated.

dl239: `FullTableIterator` will make sure that there  no duplicated data, which will hold the  partition leader located in local memory and remote tablet client. 
`SQLCusterRouter` will send request to all tablets and merge response data in local previously. In this pr,  `SQLCusterRouter` will send only one tablet that main table located in and iterator all data by `FullTableIterator`

zhanghaohit: oh. overlook the changes in `SQLClusterRouter`

zhanghaohit:
can add sql test in `sql_cluster_test` using `DistributedIterator` and `FullTableIterator`?

I tried with following case in `sql_cluster_test`, but it always failed (get less than 1000):
```c++
    std::string ddl = "create table " + name +
                      "("
                      "col1 int not null,"
                      "col2 bigint default 112 not null,"
                      "col4 string default 'test4' not null,"
                      "col5 date default '2000-01-01' not null,"
                      "col6 timestamp default 10000 not null,"
                      "index(key=col1, ts=col2)) options(partitionnum=8);";
    ok = router->ExecuteDDL(db, ddl, &status);
    ASSERT_TRUE(ok);
    ASSERT_TRUE(router->RefreshCatalog());
    std::string sql;
    int expect = 1000;
    for (int i = 0; i < expect; i++) {
        sql = "insert into " + name + " values(" + std::to_string(i) + ", 1, '1', '2021-01-01', 1);";
        ASSERT_TRUE(router->ExecuteInsert(db, sql, &status));
    }
    auto res = router->ExecuteSQL(db, "select * from " + name, &status);
    ASSERT_TRUE(res);
    int count = 0;
    while (res->Next()) {
        count++;
    }
    LOG(WARNING) << "count = " << count;
    EXPECT_EQ(count, expect);
```

vagetablechicken (approved):

zhanghaohit (approved):
LGTM

src/tablet/tablet_impl.cc
```diff
@@ -4333,96 +4340,50 @@ void TabletImpl::RefreshTableInfo() {
...
+        row_view.GetValue(row.buf(), 7, &str, &len);
+        table_info.order_by_col.assign(str, len);
+        row_view.GetValue(row.buf(), 8, &str, &len);
+        table_info.bucket_size.assign(str, len);
+
+        table_infos.emplace_back(std::move(table_info));
+        it->Next();
+    }
+    catalog_->RefreshAggrTables(table_infos);
+    DLOG(INFO) << "Refresh agg catalog in standalone mode (size = " << table_infos.size() << ")";
```
zhanghaohit: `LOG(INFO) << "Refresh agg catalog in (size = " << table_infos.size() << ")";`

no need the `standalone mode` as they are merged.

PR #1621, opened 2022-04-13
feat: select into null exception

### LINKED ISSUE #1459: batch: return error if `select into` gets empty
https://github.com/4paradigm/OpenMLDB/blob/420398b8349880bce262db06c3b143d023b56862/java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/nodes/SelectIntoPlan.scala#L39-L41

If `select` closure in `select into` result is empty, we just skip saving, and the job is succeed. But users can't find the output file. It's not intuitive.
We should throw an exception.

KaidoWang:
If select closure in select into result is empty, we just skip saving, and the job is succeed. But users can't find the output file. It's not intuitive.
We should throw an exception.

close #1459 

throw new Exception("select empty, skip save")


vagetablechicken (approved):

lumianph (approved):

tobegit3hub (approved):
LGTM

PR #1622, opened 2022-04-13
feat: aggregator recovery

### LINKED ISSUE #1338: Support fault tolerance for pre-aggregation tables
sub-task of long-time window feature

Ref to [RFC](https://docs.google.com/document/d/18dX75IxqyvH39Lu9PEcRTMG2-FMS_bPw0pYsLmmzSXU/edit)

**Additional context**
long-time window feature

nautaa:
feature
close #1338

support aggregator recovery from tablet restart.
1. recreate aggregator from persistent metadata.
2. reconstruct aggr buffer map from aggr table.
3.  update aggr from base table binlog.
    * `recovery binlog offset` starts from the minimize offset of all keys latest offset.
4. add `DecodeAggrval`.
5. refactor `AggrBuffer`.

**recover logic**

* if aggr table recovers first
1. create the aggregator
2. init will fail
3. when base table recovers, re-set base replicator and re-init
* if base table recovers first
1. when the aggr recovers, create the aggregator and the init will succeed


src/storage/aggregator.cc
```diff
@@ -143,6 +159,106 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
...
+        return false;
+    }
+    auto log_parts = base_replicator_->GetLogPart();
+    if (aggr_table_->GetRecordCnt() == 0 && log_parts->IsEmpty()) {
+        status_.store(AggrStat::kInited, std::memory_order_relaxed);
+        return true;
+    }
+    auto table_it = aggr_table_->NewTraverseIterator(0);
+    // TODO: need abstract TraverseIterator?
+    auto it = dynamic_cast<MemTableTraverseIterator*>(table_it);
```
nautaa: Is it better to add abstract TraverseIterator for use `NextPK()` to iterate between keys?

src/replica/log_replicator.h
```diff
@@ -108,6 +108,8 @@ class LogReplicator {
 
     bool DelAllReplicateNode();
 
+    std::string GetLogPath() {return log_path_;}
```
zhanghaohit: can return `const std::string&`

nautaa:
>    * `recovery binlog offset` starts from the minimize offset of all keys latest offset.

e.g.
```
key = 1, binlog offset = 2,4,6. the latest offset is 6. and then offset 7 is cached in volatile aggr.
key = 2, binlog offset = 9, 10. the latest offset is 10.
```
we need to scan the binlog after 6.

src/tablet/tablet_impl.cc
```diff
@@ -5500,22 +5531,72 @@ void TabletImpl::DropFunction(RpcController* controller, const openmldb::api::Dr
...
+        response->set_msg("pre-aggr table db path is not exist");
+        return;
+    }
+    std::string aggr_info;
+    google::protobuf::TextFormat::PrintToString(*request, &aggr_info);
+    std::string aggr_info_path = aggr_table_path + "/aggr_info.txt";
+
+    FILE* fd_write = fopen(aggr_info_path.c_str(), "w");
+    if (fd_write == NULL) {
+        PDLOG(WARNING, "fail to open file %s. err[%d: %s]", aggr_info_path.c_str(), errno, strerror(errno));
```
zhanghaohit, line 5778: if `NULL`, have to return and set error code.

src/tablet/tablet_impl.cc
```diff
@@ -5500,22 +5531,72 @@ void TabletImpl::DropFunction(RpcController* controller, const openmldb::api::Dr
...
-        response->set_code(::openmldb::base::ReturnCode::kError);
-        response->set_msg("create aggregator failed");
-        return;
+        msg.assign("create aggregator failed");
+        return false;
+    }
+
+    auto base_replicator = GetReplicator(base_meta->tid(), base_meta->pid());
+    if (!base_replicator) {
+        PDLOG(WARNING, "base table replicator is not exist. tid %u, pid %u", base_meta->tid(), base_meta->pid());
```
zhanghaohit: return

zhanghaohit (changes requested):

src/storage/aggregator.cc
```diff
@@ -143,6 +161,108 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
...
 
+bool Aggregator::Init() {
+    if (GetStat() != AggrStat::kUnInit) {
+        PDLOG(INFO, "aggregator is normal or recovering");
+        return true;
+    }
+
+    status_.store(AggrStat::kRecovering, std::memory_order_relaxed);
+    if (!base_replicator_) {
+        status_.store(AggrStat::kUnInit, std::memory_order_relaxed);
```
zhanghaohit: why not put `status_.store(AggrStat::kRecovering, std::memory_order_relaxed);` after the `if` statement. Then no need to restore the `kUnInit` stat.

src/storage/aggregator.cc
```diff
@@ -143,6 +161,108 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
...
+    }
+    auto log_parts = base_replicator_->GetLogPart();
+
+    if (aggr_table_->GetRecordCnt() == 0 && log_parts->IsEmpty()) {
+        status_.store(AggrStat::kInited, std::memory_order_relaxed);
+        return true;
+    }
+    auto it = aggr_table_->NewTraverseIterator(0);
+    it->SeekToFirst();
+    uint64_t recovery_offset = INT64_MAX;
```
zhanghaohit: why not set `UINT64_MAX`?

src/tablet/tablet_impl.cc
nautaa: if base table is not been recovered, It's ok to continue. the replicator will be set when base table recovery.

src/storage/aggregator.cc
```diff
@@ -143,6 +164,113 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
     return true;
 }
 
+bool Aggregator::Init() {
+    if (GetStat() != AggrStat::kUnInit) {
+        PDLOG(INFO, "aggregator is normal or recovering");
+        return true;
+    }
```
zhanghaohit, line 210: what if two threads are doing `Init` at the same time? They both `GetStat` of `kUnInit`, and continue to do the recovery process?

Maybe can use `cmp_and_set` atomic operations or lock

src/storage/aggregator.h
```diff
@@ -101,8 +113,14 @@ class Aggregator {
 
     uint32_t GetWindowSize() const { return window_size_; }
 
+    uint32_t GetStat() const { return status_.load(std::memory_order_relaxed); }
+
     bool GetAggrBuffer(const std::string& key, AggrBuffer* buffer);
 
+    void SetBaseReplicator(std::shared_ptr<LogReplicator> replicator) { base_replicator_ = replicator; }
+
+    std::shared_ptr<LogReplicator> GetBaseReplicator() { return base_replicator_; }
```
zhanghaohit: can use 'store` and `load`, as base table and aggr table may be recovering at the same time, causing the `base_replication` access contention?

src/tablet/tablet_impl.cc
```diff
@@ -2944,6 +2944,37 @@ int TabletImpl::LoadTableInternal(uint32_t tid, uint32_t pid, std::shared_ptr<::
...
+                    if (!google::protobuf::TextFormat::Parse(&fileInput, &request)) {
+                        PDLOG(WARNING, "parse create aggregator meta failed");
+                    }
+                    std::string msg;
+                    bool ok = CreateAggregatorInternal(&request, msg);
+                    if (!ok) {
+                        PDLOG(WARNING, "create aggregator failed. msg %s", msg.c_str());
+                    }
+                }
+                auto aggrs = GetAggregators(tid, pid);
```
zhanghaohit, line 3072: so `aggr_info.txt` only exists for every aggr table? not base table, right?

so the `tid` and `pid` here are for aggr table. So the `GetAggregators(tid, pid)` will always return null?

src/tablet/tablet_impl.cc
```diff
@@ -2944,6 +2944,37 @@ int TabletImpl::LoadTableInternal(uint32_t tid, uint32_t pid, std::shared_ptr<::
...
+                if (fd < 0) {
+                    PDLOG(ERROR, "[%s] is not exist", aggr_path.c_str());
+                } else {
+                    google::protobuf::io::FileInputStream fileInput(fd);
+                    fileInput.SetCloseOnDelete(true);
+                    if (!google::protobuf::TextFormat::Parse(&fileInput, &request)) {
+                        PDLOG(WARNING, "parse create aggregator meta failed");
+                    }
+                    std::string msg;
+                    bool ok = CreateAggregatorInternal(&request, msg);
```
zhanghaohit: so current logic is:
- if aggr table recovers first
    1. create the aggregator
    2. init will fail
    3. when base table recovers, re-set base replicator and re-init
- if base table recovers first
   1. when the aggr recovers, create the aggregator and the `init` will succeed

Is it right?

src/storage/aggregator.cc
```diff
@@ -143,6 +164,113 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
...
+        for (int i = 0; i < entry.dimensions_size(); i++) {
+            const auto& dimension = entry.dimensions(i);
+            if (dimension.idx() == index_pos_) {
+                Update(dimension.key(), entry.value(), entry.log_index(), true);
+                break;
+            }
+        }
+        cur_offset = entry.log_index();
+    }
+    if (cur_offset < aggr_latest_offset) {
```
zhanghaohit, line 309: check with @dl239, 
if leader changes, two tables with same partition and replication information,
after re-select the leader, will the two tables have the same tablet as the leader.

src/storage/aggregator.cc
```diff
@@ -218,6 +347,13 @@ bool Aggregator::FlushAggrBuffer(const std::string& key, const AggrBuffer& buffe
         PDLOG(ERROR, "Aggregator put failed");
         return false;
     }
+    ::openmldb::api::LogEntry entry;
+    entry.set_pk(key);
+    entry.set_ts(time);
+    entry.set_value(encoded_row);
+    entry.set_term(aggr_replicator_->GetLeaderTerm());
+    entry.mutable_dimensions()->CopyFrom(dimensions_);
+    aggr_replicator_->AppendEntry(entry);
```
zhanghaohit, line 380: for the `aggregator`, I think we'd better call the `TabletImpl::Put` API instead of the `Table::Put` API. Otherwise we have to do the duplicate codes for the replication logics and others.

Or we can extract a `PutInternal` in the tablet_impl and both call the `PutInternal`?

zhanghaohit (changes requested):

src/tablet/tablet_impl.cc
nautaa: > so `aggr_info.txt` only exists for every aggr table? not base table, right?
> 
> so the `tid` and `pid` here are for aggr table. So the `GetAggregators(tid, pid)` will always return null?


yep, `GetAggregators` here is used to init aggregator related to base table.

src/tablet/tablet_impl.cc
nautaa: yep. a little complicated.

src/storage/aggregator.cc
nautaa: yep, I think it has some replication logic. However, in that case, `aggregator` will depend on `tablet`? or how about using  `tablet client` in `aggregator`.

nautaa: or pass `flushed row` as parameter to `tablet`?

src/tablet/tablet_impl.cc
zhanghaohit: > 

I mean the input of the `GetAggregators` should be the `tid` and `pid` of base table. However, here they are for the aggr table?

nautaa: > > 
> 
> I mean the input of the `GetAggregators` should be the `tid` and `pid` of base table. However, here they are for the aggr table?

my mistake, will fix.

src/storage/aggregator.cc
zhanghaohit: > or pass `flushed row` as parameter to `tablet`?

you mean when calling `Put` API?

zhanghaohit: Or maybe can pass a `Put` function pointer to `Aggregator`?

nautaa: > Or maybe can pass a `Put` function pointer to `Aggregator`?

still need object to call class function pointer.

zhanghaohit:

src/storage/aggregator.cc
zhanghaohit: > > Or maybe can pass a `Put` function pointer to `Aggregator`?
> 
> still need object to call class function pointer.

Can do like this: 
`auto put_func = std::bind(&TabletImpl::Put, this, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3, std::placeholders::_4);`

src/storage/aggregator.cc
```diff
@@ -143,47 +157,136 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
...
+        PDLOG(ERROR, "base table is slower than aggregator");
+        status_.store(AggrStat::kUnInit, std::memory_order_relaxed);
+        return false;
+    }
+    PDLOG(INFO, "aggregator recovery finished");
+    status_.store(AggrStat::kInited, std::memory_order_relaxed);
+    return true;
+}
+
+bool Aggregator::GetAggrBuffer(const std::string& key, AggrBuffer*& buffer) {
```
zhanghaohit: use `AggrBuffer**` instead of `AggrBuffer*&`

src/storage/aggregator.h
```diff
@@ -101,7 +138,19 @@ class Aggregator {
...
     uint32_t GetWindowSize() const { return window_size_; }
 
-    bool GetAggrBuffer(const std::string& key, AggrBuffer* buffer);
+    uint32_t GetStat() const { return status_.load(std::memory_order_relaxed); }
+
+    bool GetAggrBuffer(const std::string& key, AggrBuffer** buffer);
+
+    void SetBaseReplicator(std::shared_ptr<LogReplicator> replicator) {
+        std::lock_guard<std::mutex> lock(mu_);
+        base_replicator_ = replicator;
```
zhanghaohit: here no need `lock`.

std::atomic_load_explicit(&tables_, std::memory_order_relaxed);

src/storage/aggregator.cc
zhanghaohit: #1657

src/storage/aggregator.h
nautaa: > here no need `lock`.
> 
> std::atomic_load_explicit(&tables_, std::memory_order_relaxed);

This has been modified in the previous commit, and it is a bit strange that this is still visible.

zhanghaohit: may need re-do

src/tablet/tablet_impl.cc
```diff
@@ -2944,6 +2944,38 @@ int TabletImpl::LoadTableInternal(uint32_t tid, uint32_t pid, std::shared_ptr<::
         ::openmldb::storage::Binlog binlog(replicator->GetLogPart(), binlog_path);
         if (snapshot->Recover(table, snapshot_offset) &&
             binlog.RecoverFromBinlog(table, snapshot_offset, latest_offset)) {
+            // recover aggregator if exists
+            std::string aggr_path = GetDBPath(db_root_path, tid, pid) + "/aggr_info.txt";
+            if (::openmldb::base::IsExists(aggr_path)) {
+                int fd = open(aggr_path.c_str(), O_RDONLY);
+                ::openmldb::api::CreateAggregatorRequest request;
+                if (fd < 0) {
+                    PDLOG(ERROR, "[%s] is not exist", aggr_path.c_str());
```
dl239: msg is open file failed

src/tablet/tablet_impl.cc
```diff
@@ -2944,6 +2944,38 @@ int TabletImpl::LoadTableInternal(uint32_t tid, uint32_t pid, std::shared_ptr<::
...
+                } else {
+                    google::protobuf::io::FileInputStream fileInput(fd);
+                    fileInput.SetCloseOnDelete(true);
+                    if (!google::protobuf::TextFormat::Parse(&fileInput, &request)) {
+                        PDLOG(WARNING, "parse create aggregator meta failed");
```
dl239, line 3061: need not to create if parse failed

src/tablet/tablet_impl.cc
```diff
@@ -2944,6 +2944,38 @@ int TabletImpl::LoadTableInternal(uint32_t tid, uint32_t pid, std::shared_ptr<::
...
+                    if (!ok) {
+                        PDLOG(WARNING, "create aggregator failed. msg %s", msg.c_str());
+                    }
+                }
+            }
+            // init aggregator related to base table if need.
+            auto aggrs = GetAggregators(tid, pid);
+            if (aggrs != nullptr) {
+                for (auto& aggr : *aggrs) {
+                    if (!aggr->GetBaseReplicator())
```
dl239: add `{ }` if not in the same line with `if` statement

src/tablet/tablet_impl.cc
```diff
@@ -5501,22 +5533,74 @@ void TabletImpl::DropFunction(RpcController* controller, const openmldb::api::Dr
...
+    std::string aggr_info;
+    google::protobuf::TextFormat::PrintToString(*request, &aggr_info);
+    std::string aggr_info_path = aggr_table_path + "/aggr_info.txt";
+
+    FILE* fd_write = fopen(aggr_info_path.c_str(), "w");
+    if (fd_write == NULL) {
+        PDLOG(WARNING, "fail to open file %s. err[%d: %s]", aggr_info_path.c_str(), errno, strerror(errno));
+        response->set_code(::openmldb::base::ReturnCode::kError);
+        response->set_msg("open aggr info file failed");
+        return;
```
dl239, line 5781: if run failed after `CreateAggregatorInternal`, need not to delete aggreator ?

src/storage/aggregator.h
```diff
@@ -124,6 +174,7 @@ class Aggregator {
  private:
     virtual bool UpdateAggrVal(const codec::RowView& row_view, const int8_t* row_ptr, AggrBuffer* aggr_buffer) = 0;
     virtual bool EncodeAggrVal(const AggrBuffer& buffer, std::string* aggr_val) = 0;
+    virtual bool DecodeAggrVal(AggrBuffer* buffer, const int8_t* row_ptr) = 0;
```
dl239: put output parameter after input parameter

src/storage/aggregator.cc
```diff
@@ -143,48 +156,136 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
     return true;
 }
 
-bool Aggregator::GetAggrBuffer(const std::string& key, AggrBuffer* buffer) {
+bool Aggregator::Init() {
+    std::unique_lock<std::mutex> lock(mu_);
+    if (GetStat() != AggrStat::kUnInit) {
+        PDLOG(INFO, "aggregator status is %u", GetStat());
```
dl239: It's to better to print readable status string

src/storage/aggregator.cc
```diff
@@ -143,48 +156,136 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
...
+        aggr_latest_offset = std::max(aggr_latest_offset, buffer.binlog_offset_);
+        int64_t latest_ts = buffer.ts_end_ + 1;
+        uint64_t latest_binlog = buffer.binlog_offset_ + 1;
+        buffer.clear();
+        buffer.ts_begin_ = latest_ts;
+        buffer.binlog_offset_ = latest_binlog;
+        if (window_type_ == WindowType::kRowsRange) {
+            buffer.ts_end_ = latest_ts + window_size_ - 1;
+        }
+        aggr_buffer_map_.emplace(it->GetPK(), std::move(tmp_buffer));
```
dl239: `mutex` in AggrBufferLocked cannot to copy or move. refer [here](https://en.cppreference.com/w/cpp/thread/mutex)

src/storage/aggregator.cc
```diff
@@ -143,48 +156,136 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
...
+        bool ok = entry.ParseFromString(record.ToString());
+        if (!ok) {
+            PDLOG(WARNING, "parse binlog failed");
+            continue;
+        }
+        if (cur_offset >= entry.log_index()) {
+            continue;
+        }
+
+        if (entry.has_method_type() && entry.method_type() == ::openmldb::api::MethodType::kDelete) {
```
dl239, line 296: not tackle the delete ?

src/storage/aggregator.cc
```diff
@@ -143,48 +156,136 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
...
+                break;
+            }
+        }
+        cur_offset = entry.log_index();
+    }
+    if (cur_offset < aggr_latest_offset) {
+        PDLOG(ERROR, "base table is slower than aggregator");
+        status_.store(AggrStat::kUnInit, std::memory_order_relaxed);
+        return false;
+    }
```
dl239, line 313: need not to recover from snapshot ?

nautaa: > need not to recover from snapshot ?

 binlog is enough for recover?

src/storage/aggregator.cc
nautaa: yep, delete will support in #1673.

src/storage/aggregator.cc
```diff
@@ -219,6 +341,13 @@ bool Aggregator::FlushAggrBuffer(const std::string& key, const AggrBuffer& buffe
         PDLOG(ERROR, "Aggregator put failed");
         return false;
     }
+    ::openmldb::api::LogEntry entry;
+    entry.set_pk(key);
+    entry.set_ts(time);
+    entry.set_value(encoded_row);
+    entry.set_term(aggr_replicator_->GetLeaderTerm());
+    entry.mutable_dimensions()->CopyFrom(dimensions_);
+    aggr_replicator_->AppendEntry(entry);
```
zhanghaohit, line 380: have to mimic the behavior for `Tablet::Put`:
```c++
        if (FLAGS_binlog_notify_on_put) {
            aggr_replicator_->Notify();
        }
```

src/storage/aggregator.cc
zhanghaohit: There is a potential issue here. As some old binlog files may be deleted after snapshotting, there may be a scenario where some entries that are required for aggr recover are deleted from binlog (but exists in the snapshot).

Maybe a quickfix is:
- whenever tablet doing `SchedDelBinlog`, we force flushing the aggregators (but still keep the memory states for newer updates). 
- Later flushing triggered by `CheckBufferFilled` will override or update the aggr result in the aggr table.

When we change the aggregator update to be async, we can have a nicer strategy to guarantee this.

What do you think? @dl239 @nautaa

src/storage/aggregator.cc
```diff
@@ -27,13 +30,35 @@ namespace openmldb {
 namespace storage {
 
 using ::openmldb::base::StringCompare;
+
+std::string AggrStatToString(uint32_t type) {
```
zhanghaohit: why not use `AggrStat`?

src/storage/aggregator.h
```diff
@@ -113,7 +160,10 @@ class Aggregator {
     std::mutex mu_;
     DataType aggr_col_type_;
     DataType ts_col_type_;
+    std::shared_ptr<LogReplicator> base_replicator_;
     std::shared_ptr<Table> aggr_table_;
+    std::shared_ptr<LogReplicator> aggr_replicator_;
+    std::atomic<uint32_t> status_;
```
zhanghaohit: can use `std::atomic<AggrStat>`?

src/storage/aggregator.cc
```diff
@@ -399,12 +562,66 @@ bool MinMaxBaseAggregator::EncodeAggrVal(const AggrBuffer& buffer, std::string*
     return true;
 }
 
+bool MinMaxBaseAggregator::DecodeAggrVal(const int8_t* row_ptr, AggrBuffer* buffer) {
+    char* aggr_val = NULL;
+    uint32_t ch_length = 0;
```
zhanghaohit, line 602: can we also add the checking for the `ch_length`?

src/storage/aggregator.h
```diff
@@ -64,8 +74,32 @@ struct AggrBuffer {
...
+            }
+        }
+    }
+    AggrBuffer& operator=(const AggrBuffer& buffer) = delete;
+    ~AggrBuffer() { clear(); }
     void clear() {
+        if (data_type_ == DataType::kString || data_type_ == DataType::kVarchar) {
+            if (aggr_val_.vstring.data != NULL) {
+                delete[] aggr_val_.vstring.data;
+                aggr_val_.vstring.data = NULL;
```
zhanghaohit: also reset the `vstring.len`

src/storage/aggregator.h
```diff
@@ -101,7 +138,17 @@ class Aggregator {
 
     uint32_t GetWindowSize() const { return window_size_; }
 
-    bool GetAggrBuffer(const std::string& key, AggrBuffer* buffer);
+    uint32_t GetStat() const { return status_.load(std::memory_order_relaxed); }
```
zhanghaohit: can return the `AggrStat` directly?

src/tablet/tablet_impl.cc
```diff
@@ -3036,6 +3036,40 @@ int TabletImpl::LoadTableInternal(uint32_t tid, uint32_t pid, std::shared_ptr<::
...
+                    } else {
+                        std::string msg;
+                        bool ok = CreateAggregatorInternal(&request, msg);
+                        if (!ok) {
+                            PDLOG(WARNING, "create aggregator failed. msg %s", msg.c_str());
+                        }
+                    }
+                }
+            }
+            // init aggregator related to base table if need.
```
zhanghaohit: the following code should be wrapped in the `else` block?

zhanghaohit (changes requested):

src/storage/aggregator.cc
```diff
@@ -143,48 +177,159 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
     return true;
 }
 
-bool Aggregator::GetAggrBuffer(const std::string& key, AggrBuffer* buffer) {
+bool Aggregator::FlushAll() {
     std::lock_guard<std::mutex> lock(mu_);
```
zhanghaohit: oh. cannot use the global lock to lock and flushing. Blocking all the updates... 

It's not a good choice to do the flushing. let me think a better workaround

src/storage/aggregator.cc
```diff
@@ -143,48 +177,159 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
...
+            return false;
+        }
+        int64_t latest_ts = aggr_buffer.ts_end_ + 1;
+        uint64_t latest_binlog = aggr_buffer.binlog_offset_ + 1;
+        aggr_buffer.clear();
+        aggr_buffer.ts_begin_ = latest_ts;
+        aggr_buffer.binlog_offset_ = latest_binlog;
+        if (window_type_ == WindowType::kRowsRange) {
+            aggr_buffer.ts_end_ = latest_ts + window_size_ - 1;
+        }
```
zhanghaohit: no need to do the reset for force flushing

zhanghaohit:
The reason why we are using binlog to do the recover of agg table, is the out of order issue which updates the already flushed results. Binlog cleaning makes it not easy to do it correct. As the deleted binlog may contains the data of the volatile AggrBuffer. `FlushAll` is heavy and definitely blocking the normal `update`.

Why not we do it in two steps:
- recover the volatile AggrBuffer using the base table directly (can use `WindowIterator`)
- recover the out of order cases using the binlog for the data that are already persistent in the aggr table (only recover the data before the `start_ts` of current AggrBuffer) - Here the binlog only is enough as the deleted binlog are guaranteed already applied and persistent in the aggr table.

src/storage/aggregator.cc
```diff
@@ -143,48 +177,159 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
...
-        case DataType::kInt:
-        case DataType::kTimestamp:
-        case DataType::kBigInt: {
-            int64_t origin_val = *reinterpret_cast<int64_t*>(ch);
-            buffer->aggr_val_.vlong = origin_val;
-            break;
+    status_.store(AggrStat::kRecovering, std::memory_order_relaxed);
+    auto log_parts = base_replicator_->GetLogPart();
+
+    if (aggr_table_->GetRecordCnt() == 0 && log_parts->IsEmpty()) {
```
zhanghaohit, line 219: what if the aggregator is newly-created, but the binlog is not empty, i.e.,there is already data in the table before deployment. 

It seems we can also handle?

src/storage/aggregator.cc
```diff
@@ -143,48 +177,159 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
...
+        status_.store(AggrStat::kInited, std::memory_order_relaxed);
+        return true;
+    }
+    auto it = aggr_table_->NewTraverseIterator(0);
+    it->SeekToFirst();
+    uint64_t recovery_offset = UINT64_MAX;
```
zhanghaohit, line 225: what if the `aggr_table_` is empty?  `recovery_offset` is kept `UINT64_MAX`. We read nothing from the binlog?

src/tablet/tablet_impl_test.cc
```diff
@@ -6316,6 +6330,146 @@ TEST_F(TabletImplTest, CreateAggregator) {
     }
 }
 
+TEST_F(TabletImplTest, AggregatorRecovery) {
```
zhanghaohit, line 6410: add multiple dimensions case.

each dimension has a corresponding deploy

src/storage/aggregator.cc
```diff
@@ -143,48 +177,159 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
...
+            continue;
+        }
+
+        // TODO(nauta): When the base table key is deleted, the pre-aggr table needs to be deleted at the same time.
+        if (entry.has_method_type() && entry.method_type() == ::openmldb::api::MethodType::kDelete) {
+            PDLOG(WARNING, "unsupport delete method for pre-aggr table");
+            continue;
+        }
+        for (int i = 0; i < entry.dimensions_size(); i++) {
+            const auto& dimension = entry.dimensions(i);
```
zhanghaohit, line 301: Here I think there is a bug in the aggregation logics (also in [UpdateAggrs](https://github.com/4paradigm/OpenMLDB/blob/47dc3a2cb0007b503eece19f78c134c1c7a2ceed/src/tablet/tablet_impl.cc#L4100)). Each dimension corresponds to one index (i.e., `index_pos` of `CreateAggregatorRequest`).

So different dimension should corresponds to different aggregators. We can only update its related dimension.

In the `TabletImpl::aggregators_`, we should keep the mapping of (tid, pid, index_pos) -> aggregator. Each `dimension` only should be updated in its corresponding aggregator.

src/tablet/tablet_impl_test.cc
zhanghaohit: more ut here:
- out of order cases
- empty `aggr table`
- some binlog files were deleted

zhanghaohit (changes requested):

src/storage/aggregator.cc
zhanghaohit: Double check. current impl should be ok. already do the checking `dimension.idx()` == `index_pos_`

src/storage/aggregator.h
```diff
@@ -97,7 +97,6 @@ class AggrBuffer {
         if (data_type_ == DataType::kString || data_type_ == DataType::kVarchar) {
             if (aggr_val_.vstring.data != NULL) {
                 delete[] aggr_val_.vstring.data;
-                aggr_val_.vstring.data = NULL;
```
zhanghaohit: why remove the resetting?

also resetting `vstring.len`

src/tablet/tablet_impl.cc
```diff
@@ -3036,6 +3036,41 @@ int TabletImpl::LoadTableInternal(uint32_t tid, uint32_t pid, std::shared_ptr<::
...
+                }
+            } else {
+                // init aggregator related to base table if need.
+                auto aggrs = GetAggregators(tid, pid);
+                if (aggrs != nullptr) {
+                    for (auto& aggr : *aggrs) {
+                        if (!aggr->GetBaseReplicator()) {
+                            aggr->SetBaseReplicator(replicator);
+                        }
+                        if (!aggr->Init()) {
```
zhanghaohit: here we can call `aggr->Init(replicator)` directly? No need the `GetBaseReplicator` and `SetBaseReplicator` APIs?

src/storage/aggregator.cc
nautaa: we do the aggregator `FlushAll` only if there are binlog files deleted. This seems acceptable? 
Maybe in the future a unified interface can be used to restore tables and aggregators from binlog and snapshot.

src/storage/aggregator.cc
```diff
@@ -120,9 +146,21 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
...
+            PDLOG(ERROR, "logical error: current offset is smaller than binlog offset");
+            return false;
+        }
+    }
+
     if (cur_ts < aggr_buffer.ts_begin_) {
         // handle the case that the current timestamp is smaller than the begin timestamp in aggregate buffer
         lock.unlock();
+        if (recover) { // avoid out-of-order duplicate writes during the recovery phase
+            return true;
```
zhanghaohit, line 163: cpplint

src/storage/aggregator.cc
```diff
@@ -143,48 +181,161 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
     return true;
 }
 
-bool Aggregator::GetAggrBuffer(const std::string& key, AggrBuffer* buffer) {
+bool Aggregator::FlushAll() {
+    //TODO(nauta): optimize the flush process
```
zhanghaohit: cpplint

src/tablet/tablet_impl_test.cc
```diff
@@ -6391,6 +6405,214 @@ TEST_F(TabletImplTest, CreateAggregator) {
...
+        TabletImpl tablet;
+        tablet.Init("");
+        ::openmldb::api::TableMeta base_table_meta;
+        // base table
+        uint32_t id = counter++;
+        base_table_id = id;
+        ::openmldb::api::CreateTableRequest request;
+        ::openmldb::api::TableMeta* table_meta = request.mutable_table_meta();
+        table_meta->set_tid(id);
+        AddDefaultAggregatorBaseSchema(table_meta);
```
zhanghaohit, line 6423: make it a multiple-index table? so dimensions > 1

zhanghaohit (changes requested):
some minor comments

fix the cpplint

src/tablet/tablet_impl_test.cc
```diff
@@ -181,7 +181,8 @@ void AddDefaultAggregatorBaseSchema(::openmldb::api::TableMeta* table_meta) {
     SchemaCodec::SetColumnDesc(table_meta->add_column_desc(), "col3", openmldb::type::DataType::kInt);
     SchemaCodec::SetColumnDesc(table_meta->add_column_desc(), "col4", openmldb::type::DataType::kDouble);
 
-    SchemaCodec::SetIndex(table_meta->add_column_key(), "idx", "id", "ts_col", ::openmldb::type::kAbsoluteTime, 0, 0);
+    SchemaCodec::SetIndex(table_meta->add_column_key(), "idx1", "id", "ts_col", ::openmldb::type::kAbsoluteTime, 0, 0);
+    SchemaCodec::SetIndex(table_meta->add_column_key(), "idx2", "col3", "ts_col", ::openmldb::type::kAbsoluteTime, 0, 0);
```
zhanghaohit: cpplint

zhanghaohit (approved):
LGTM

src/tablet/tablet_impl.cc
nautaa: @dl239 If it fails, the else block will not be executed.

dl239 (approved):

PR #1660, opened 2022-04-19
fix: memroy leak in zk_client.cc

### LINKED ISSUE #1659: Memory leak in zk_client.cc
**Bug Description**
`zoo_get_children` will allocate memory in `struct String_vector`, we need to use `deallocate_String_vector` to free memory by ourself.

https://github.com/4paradigm/OpenMLDB/blob/4403222f75820d971bf00ed7c0af4c73391eb6b1/src/zk/zk_client.cc#L498
https://github.com/4paradigm/OpenMLDB/blob/4403222f75820d971bf00ed7c0af4c73391eb6b1/src/zk/zk_client.cc#L523

There are two place to call `zoo_get_children`, I think we can add `deallocate_String_vector(&data)` at line 505 and line 530 to fix this bug.

**Expected Behavior**

**Steps to Reproduce**

1. 
2. 
3. 
4.

wuxiaobai24:
Fix memory leak in zk_client.cc.

It should resolve #1659

### DIFF
--- a/src/zk/zk_client.cc
+++ b/src/zk/zk_client.cc
@@ -23,6 +23,7 @@
 #include "base/strings.h"
 #include "boost/algorithm/string.hpp"
 #include "boost/lexical_cast.hpp"
+#include "absl/cleanup/cleanup.h"
 
 namespace openmldb {
 namespace zk {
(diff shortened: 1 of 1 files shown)


dl239 (approved):

src/zk/zk_client.cc
```diff
@@ -503,6 +503,7 @@ bool ZkClient::GetChildrenUnLocked(const std::string& path, std::vector<std::str
     for (int32_t i = 0; i < data.count; i++) {
         children.push_back(std::string(data.data[i]));
     }
+    deallocate_String_vector(&data);
```
aceforeverd: how about free in RAII way, in case program accidentally return before e.g in line 504. 
Use `abs::Cleanup`: https://github.com/abseil/abseil-cpp/blob/master/absl/cleanup/cleanup.h#L81

wuxiaobai24: ok

aceforeverd (approved):

aceforeverd:
thank you！

PR #1665, opened 2022-04-20
refactor: load dynamic lib in hybridse

dl239:
java sdk cannot extract function symbols from dynamic lib, so we should  load so in hybridse engine

### DIFF
--- a/hybridse/src/udf/dynamic_lib_manager.h
+++ b/hybridse/src/udf/dynamic_lib_manager.h
@@ -43,9 +43,7 @@ class DynamicLibManager {
 
     base::Status ExtractFunction(const std::string& name, bool is_aggregate,
             const std::string& file, std::vector<void*>* funs) {
-        if (funs == nullptr) {
-            return {common::kExternalUDFError, "funs is nullptr"};
-        }
+        CHECK_TRUE(funs != nullptr, common::kExternalUDFError, "funs is nullptr")
         std::shared_ptr<DynamicLibHandle> so_handle;
         {
             std::lock_guard<std::mutex> lock(mu_);
@@ -57,9 +55,7 @@ class DynamicLibManager {
         }
         if (!so_handle) {
             void* handle = dlopen(file.c_str(), RTLD_LAZY);
-            if (handle == nullptr) {
-                return {common::kExternalUDFError, "can not open the dynamic library: " + file};
-            }
+            CHECK_TRUE(handle != nullptr, common::kExternalUDFError, "can not open the dynamic library: " + file)
             so_handle = std::make_shared<DynamicLibHandle>(handle);
             std::lock_guard<std::mutex> lock(mu_);
             handle_map_.emplace(file, so_handle);
--- a/hybridse/src/udf/dynamic_lib_manager.h
+++ b/hybridse/src/udf/dynamic_lib_manager.h
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2021 4Paradigm
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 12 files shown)


hybridse/src/udf/dynamic_lib_manager.h
```diff
@@ -0,0 +1,123 @@
...
+
+namespace hybridse {
+namespace udf {
+
+struct DynamicLibHandle {
+    explicit DynamicLibHandle(void* ptr) {
+        handle = ptr;
+        ref_cnt = 1;
+    }
+    void* handle = nullptr;
```
aceforeverd, line 36: suffix `_`

hybridse/src/udf/dynamic_lib_manager.h
```diff
@@ -0,0 +1,123 @@
...
+ public:
+    DynamicLibManager() = default;
+
+    base::Status ExtractFunction(const std::string& name, bool is_aggregate, const std::string& file, std::vector<void*>* funs) {
+        if (funs == nullptr) {
+            return {common::kExternalUDFError, "funs is nullptr"};
+        }
+        std::shared_ptr<DynamicLibHandle> so_handle;
+        {
+            std::lock_guard<std::mutex> lock(mu_);
```
aceforeverd, line 49: should possible use read_write lock ?  like https://github.com/4paradigm/OpenMLDB/blob/e2b0b51bbcb7381c45723c96112ada171badaf22/src/statistics/query_response_time/deploy_query_response_time.h#L102-L103

hybridse/src/udf/dynamic_lib_manager.h
```diff
@@ -0,0 +1,123 @@
...
+            if (dlclose(so_handle->handle) != 0) {
+                return {common::kExternalUDFError, "dlclose run error. file is " + file};
+            }
+        }
+        return {};
+    }
+
+ private:
+    std::mutex mu_;
+    std::map<std::string, std::shared_ptr<DynamicLibHandle>> handle_map_;
```
aceforeverd, line 115: cpplint

hybridse/src/udf/udf_library.cc
```diff
@@ -197,18 +197,41 @@ UdafRegistryHelper UdfLibrary::RegisterUdaf(const std::string& name) {
...
     if (HasFunction(canon_name)) {
         return {kCodegenError, name + " has exist"};
     }
+    std::vector<void*> funs;
+    if (file.empty()) {
+        // use trivial_fun for compile only
+        funs.emplace_back(reinterpret_cast<void*>(udf::v1::trivial_fun));
+    }  else {
+        auto status = lib_manager_.ExtractFunction(canon_name, is_aggregate, file, &funs);
+        if (!status.isOK()) {
```
aceforeverd: use marco `CHECK_STATUS` to preserve line info in trace

hybridse/src/udf/udf_library.cc
```diff
@@ -197,18 +197,41 @@ UdafRegistryHelper UdfLibrary::RegisterUdaf(const std::string& name) {
...
+            return status;
+        }
+    }
+    if (funs.empty() || funs[0] == nullptr) {
+        return {kCodegenError, name + " is nullptr"};
```
aceforeverd: same here

hybridse/src/udf/dynamic_lib_manager.h
```diff
@@ -0,0 +1,123 @@
...
+        return base::Status::OK();
+    }
+
+    base::Status RemoveHandler(const std::string& file) {
+        std::shared_ptr<DynamicLibHandle> so_handle;
+        {
+            std::lock_guard<std::mutex> lock(mu_);
+            auto iter = handle_map_.find(file);
+            if (iter != handle_map_.end()) {
+                iter->second->ref_cnt--;
```
aceforeverd, line 99: not thread safe. Does it matter ?

dl239: decrease is already in lock scope

hybridse/src/udf/dynamic_lib_manager.h
dl239: struct members need not end with '_'. refer [Variable_Names](https://google.github.io/styleguide/cppguide.html#Variable_Names)

hybridse/src/udf/dynamic_lib_manager.h
dl239: only used in create/drop function. it's not frequently to be called, so it's no matter what's mutex used

tobegit3hub (approved):
LGTM

aceforeverd (approved):

Issue #1639, opened 2022-04-15
feat: Add a UDF function similar to Hive get_json_object


kitty-eu-org:
**Describe the feature you'd like**

Add a UDF function similar to hive's get_json_object function, input json string, and output the parsed string

**Additional context**

<img width="1211" alt="image" src="https://user-images.githubusercontent.com/46663318/163578149-429859a2-44d9-42eb-8a27-7cb58bfea18e.png">

aceforeverd:
thank you. This will be part of built-in json functions

aceforeverd:
@hezhaozhao-git Hi, I'm adding the `get_object_function` in #3429, the main difference with hive's `get_json_object` is the second parameter, for openmldb, it is the JSON string representation from https://datatracker.ietf.org/doc/html/rfc6901.

 check it out if you're still interested :)

PR #1694, opened 2022-04-21
feat: add a udf  function similar to Hive get_json_object

kitty-eu-org:

feature

https://github.com/4paradigm/OpenMLDB/issues/1639

add a udf  function similar to Hive get_json_object


kitty-eu-org:
@aceforeverd this test error is not the code I submitted，cpplint error is the code of the 3rd party library I am referencing,What should I do to get rid of these errors？
image

aceforeverd:
> @aceforeverd this test error is not the code I submitted，cpplint error is the code of the 3rd party library I am referencing,What should I do to get rid of these errors？ image

HI, the PR is kind huge and not able to review. I advise you to rebase on the latest commit and keep the diff only relative to this PR title.
For thid-party, we put them in `thid-party` cmake module, third-party source itself will not included. Meanwhile, we encourage discussion over third-party additions before actual work. So yet another advise,  I think we should separate the PR into two: one for third-party change and another for json udf.
Besides, there are json functions not only `json_get_object`, but also `from_json`, `json_array_length` etc: [json functions](https://docs.databricks.com/sql/language-manual/sql-ref-functions-builtin.html#json-functions), so make a  separate `json_udf.cc` for those should necessary

kitty-eu-org:
@aceforeverd 
1. I want to introduce a modern C++ json parsing library, it is [nlohmann_json](https://github.com/nlohmann/json)
2. When I wrote the get_json_object function, I followed the specifications of the [openmldb document](https://openmldb.ai/docs/zh/v0.4/developer/built_in_function_develop_guide.html). There is no mention of the need to re-open a udf_xx.cc, and there is no guide for introducing third-party packages. I think these guides should be added to the document.
3. In response to what you said, I will try my best to modify it. I also think it is better to put the json function in another udf_json.cc file

lumianph:
> 

yes. the document is a bit simple. We should improve the doc after this case, to make it more friendly for developers @aceforeverd

@hezhaozhao-git thank you for your contribution very much, if any concerns, you may discuss with us offline

aceforeverd:
Thank you
> @aceforeverd
> 
> 1. I want to introduce a modern C++ json parsing library, it is [nlohmann_json](https://github.com/nlohmann/json)

nlohmann seems great for me, but how is it compared to other json library ? e.g jsoncpp, rapidjson or even jq, and we're using rapidjson elsewhere @vagetablechicken 

> 2. When I wrote the get_json_object function, I followed the specifications of the [openmldb document](https://openmldb.ai/docs/zh/v0.4/developer/built_in_function_develop_guide.html). There is no mention of the need to re-open a udf_xx.cc, and there is no guide for introducing third-party packages. I think these guides should be added to the document.

Yes, it should improved. Json functions is type of functions same as window function, we define those in `hybridse/src/udf/default_defs`

> 3. In response to what you said, I will try my best to modify it. I also think it is better to put the json function in another udf_json.cc file

kitty-eu-org:
For the comparison of the [nlohmann_json](https://github.com/nlohmann/json) library, you can see this [benchmark](https://github.com/miloyip/nativejson-benchmark),Its performance is strong enough and has good ease of use

PR #1737, opened 2022-04-29
fix: update ttl if new ttl greater than current ttl when depoly sql

### LINKED ISSUE #1131: Auto Create table throw error when new ttl greater than current ttl

Now OpenMLDB will auto-create a table when we deploy SQL in CLI. So when the TTL > the original table's TTL, we should update the TTL or throw error if we can't support ttl update
## Steps to Reproduce

1.
2.
3.
4.

keyu813:
close #1131


src/sdk/sql_cluster_router.cc
```diff
@@ -2993,7 +2993,49 @@ hybridse::sdk::Status SQLClusterRouter::HandleDeploy(const hybridse::node::Deplo
...
+                    ::openmldb::type::TTLType type = column_key.ttl().ttl_type();
+                    uint64_t old_ttl = 0;
+                    uint64_t new_ttl = 0;
+                    if (type == ::openmldb::type::TTLType::kAbsoluteTime) {
+                        old_ttl = old_column_key.ttl().abs_ttl();
+                        new_ttl = column_key.ttl().abs_ttl();
+                    } else {
+                        old_ttl = old_column_key.ttl().abs_ttl();
+                        new_ttl = column_key.ttl().lat_ttl();
+                    }
```
dl239, line 3040: check `abs_ttl` and `lat_ttl` at same time

src/cmd/sql_cmd_test.cc
```diff
@@ -506,33 +564,34 @@ void CreateDBTableForLongWindow(const std::string& base_db, const std::string& b
...
         std::string date;
+        std::string ts_str = std::to_string(ts + i);
         if (i < 10) {
             date = absl::StrCat("1900-01-0", std::to_string(i));
         } else {
             date = absl::StrCat("1900-01-", std::to_string(i));
         }
         std::string insert =
-            absl::StrCat("insert into ", base_table, " values('str1', 'str2', ", val, ", ", val, ", ", val, ", ", val,
+            absl::StrCat("insert into ", base_table, " values('str1', 'str2', ", ts_str, ", ", val, ", ", val, ", ", val,
```
dl239: fix lint warnings

src/sdk/sql_cluster_router.cc
```diff
@@ -2951,7 +2951,60 @@ hybridse::sdk::Status SQLClusterRouter::HandleDeploy(const hybridse::node::Deplo
...
+                        new_lat_ttl = column_key.ttl().lat_ttl();
+                    }
+                    if (new_abs_ttl > old_abs_ttl || new_lat_ttl > old_lat_ttl) {
+                        // update ttl
+                        auto ns_ptr = cluster_sdk_->GetNsClient();
+                        std::string err;
+                        uint64_t abs_ttl = 0;
+                        uint64_t lat_ttl = 0;
+                        if (type == ::openmldb::type::TTLType::kAbsoluteTime) {
+                            abs_ttl = new_abs_ttl;
```
dl239: why not use `new_abs_ttl` and `new_lat_ttl` directly

src/cmd/sql_cmd_test.cc
```diff
@@ -402,6 +402,64 @@ TEST_P(DBSDKTest, Deploy) {
...
+    ASSERT_EQ(column_key.ts_name(), "c7");
+    ASSERT_TRUE(column_key.has_ttl());
+    ASSERT_EQ(column_key.ttl().ttl_type(), ::openmldb::type::TTLType::kLatestTime);
+    ASSERT_EQ(column_key.ttl().lat_ttl(), 2);
+
+    ASSERT_FALSE(cs->GetNsClient()->DropTable("test1", "trans", msg));
+    ASSERT_TRUE(cs->GetNsClient()->DropProcedure("test1", "demo", msg));
+    ASSERT_TRUE(cs->GetNsClient()->DropTable("test1", "trans", msg));
+    ASSERT_TRUE(cs->GetNsClient()->DropDatabase("test1", msg));
+}
```
dl239, line 478: add `ttl_type` mismatch case

src/sdk/sql_cluster_router.cc
```diff
@@ -2951,7 +2951,50 @@ hybridse::sdk::Status SQLClusterRouter::HandleDeploy(const hybridse::node::Deplo
                 }
             }
             if (index_id_set.count(openmldb::schema::IndexUtil::GetIDStr(column_key)) > 0) {
-                // skip exist index
+                // if type match && new ttl greater than old ttl, update ttl, else skip exist index
+                ::openmldb::common::ColumnKey old_column_key;
+                for (const auto& table_column_key : it->second.column_key()) {
```
vagetablechicken: give `it->second` a name, and `kv.second`.
Too confused.

src/sdk/sql_cluster_router.cc
```diff
@@ -2951,7 +2951,50 @@ hybridse::sdk::Status SQLClusterRouter::HandleDeploy(const hybridse::node::Deplo
...
+                    if (openmldb::schema::IndexUtil::GetIDStr(table_column_key) ==
+                        openmldb::schema::IndexUtil::GetIDStr(column_key)) {
+                        old_column_key = table_column_key;
+                        break;
+                    }
+                }
+                if (old_column_key.ttl().ttl_type() != column_key.ttl().ttl_type()) {
```
vagetablechicken, line 3008: Why not making a map <id, idx_in_column_key_list>, instead of set?

src/sdk/sql_cluster_router.cc
```diff
@@ -2951,7 +2951,50 @@ hybridse::sdk::Status SQLClusterRouter::HandleDeploy(const hybridse::node::Deplo
...
+                        // update ttl
+                        auto ns_ptr = cluster_sdk_->GetNsClient();
+                        std::string err;
+                        bool ok = ns_ptr->UpdateTTL(kv.first, type, new_abs_ttl, new_lat_ttl, column_key.index_name(), err);
+                        if (!ok) {
+                            return {::hybridse::common::StatusCode::kCmdError, "update ttl failed"};
+                        }
+                    }
+                }
                 continue;
```
vagetablechicken: continue what?

src/sdk/sql_cluster_router.cc
```diff
@@ -2951,7 +2951,50 @@ hybridse::sdk::Status SQLClusterRouter::HandleDeploy(const hybridse::node::Deplo
...
+                ::openmldb::common::ColumnKey old_column_key;
+                for (const auto& table_column_key : it->second.column_key()) {
+                    if (openmldb::schema::IndexUtil::GetIDStr(table_column_key) ==
+                        openmldb::schema::IndexUtil::GetIDStr(column_key)) {
+                        old_column_key = table_column_key;
+                        break;
+                    }
+                }
+                if (old_column_key.ttl().ttl_type() != column_key.ttl().ttl_type()) {
+                    return {::hybridse::common::StatusCode::kCmdError,
```
vagetablechicken, line 3009: if we have 'c1|ts|abs ttl'，we can't add a new 'c1|ts|lat ttl'？
And why return? The other indexes we want to add will be skip?

vagetablechicken (changes requested):

src/sdk/sql_cluster_router.cc
dl239: skip add to `new_indexs` which will be executing add index later

src/sdk/sql_cluster_router.cc
keyu813: > Why not making a map <id, idx_in_column_key_list>, instead of set?

also can use map, here we only need to make sure whether columnkey exist, so use set here

src/sdk/sql_cluster_router.cc
keyu813: two indexes will be consider as the same index if the column names and ts are same(see `GetIDStr()`). so we can't add a new 'c1|ts|lat ttl'.   Return because new index ttl type mismatch old index ttl type when update the same index

src/sdk/sql_cluster_router.cc
vagetablechicken: but you loop and find it in `for (const auto& table_column_key : it->second.column_key())`

src/sdk/sql_cluster_router.cc
vagetablechicken: Why are they the same? If you don't want two indexes, you should merge them to one, abs or lat

src/sdk/sql_cluster_router.cc
vagetablechicken: Not a good writing, @keyu813 extract the code  above `continue` to some methods. Make the logic more clear.

src/sdk/sql_cluster_router.cc
```diff
@@ -2921,39 +2921,42 @@ hybridse::sdk::Status SQLClusterRouter::HandleDeploy(const hybridse::node::Deplo
     auto index_map = base::DDLParser::ExtractIndexes(select_sql, table_schema_map);
     std::map<std::string, std::vector<::openmldb::common::ColumnKey>> new_index_map;
     for (auto& kv : index_map) {
-        auto it = table_map.find(kv.first);
+        std::string table_name = kv.first;
+        std::vector<::openmldb::common::ColumnKey> extract_column_keys = kv.second;
```
vagetablechicken, line 2967: it's copy, use ref, e.g. `auto&`

src/cmd/sql_cmd_test.cc
```diff
@@ -1534,6 +1615,7 @@ TEST_P(DBSDKTest, SelectWithAddNewIndex) {
                     absl::StrCat("insert into ", tb1_name, " values(2,'bb',1,1590738990000,1637056523316);"),
                     absl::StrCat("insert into ", tb1_name, " values(3,'aa',3,1590738990000,1637057123257);"),
                     absl::StrCat("insert into ", tb1_name, " values(4,'aa',1,1590738990000,1637057123317);"),
+                    // todo: add ts column will have wrong result (#issue1722)
```
dl239: issue 1722 have been closed. rm annotation

src/sdk/sql_cluster_router.h
```diff
@@ -354,6 +354,18 @@ class SQLClusterRouter : public SQLRouter {
...
     hybridse::sdk::Status HandleDeploy(const hybridse::node::DeployPlanNode* deploy_node);
 
+    hybridse::sdk::Status HandleIndex(const std::set<std::pair<std::string, std::string>>& table_pair,
+                                      const std::string& select_sql);
+
+    hybridse::sdk::Status GetNewIndex(
+        const std::map<std::string, ::openmldb::nameserver::TableInfo>& table_map,
+        const std::map<std::string, std::vector<::openmldb::common::ColumnKey>>& index_map,
+        std::map<std::string, std::vector<::openmldb::common::ColumnKey>>& new_index_map);
+
```
dl239, line 364: resolve lint warning

dl239 (approved):

src/sdk/sql_cluster_router.cc
```diff
@@ -2920,38 +2945,96 @@ hybridse::sdk::Status SQLClusterRouter::HandleDeploy(const hybridse::node::Deplo
...
-        for (const auto& column_desc : it->second.column_desc()) {
+        for (const auto& column_desc : table.column_desc()) {
             col_set.insert(column_desc.name());
         }
-        std::set<std::string> index_id_set;
-        for (const auto& column_key : it->second.column_key()) {
-            index_id_set.insert(openmldb::schema::IndexUtil::GetIDStr(column_key));
+        std::map<std::string, ::openmldb::common::ColumnKey> id_columnkey_map;
+        for (const auto& column_key : table.column_key()) {
+            id_columnkey_map.emplace(openmldb::schema::IndexUtil::GetIDStr(column_key), column_key);
```
vagetablechicken, line 2983: What if table info already have two indexes which are the same id?

src/sdk/sql_cluster_router.cc
```diff
@@ -2920,38 +2945,96 @@ hybridse::sdk::Status SQLClusterRouter::HandleDeploy(const hybridse::node::Deplo
...
                     return {::hybridse::common::StatusCode::kCmdError,
-                            "col " + col + " is not exist in table " + kv.first};
+                            "col " + col + " is not exist in table " + table_name};
                 }
             }
-            if (index_id_set.count(openmldb::schema::IndexUtil::GetIDStr(column_key)) > 0) {
-                // skip exist index
+            std::string index_id = openmldb::schema::IndexUtil::GetIDStr(column_key);
+            // index exist, if type match && new ttl greater than old ttl, update ttl, else skip
+            if (id_columnkey_map.find(index_id) != id_columnkey_map.end()) {
```
vagetablechicken: auto id_columnkey = id_columnkey_map.find();
if (it != end) {
  auto old = id_columnkey->second;
}

Avoid search twice.

vagetablechicken:
```
if index exists:
  UpdateIndex()
else
  AddIndex()
```
avoid continue

vagetablechicken (approved):

PR #1754, opened 2022-05-06
fix(grafana dashboard): datasource.uid should point to ${datasource}

aceforeverd:
instead specific uid string, for non-existed uid, grafana ui will default points to `${datasource}` menu, that's naughty.


dl239 (approved):

tobegit3hub (approved):
LGTM

PR #1790, opened 2022-05-10
refactor: check if null in PreparedStatement

### LINKED ISSUE #1759: check if null in PreparedStatement
https://github.com/4paradigm/OpenMLDB/blob/9389d90c260ec395c32cbae4d98a42a001183a16/java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/PreparedStatement.java#L274-L277

`currentRow` may be null, check it before call `GetSchema`. If it's null, throw an exception.

And then, Line 279 `this.currentRow == null` below is useless. Delete it.

https://github.com/4paradigm/OpenMLDB/blob/9389d90c260ec395c32cbae4d98a42a001183a16/java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/PreparedStatement.java#L279-L281

vighnesh-kadam:
PR checks if "currentRow" variable value is NULL and its checked  before calling  GetSchema method. If it's null, throws an exception.
Also another if statement previously on line 279 is deleted.

[https://github.com/4paradigm/OpenMLDB/issues/1759]

As directed by  @vagetablechicken  "if" statement is placed to check for null value of  "currentRow" variable and if value is null exception is thrown.

### DIFF
--- a/java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/PreparedStatement.java
+++ b/java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/PreparedStatement.java
@@ -273,12 +273,13 @@ public class PreparedStatement implements java.sql.PreparedStatement {
             }
             this.currentRow = SQLRequestRow.CreateSQLRequestRowFromColumnTypes(columnTypes);
             // TODO(hw): check if null
+            if (this.currentRow == null) {
+            throw new SQLException("fail to build data with null row");
+        }
             this.currentSchema = this.currentRow.GetSchema();
             this.orgTypes = this.types;
         }
-        if (this.currentRow == null) {
-            throw new SQLException("fail to build data with null row");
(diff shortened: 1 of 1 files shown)


java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/PreparedStatement.java
```diff
@@ -273,12 +273,13 @@ protected void dataBuild() throws SQLException {
             }
             this.currentRow = SQLRequestRow.CreateSQLRequestRowFromColumnTypes(columnTypes);
             // TODO(hw): check if null
+            if (this.currentRow == null) {
+            throw new SQLException("fail to build data with null row");
```
dl239, line 277: hi @vighnesh-kadam, thank you for your pr.  It need to format the `if` block here and remove the `TODO` note

PR #1802, opened 2022-05-11
fix: define the ordering for rows having same key/ts in window union but from different datasource

### LINKED ISSUE #1776: window with union: ordering for union rows and original rows with the same order key is undefined
**Bug Description**

in https://github.com/4paradigm/OpenMLDB/blob/361f44d433c2a7682a1ede4792afb8df4d1d5c74/hybridse/src/vm/runner.cc#L1530, union rows in rows filters out those with the same order key with current row, 

a correct case should be:
```yaml
- id: 66
    desc: 0s preceding
    inputs:
      - name: t1
        columns:
          - id int
          - ts timestamp
          - g int
        indexs:
          - idx:id:ts
        data: |
          1, 100, 111
      - name: t2
        columns:
          - ts timestamp
          - g int
          - val int
        indexs:
          - idx:g:ts
        data: |
          100, 111, 233
          100, 111, 17
          100, 111, 200
    # id, ts, g,   val
    # 1, 100, 111, 21
    # 1, 100, 111, 233
    sql: |
       select max(val) over w as mv
       from (select id, ts, g, 21 as val from t1)
       window w as(
       union (select 1 as id, ts, g, val from t2)
              partition by `g` order by `ts`
              rows_range between 1d preceding and 0s preceding EXCLUDE CURRENT_TIME INSTANCE_NOT_IN_WINDOW);
    expect:
      columns:
        - mv int
      order: mv
      data: |
        21
```

**Expected Behavior**

**Steps to Reproduce**

1. 
2. 
3. 
4.

aceforeverd:

close #1776 

### DIFF
--- a/cases/function/window/test_window_union.yaml
+++ b/cases/function/window/test_window_union.yaml
@@ -901,4 +901,68 @@ cases:
       data: |
         1, 6, 999, 0, 200, 233
         2, 7, 10000, 0, 21, 200
+  - id: 19-2
+    mode: cluster-unsupport
+    desc: |
+      rows order for pure history window union
+    inputs:
+      - name: t1
+        columns:
+          - id int
+          - ts timestamp
+          - g int
+          - val int
+        indexs:
+          - idx:g:ts
+        data: |
+          1, 100, 111, 21
+          2, 100, 111, 10000
+      - name: t2
+        columns:
+          - id int
+          - ts timestamp
+          - g int
+          - val int
+        indexs:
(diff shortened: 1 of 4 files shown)


tobegit3hub (approved):
LGTM

zhanghaohit (approved):

aceforeverd:
[Bugs (Normal)](https://trello.com/c/WhbjlRN2/69-bugs-normal)

Issue #1813, opened 2022-05-13
`docker run 4pdosc/openmldb` without command following return silently


Patrick-Star125:
**Bug Description**
在Ubuntu虚拟机安装openmldb无法运行，环境如下
1.Ubuntu 20.04 LTS
2.Docker 20.10.12
3.VMware虚拟机软件
4.配置的镜像源
~~~bash
{
  "registry-mirrors": [
    "https://hub-mirror.c.163.com",
    "https://ustc-edu-cn.mirror.aliyuncs.com",
    "https://ghcr.io",
    "https://mirror.baidubce.com"
  ]
}
~~~
执行docker run语句如下，有两种状态
~~~bash
//不添加restart
sudo docker run -d -p 5432:5432 \
		--name openmldb \
		-v /home/openmldb_data:/data \
		4pdosc/openmldb:0.4.4
~~~

用docker ps查看，直接退出，也没有端口映射
~~~bash
CONTAINER ID   IMAGE                   COMMAND   CREATED          STATUS                      PORTS     NAMES
da016ba693d5   4pdosc/openmldb:0.4.4   "bash"    14 seconds ago   Exited (0) 14 seconds ago             openmldb
~~~
再执行docker run语句如下
~~~bash
//添加restart
sudo docker run -d -p 5432:5432 \
		--name openmldb \
		--restart=unless-stopped \
		-v /home/openmldb_data:/data \
		4pdosc/openmldb:0.4.4
~~~
用docker ps查看，持续重启，也没有端口映射
~~~bash
CONTAINER ID   IMAGE                   COMMAND   CREATED          STATUS                          PORTS     NAMES
8bfad46f0786   4pdosc/openmldb:0.4.4   "bash"    11 minutes ago   Restarting (0) 12 seconds ago             openmldb
~~~
复现过程如上
另外，进入挂载出来的`/home/openmldb_data`中也没有任何文件，挂载失败

lumianph:
@vagetablechicken  you may also help take a look

aceforeverd:
1. `docker inspect 4pdosc/openmldb:0.4.4` to see if digest matches https://hub.docker.com/layers/openmldb/4pdosc/openmldb/0.4.4/images/sha256-32d8854fe8fe43c5a4479ea51179a0c2a9441ed0726f4dd27d882cab2438dd2f?context=explore
2. is there any log in `docker logs <container_id>` ?

Patrick-Star125:
[...]

PR #1816, opened 2022-05-13
feat: add a default CMD for openmldb docker

aceforeverd:
help prevent #1813


vagetablechicken (approved):

tobegit3hub (approved):
LGTM

vagetablechicken:
OpenMLDB Docker ci is still failed, maybe we should sleep a while after init?

And the `OPENMLDB_VERSION ` we set before workflow dispatch won't be the OPENMLDB_VERSION in dockerfile? It's only the version of docker image.

PR #1820, opened 2022-05-16
fix(#1761): explain require next <cr> and space ending problem

### LINKED ISSUE #1761: Solve problems in SQL commands
I find two problems when using the explain+sql commands via terminal

(1) Need to click the "enter" key twice before seeing the explain results:

127.0.0.1:6527/test> explain select * from test;
127.0.0.1:6527/test> 	DATA_PROVIDER(request=test)

(2) Need to split the sql with the semicolon (;) with a character (e.g., an enter key or space), otherwise it will report errors:

127.0.0.1:6527/test> select * from (select * from
                  ->  window win1 as (partition by c1 order by c2 rows_range between 172799999 preceding and 0s preceding MAXSIZE 1000),
                  -> win2 as (partition by c1 order by c2 rows_range between 172799999 preceding and 0s preceding MAXSIZE 1000))
                  -> as out0
                  ->  last join (select * from test t1 last join test t2 on t1.c2=t2.c2) as out1 on out0.c2=out1.c2; 
                  -> 
                  -> ;
Error: Syntax error: Expected end of input but got ";" [at 6:1]

127.0.0.1:6527/test> select * from (select * from
                  ->  window win1 as (partition by c1 order by c2 rows_range between 172799999 preceding and 0s preceding MAXSIZE 1000),
                  -> win2 as (partition by c1 order by c2 rows_range between 172799999 preceding and 0s preceding MAXSIZE 1000))
                  -> as out0
                  ->  last join (select * from test t1 last join test t2 on t1.c2=t2.c2) as out1 on out0.c2=out1.c2
                  -> ;
 ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- 
  c1   c2   c3   c6   c7   c1   c2   c3   c6   c7   c1   c2   c3   c6   c7  
 ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ----

aceforeverd:
close #1761 

two problem in cli:
1. explain result missing a `<CR>`
2. fail to precess ending space

### DIFF
--- a/src/cmd/sql_cmd.h
+++ b/src/cmd/sql_cmd.h
@@ -73,7 +73,7 @@ void HandleSQL(const std::string& sql) {
                 while (result_set->Next()) {
                     std::string val;
                     result_set->GetAsString(0, val);
-                    std::cout << val;
+                    std::cout << val << std::endl;
                 }
             } else {
                 ::hybridse::base::TextTable t('-', ' ', ' ');
--- a/src/cmd/sql_cmd.h
+++ b/src/cmd/sql_cmd.h
@@ -63,6 +63,27 @@ const std::string VERSION = std::to_string(OPENMLDB_VERSION_MAJOR) + "." +  // N
 ::openmldb::sdk::DBSDK* cs = nullptr;
 ::openmldb::sdk::SQLClusterRouter* sr = nullptr;
 
+// strip any whitespace characters begining of the last unfinished statement from `input` (string after last semicolon)
+// final SQL strings are appended into `output`
+//
+// this help handle SQL strings that has space trailing but do not expected to be a statement after semicolon
+void StripStartingSpace(const absl::string_view input, std::string* output) {
+    auto last_semicolon_pos = input.find_last_of(';');
+    if (last_semicolon_pos != std::string::npos && input.back() != ';') {
+        absl::string_view last_stmt = input;
+        last_stmt.remove_prefix(last_semicolon_pos + 1);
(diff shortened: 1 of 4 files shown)


src/cmd/sql_cmd.h
```diff
@@ -73,7 +94,7 @@ void HandleSQL(const std::string& sql) {
                 while (result_set->Next()) {
                     std::string val;
                     result_set->GetAsString(0, val);
-                    std::cout << val;
+                    std::cout << val << std::endl;
```
aceforeverd: check `desc table`

src/cmd/sql_cmd_test.cc
```diff
@@ -1969,6 +1969,27 @@ TEST_P(DBSDKTest, DeployStatsOnlyCollectDeployProcedure) {
     }
 }
 
+class StripSpaceTest : public ::testing::TestWithParam<std::pair<std::string_view, std::string_view>> {};
+
+std::vector<std::pair<std::string_view, std::string_view>> strip_cases = {
```
vagetablechicken, line 1974: how about "show\ncomponents\n;"

aceforeverd: statement str inside is handle by zetasql, should be fine.

anyway case like `show components; set val = ';    '` probably not work as expect

aceforeverd:
[Bugs (Normal)](https://trello.com/c/WhbjlRN2/69-bugs-normal)

dl239 (approved):

vagetablechicken (approved):

Issue #1296, opened 2022-02-23
Support long time window with filters


zhanghaohit:
**Is your feature request related to a problem? Please describe.**
long-time window extension

**Describe the solution you'd like**
- pre-aggregation table are partitioned based on the condition
- `reduce` Node should be extended to support partitioned pre-aggregation tables

**Additional context**
e.g., sql query may look like this.
```sql
SELECT count_where(col1, col2=0) over w1 FROM t1
WINDOW w1 AS (PARTITION BY id ORDER BY ts ROWS_RANGE BETWEEN 10d PRECEDING AND CURRENT ROW);
```

PR #1821, opened 2022-05-16
feat: support `count_where` op in aggregator

nautaa:
feature

#1296
support `count_where` op in aggregator.

in this pr, we only support `binary expr with single column` in `where condition` of `count_where`.
eg.
``` 
count_where(a, b>3)
count_where(a, b='id')
```
the `aggr_val` is updated with the same `key` and `filter col`.


src/base/ddl_parser_test.cc
```diff
@@ -651,6 +651,28 @@ TEST_F(DDLParserTest, extractLongWindow) {
...
+        ASSERT_TRUE(extract_status.IsOK());
+        ASSERT_EQ(window_infos.size(), 1);
+        ASSERT_EQ(window_infos[0].window_name_, "w1");
+        ASSERT_EQ(window_infos[0].aggr_func_, "count_where");
+        ASSERT_EQ(window_infos[0].aggr_col_, "c3");
+        ASSERT_EQ(window_infos[0].partition_col_, "c1");
+        ASSERT_EQ(window_infos[0].order_col_, "c6");
+        ASSERT_EQ(window_infos[0].bucket_size_, "1000");
+        ASSERT_EQ(window_infos[0].filter_col_, "c1");
+    }
```
zhanghaohit, line 697: add cases:
- `count_where(c3, 2=c1)`
- unsupported/fail case cases

src/sdk/sql_cluster_router.cc
```diff
@@ -3199,20 +3200,21 @@ hybridse::sdk::Status SQLClusterRouter::HandleLongWindows(
...
-                absl::StrCat("pre_", deploy_node->Name(), "_", lw.window_name_, "_", lw.aggr_func_, "_", aggr_col);
+                absl::StrCat("pre_", deploy_node->Name(), "_", lw.window_name_, "_", lw.aggr_func_, "_", aggr_col,
+                             lw.filter_col_.empty() ? "" : "_" + lw.filter_col_);
             ::hybridse::sdk::Status status;
             std::string insert_sql =
                 absl::StrCat("insert into ", meta_db, ".", meta_table, " values('" + aggr_table, "', '", aggr_db,
                              "', '", base_db, "', '", base_table, "', '", lw.aggr_func_, "', '", lw.aggr_col_, "', '",
-                             lw.partition_col_, "', '", lw.order_col_, "', '", lw.bucket_size_, "');");
+                             lw.partition_col_, "', '", lw.order_col_, "', '", lw.bucket_size_, "', '",
+                             lw.filter_col_.empty() ? "NULL" : lw.filter_col_, "');");
```
zhanghaohit: shall we insert an empty string `""` or `NULL` instead of `"NULL"`?

src/storage/aggregator.cc
```diff
@@ -350,28 +354,40 @@ bool Aggregator::GetAggrBufferFromRowView(const codec::RowView& row_view, const
     return true;
 }
 
-bool Aggregator::FlushAggrBuffer(const std::string& key, const AggrBuffer& buffer) {
+bool Aggregator::FlushAggrBuffer(const std::string& aggr_key, const AggrBuffer& buffer) {
```
zhanghaohit: why not pass both `key` and `filter_key`? then no need to parse and copy the string here

src/storage/aggregator.h
```diff
@@ -143,6 +155,8 @@ class Aggregator {
 
     bool GetAggrBuffer(const std::string& key, AggrBuffer** buffer);
 
+    virtual std::string GetAggregateKey(const std::string& pk, int8_t* /* row_ptr */) { return pk; }
```
zhanghaohit: every calling will do a string copy.

src/storage/aggregator.cc
```diff
@@ -115,13 +116,15 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
             return false;
         }
     }
+    auto aggr_key = GetAggregateKey(key, row_ptr);
```
zhanghaohit: can we avoid to do the `combination`? 
- frequent str concat and split is costly
- `key_end` seems not straightforward

One alternative is two-level mapping: {key -> {filter, AggreBuffer}}

src/storage/aggregator_test.cc
```diff
@@ -266,6 +267,33 @@ void CheckAvgAggrResult(std::shared_ptr<Table> aggr_table, DataType data_type, i
...
+        auto tmp_val = it->GetValue();
+        std::string origin_data = tmp_val.ToString();
+        codec::RowView origin_row_view(aggr_table->GetTableMeta()->column_desc(),
+                                       reinterpret_cast<int8_t*>(const_cast<char*>(origin_data.c_str())),
+                                       origin_data.size());
+        char* ch = NULL;
+        uint32_t ch_length = 0;
+        origin_row_view.GetString(4, &ch, &ch_length);
+        int64_t origin_val = *reinterpret_cast<int64_t*>(ch);
+        if (origin_val != count) {
```
zhanghaohit: what's the checking here? shouldn't `ASSERT_EQ(origin_val, count);`?

src/storage/aggregator.cc
```diff
@@ -850,17 +866,40 @@ bool CountAggregator::DecodeAggrVal(const int8_t* row_ptr, AggrBuffer* buffer) {
...
+CountWhereAggregator::CountWhereAggregator(const ::openmldb::api::TableMeta& base_meta,
+                                           const ::openmldb::api::TableMeta& aggr_meta,
+                                           std::shared_ptr<Table> aggr_table,
+                                           std::shared_ptr<LogReplicator> aggr_replicator, const uint32_t& index_pos,
+                                           const std::string& aggr_col, const AggrType& aggr_type,
+                                           const std::string& ts_col, WindowType window_tpye, uint32_t window_size,
+                                           const std::string& filter_col)
+    : CountAggregator(base_meta, aggr_meta, aggr_table, aggr_replicator, index_pos, aggr_col, aggr_type, ts_col,
+                      window_tpye, window_size),
+      filter_col_(filter_col) {
```
zhanghaohit: check the `filter_col` is not empty

zhanghaohit (changes requested):

zhanghaohit (approved):
LGTM

dl239 (approved):

PR #1854, opened 2022-05-20
fix: python use db and split tests

### LINKED ISSUE #1852: python: use db bug
If create_engine without db, we will pass default db as "".
So code like this below can't pass.

```
create_engine without db
execute("use db")
execute("select * from t1;")
```

The bug is around `self.sdk.ExecuteSQL`, cause it will use the db in parameter list as the default db.

vagetablechicken:
closes #1852


aceforeverd:
[Bugs (0.5.x)](https://trello.com/c/WhbjlRN2/69-bugs-05x)

vagetablechicken:
add doc
1. url has db, strict
2. usl no db,

aceforeverd (approved):

tobegit3hub (approved):
LGTM

Issue #1200, opened 2022-02-11
python SDK `import openmldb` failed in ipython/notebook MacOS


HuilinWu2:
```
ImportError: dlopen(/Users/huilin/PycharmProjects/0211/venv/lib/python3.8/site-packages/openmldb/dbapi/../native/_sql_router_sdk.so, 2): initializer function 0x7fff67bc5357 not in mapped image for /Users/huilin/PycharmProjects/0211/venv/lib/python3.8/site-packages/openmldb/dbapi/../native/_sql_router_sdk.so
```

However, it works in Linux with python/ipython, Mac with python.

vagetablechicken:
I've got this error too.  Just in python3.9, it seems like this error does not only in ipython or notebook.

vagetablechicken:
One more error when running python on mac:

```
[libprotobuf FATAL /Users/runner/work/crossbow/crossbow/vcpkg/buildtrees/protobuf/src/23fa7edd52-3ba2225d30.clean/src/google/protobuf/stubs/common.cc:87] This program was compiled against version 3.6.1 of the Protocol Buffer runtime library, which is not compatible with the installed version (3.15.8).  Contact the program author for an update.  If you compiled the program yourself, make sure that your headers are from the same version of Protocol Buffers as your link-time library.  (Version verification failed in "/Users/ark/Code/OpenMLDB/src/proto/common.pb.cc".)
libc++abi.dylib: terminating with uncaught exception of type google::protobuf::FatalException: This program was compiled against version 3.6.1 of the Protocol Buffer runtime library, which is not compatible with the installed version (3.15.8).  Contact the program author for an update.  If you compiled the program yourself, make sure that your headers are from the same version of Protocol Buffers as your link-time library.  (Version verification failed in "/Users/ark/Code/OpenMLDB/src/proto/common.pb.cc".)
```

vagetablechicken:
still get initializer function error on openmldb python sdk 0.5.0, my machine is macos-10.15,

~~ld: warning: object file (/Users/ark/Code/OpenMLDB/.deps/usr/lib/libzetasql.a(unparser.o)) was built for newer macOS version (12.1) than being linked (10.15)~~
~~zetasql is built on 12.1.~~
~~I'll try to use 10.15 zetasql or build bundle in my own machine.~~

Not the root cause, forget it.

vagetablechicken:
10.15 zetasql still failed.
So I test released 050 python sdk again.
**Important**
[...]

PR #1858, opened 2022-05-20
fix: import in fixed order to avoid dlopen error

vagetablechicken:
Ref #1200 
Workaround, not be thoroughly solved.


tobegit3hub (approved):
LGTM

aceforeverd (approved):

PR #1870, opened 2022-05-23
fix: batch: compile warning that check a long for null #1408

### LINKED ISSUE #1408: batch: compile warning that check a long for null
```
[WARNING] /Users/tobe/code/4pd/OpenMLDB/java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/nodes/WindowAggPlan.scala:530: warning: com
paring values of types Long and Null using `==' will always yield false
[WARNING]     if (key == null) {
[WARNING]             ^
```

You can reproduce the issue by compiling java modules with `mvn clean package`.

Ivyee17:
close #1408


tobegit3hub:
There are some test failures. Please make sure you can run tests in local @Ivyee17 .

Ivyee17:
May be I will try it once the issue is assigned to me. I misunderstand that this issue has already assigned to me. 😂

dl239:
> May be I will try it once the issue is assigned to me. I misunderstand that this issue has already assigned to me. 😂

hi @Ivyee17 the issue have been assigned to you.

Ivyee17:
It seems there are some errors related to database in CI and maybe it's not related with this PR. If there's any problem related with this PR, feel free to contact me!

Logs:
W0525 16:24:44.412210  2162 ddl_parser.cc:401] hybrid engine compile sql failed, Fail to find column not_exist

Ivyee17:
It has been correct now. @dl239

Ivyee17:
If there is any question about this issue, feel free to contact me. @vagetablechicken @tobegit3hub

java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/nodes/WindowAggPlan.scala
```diff
@@ -730,7 +730,7 @@ object WindowAggPlan {
 
 
 
-  def isValidOrder(key: Long): Boolean = {
+  def isValidOrder(key: java.lang.Long): Boolean = {
```
vagetablechicken, line 733: `isValidOrder` always check the orderKey from `val orderKey = computer.extractKey(row)`.
But `extractKey` returns scala Long, you should make it return java Long too.

Ivyee17: > There are some test failures. Please make sure you can run tests in local @Ivyee17 .

I've already made changes in the previous PR, it will cause test failures, as @dl239 said before. @vagetablechicken

Ivyee17: The failure is about delete-extract-key, if changing it to java.lang.Lang too, then this test will fail. Searching a key which is deleted before will be true, not false. I wonder if there's mistakes in `extractKey` function. Besides, I do think it's far beyond what should be done in this PR. I just need to resolve the warning log, not the complete process, right? Another issue and pr should be taken instead of this issue!!

tobegit3hub (approved):
LGTM

vagetablechicken:
After check, the pr should only fix `isValidOrder` argument type. We want it to be java Long.
Actually we only pass scala Long to `isValidOrder`,  the mistake will be tracked by another issue #1903 .

vagetablechicken (approved):

PR #1872, opened 2022-05-24
fix: autoload when register offline table

### LINKED ISSUE #1565: taskmanager: registerOpenmldbOfflineTable should use options
https://github.com/4paradigm/OpenMLDB/blob/ba4c408c1b4dd9c488e31308b33044ccab988a07/java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/api/OpenmldbSession.scala#L269

Now we can't read offline path with options. e.g. when soft copy, csv file has header, but `spark.read.csv()` won't parse header. And we can't load without header, it's no schema(col name, type name).

vagetablechicken:
closes #1565


aceforeverd (approved):

tobegit3hub (approved):
LGTM

vagetablechicken:
[soft copy](https://trello.com/c/e24xTtov/68-soft-copy)

PR #1882, opened 2022-05-25
feat: optimize window computation with UnsafeRowOpt by removing zipped df

### LINKED ISSUE #1876: UnsafeRowOpt Spark planner for window to remove zipped dataframe
Now we need to zip the original DataFrame and internal DataFrame for window planner with UnsafeRowOpt. It would be great to remove the original DataFrame which may reduce the memory usage and improve the performance.

tobegit3hub:
* Resolve https://github.com/4paradigm/OpenMLDB/issues/1876
* Reduce the memory usage of Spark jobs with UnsafeRowOpt
* Reduce the time of Spark jobs with UnsafeRowOpt


java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/window/WindowComputer.scala
```diff
@@ -226,6 +238,11 @@ class WindowComputer(config: WindowAggConfig, jit: HybridSeJitWrapper, keepIndex
     SparkRowUtil.getLongFromIndex(config.orderIdx, orderField.dataType, curRow)
   }
 
+  def extractUnsafeKey(curRow: UnsafeRow): Long = {
```
vagetablechicken, line 241: returns scala Long here, so `null` from `unsafeGetLongFromIndex ` will return 0. Is it ok?

tobegit3hub: It is the same logic of old code. We may re-consider the logic of handling null data later.

The returning zero will be passed later and be treated as `invalid` timestamp.

aceforeverd (approved):

dl239 (approved):

tobegit3hub:
[performance optimization for offline processing](https://trello.com/c/IkCOLr7m/8-performance-optimization-for-offline-processing)

PR #1889, opened 2022-05-26
fix: resolve the compile warning of openmldb-batch

### LINKED ISSUE #1320: Resolve the compile warning of openmldb-batch
Now we can get logs of warning of Java module `openmldb-batch` when compiling.

```
[WARNING] /Users/tobe/code/4pd/OpenMLDB/java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/SparkPlanner.scala:342: warning: object Jav
aConversions in package collection is deprecated (since 2.12.0): use JavaConverters
[WARNING]       engine = new SqlEngine(sql, dbs, engineOptions, config.defaultDb)
[WARNING]                                   ^
```

You can reproduce the issue by compiling java modules with `mvn clean package`.

mangoGoForward:
Signed-off-by: mango <xu.weiKyrie@foxmail.com>

Bug fix

Resolved #1320

None

### DIFF
--- a/java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/SparkPlanner.scala
+++ b/java/openmldb-batch/src/main/scala/com/_4paradigm/openmldb/batch/SparkPlanner.scala
@@ -35,7 +35,7 @@ import org.apache.spark.SparkFiles
 import org.apache.spark.sql.{DataFrame, SparkSession}
 import org.slf4j.LoggerFactory
 
-import scala.collection.JavaConversions.seqAsJavaList
+import scala.collection.JavaConverters.seqAsJavaListConverter
 import scala.collection.mutable
 import scala.reflect.io.File
 


tobegit3hub:
Using `seqAsJavaListConverter` may cause the UT failures. Please check again and rerun the unit tests.

Ivyee17:
> Using `seqAsJavaListConverter` may cause the UT failures. Please check again and rerun the unit tests.

But I didn't use this converter. I just made a revision: to make `Long` type into `java.lang.Long` type in order to make sure the null can happen in this function. In fact, if using the `Long` type as the signature of function, there's no way to have `null` value at all! 
So, would you mind if I just delete the TODO for null value problem. It can prevent the warning to compare with `Long` and `Null`, which is always to be false. In Scala, if we take `null` as input with type `Long`, it'll throw an error instead of a warning. If the original code can be run successfully, then it cannot be `null` in fact.

mangoGoForward:
I got some errors when execute the command of `mvn clean package`, but I can't to solve, I am not good at scala, so I will close this pr, and if other are interested this related issue, please to contribute to solve.

tobegit3hub:
Anyway, thanks for your contribution @mangoGoForward 😄

PR #1892, opened 2022-05-26
fix: core dump in macos running sql_cluster_test

aceforeverd:
1. fix a `sql_cluster_test` core in darwin: when there is no partition data for a table locally, the `DistributeWindowIterator::SeekToFirst` set `it_` still. This cores the program when user want to `DistributeWindowIterator::GetKey()` because `it_` is not null, so it try to get key data from `it_`
2. enable more last join and window union test for cluster/request mode
3. update codecov.yml
4. improve `toydb_run_engine` bootstrapping

### DIFF
--- a/src/catalog/distribute_iterator.cc
+++ b/src/catalog/distribute_iterator.cc
@@ -179,47 +179,31 @@ void DistributeWindowIterator::Reset() {
     cur_pid_ = INVALID_PID;
 }
 
+// seek to the pos where key = `key` on success
+// or do nothing on fail (preserve old stat)
 void DistributeWindowIterator::Seek(const std::string& key) {
     DLOG(INFO) << "seek to key " << key;
-    Reset();
-    if (!tables_) {
+    const auto& stat = SeekByKey(key);
+    if (stat.pid < 0) {
+        DLOG(INFO) << "no pos found for key " << key;
         return;
     }
-    if (pid_num_ > 0) {
-        cur_pid_ = (uint32_t)(::openmldb::base::hash64(key) % pid_num_);
+
+    if (stat.it != nullptr) {
+        Reset();
+        it_.reset(stat.it);
+    } else if (stat.kv_it != nullptr) {
+        Reset();
+        response_vec_.push_back(stat.kv_it->GetResponse());
+        kv_it_ = stat.kv_it;
+    } else {
+        DLOG(INFO) << "no pos found for key " << key;
+        return;
     }
-    auto iter = tables_->find(cur_pid_);
-    if (iter != tables_->end()) {
-        it_.reset(iter->second->NewWindowIterator(index_));
-        it_->Seek(key);
-        if (it_->Valid()) {
-            return;
-        }
-    }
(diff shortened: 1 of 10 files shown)


zhanghaohit (approved):
LGTM

vagetablechicken (approved):

src/catalog/distribute_iterator.cc
```diff
@@ -227,22 +227,30 @@ void DistributeWindowIterator::SeekToFirst() {
...
+            delete it;
         }
     }
     for (const auto& kv : tablet_clients_) {
         uint32_t count = 0;
-        cur_pid_ = kv.first;
-        kv_it_ = kv.second->Traverse(tid_, cur_pid_, index_name_, "", 0, FLAGS_traverse_cnt_limit, count);
-        if (kv_it_ && kv_it_->Valid()) {
-            response_vec_.emplace_back(kv_it_->GetResponse());
+        auto it = kv.second->Traverse(tid_, kv.first, index_name_, "", 0, FLAGS_traverse_cnt_limit, count);
```
dl239, line 228: the `Seek` has the same problem. it's better to check it_ validation in `GetKey` and `GetValue` function

aceforeverd: we should ensure only zero or one of `it_` and `kv_it_` is not null.
So the check here is whether `it_` and `kv_it_` both not null.
It is `it_` and `kv_it_` 's duty to check valid itself before `GetKey` or `GetValue`

aceforeverd: we should improve type safety for iterators, in both definition and implementation. E.g the `GetKey` method of `WindowIterator` should mark as `const`

dl239 (approved):

src/catalog/distribute_iterator.cc
```diff
@@ -179,70 +179,108 @@ void DistributeWindowIterator::Reset() {
...
+        if (it != nullptr) {
+            it->Seek(key);
+            if (it->Valid()) {
+                return {static_cast<int32_t>(pid), it, {}};
+            }
+            delete it;
+        }
+    }
+    DLOG(INFO) << "seek to key " << key << " from remote. "
+               << " cur_pid " << cur_pid_;
```
aceforeverd: wrong log

src/catalog/distribute_iterator.cc
```diff
@@ -179,70 +179,108 @@ void DistributeWindowIterator::Reset() {
...
+        std::string msg;
+        auto it = client_iter->second->Scan(tid_, pid, key, index_name_, 0, 0, FLAGS_traverse_cnt_limit, msg);
+        if (it != nullptr && it->Valid()) {
+            return {static_cast<int32_t>(pid), {}, it};
+        }
+    }
+
+    // fallback
+    for (const auto& kv : *tables_) {
+        if (kv.first <= cur_pid_) {
```
aceforeverd: wrong compare

aceforeverd:
[window union across tables](https://trello.com/c/9jpyUwAv/67-window-union-across-tables)

PR #1896, opened 2022-05-27
feat: built-in-function: support char(int) function #817

### LINKED ISSUE #817: built-in-function: Support `char(int)` function

char(expr) 
Returns the ASCII character having the binary equivalent to expr. If n is larger than 256 the result is equivalent to chr(n % 256)

Examples:
```sql
> SELECT char(65);
 A
```

### Reference:
https://spark.apache.org/docs/2.3.0/api/sql/#char

We do provide a built-in function development guide for you. 
Please check [OpenMLDB Build-In Function Develop Guide](https://github.com/4paradigm/OpenMLDB/blob/main/docs/en/built_in_function_develop_guide.md) or the [Chinese version](https://github.com/4paradigm/OpenMLDB/blob/main/docs/cn/built_in_function_develop_guide.md) for help.

Please make sure you:
- [ ] Add built-in C++ function in src/udf/udf.h and src/udf/udf.cc
- [ ] Register function to default library from function `void DefaultUdfLibrary::IniStringUdf()` in src/udf/default_udf_library.cc.
- [ ] Add related unit test in src/codegen/udf_ir_builder_test.cc
- [ ] Documenting function 

**Additional context**

Ivyee17:
Close #817


hybridse/src/codegen/udf_ir_builder_test.cc
```diff
@@ -1007,6 +1007,12 @@ TEST_F(UdfIRBuilderTest, degrees) {
     CheckUdf<double, double>(udf_name, -90.0, -pi/2);
     CheckUdf<Nullable<double>, Nullable<double>>(udf_name, nullptr, nullptr);
 }
+TEST_F(UdfIRBuilderTest, char_test) {
+    auto udf_name = "char";
+    CheckUdf<StringRef,int32_t>(udf_name, StringRef("A"), 65);
```
aceforeverd: style issue

hybridse/src/udf/default_udf_library.cc
```diff
@@ -895,6 +895,20 @@ void DefaultUdfLibrary::InitStringUdf() {
             @since 0.4.0)");
     RegisterAlias("lower", "lcase");
     RegisterAlias("upper", "ucase");
+        RegisterExternal("char")
```
aceforeverd: styling, not align with above line

hybridse/src/udf/default_udf_library.cc
```diff
@@ -895,6 +895,20 @@ void DefaultUdfLibrary::InitStringUdf() {
...
+        .doc(R"(
+            @brief Returns the ASCII character having the binary equivalent to expr. If n is larger than 256 the result is equivalent to chr(n % 256).
+
+            Example:
+
+            @code{.sql}
+                SELECT char(65);
+                --output "A"
+            @endcode
+            @since 0.5.0)");
```
aceforeverd: since 0.6.0

hybridse/src/udf/default_udf_library.cc
```diff
@@ -895,6 +895,20 @@ void DefaultUdfLibrary::InitStringUdf() {
             @since 0.4.0)");
     RegisterAlias("lower", "lcase");
     RegisterAlias("upper", "ucase");
+        RegisterExternal("char")
+        .args<int32_t>(
+            static_cast<void (*)(int32_t, StringRef*)>(udf::v1::int_to_char))
+        .return_by_arg(true)
+        .doc(R"(
+            @brief Returns the ASCII character having the binary equivalent to expr. If n is larger than 256 the result is equivalent to chr(n % 256).
```
aceforeverd: char(n % 256)

hybridse/src/codegen/udf_ir_builder_test.cc
```diff
@@ -1007,6 +1007,12 @@ TEST_F(UdfIRBuilderTest, degrees) {
...
+    CheckUdf<StringRef,int32_t>(udf_name, StringRef("B"), 322);
```
aceforeverd: check when val is null

hybridse/src/udf/udf.cc
```diff
@@ -157,6 +157,17 @@ int32_t weekofyear(Date *date) {
     }
 }
 
+void int_to_char(int32_t val, StringRef* output){
+    if (val > 256)
```
aceforeverd: what if val < 0 ?

hybridse/src/udf/udf.cc
```diff
@@ -157,6 +157,17 @@ int32_t weekofyear(Date *date) {
     }
 }
 
+void int_to_char(int32_t val, StringRef* output){
+    if (val > 256) 
+        val = val % 256;
+    char *buffer = AllocManagedStringBuf(1);
+    output->size_ = 1;
+    char v=(char)(val);
+    std::string words(&v);
```
aceforeverd: not necessary create another string

aceforeverd (changes requested):

Ivyee17:
Seems like it'll never ever finished checking.

hybridse/src/udf/udf.cc
Ivyee17: so what should I revise?
I just removed the buffer, and hope to match the advice you given (which is not creating another string).

hybridse/src/udf/udf.cc
Ivyee17: The api just doesn’t mention about Val<0. If you define the operation it should be, then I’ll glad to have more tests.

hybridse/src/codegen/udf_ir_builder_test.cc
Ivyee17: ok

Ivyee17:
cicd/cpp CI is down. maybe need to restart? @aceforeverd

hybridse/src/udf/udf.cc
aceforeverd: As a general guide, you can reference the result of `char(-xxx)` from spark-sql

hybridse/src/codegen/udf_ir_builder_test.cc
```diff
@@ -1007,6 +1007,12 @@ TEST_F(UdfIRBuilderTest, degrees) {
     CheckUdf<double, double>(udf_name, -90.0, -pi/2);
     CheckUdf<Nullable<double>, Nullable<double>>(udf_name, nullptr, nullptr);
 }
+TEST_F(UdfIRBuilderTest, char_test) {
```
aceforeverd: use CamelCase for test case name

hybridse/src/udf/udf.cc
```diff
@@ -157,6 +157,16 @@ int32_t weekofyear(Date *date) {
...
 }
 
+void int_to_char(int32_t val, StringRef* output) {
+    if (val > 256)
+        val = val % 256;
+    output->size_ = 1;
+    char v = static_cast<char>(val);
+    std::string s;
+    s = c;
+    output->data_ = s.c_str();
```
aceforeverd: not correct, you should write data directly to the `AllocManagedStringBuf` returned buf.

aceforeverd (changes requested):

hybridse/src/udf/udf.cc
Ivyee17: Are you sure? Maybe you can talk with @aceforeverd , he says that I'm not allowed to use another string!!!
image

Ivyee17: oh sorry, that's yourself. then what do you mean? could you give me an example? I'm not the staff of openMLDB, and I don't think I can understand what you say as accurately as your staff, especially you assign this issue as beginner's problem AFTER I take this PR, with so many problems (such as just one blank space, which needs nearly 2 hours to run this entire CI to ensure its correctness). @aceforeverd

hybridse/src/codegen/udf_ir_builder_test.cc
Ivyee17: ok.

hybridse/src/udf/udf.cc
Ivyee17: ok. I will leave it as the same as the number which is bigger than 256. Hope you can satisfy.

hybridse/src/udf/udf.cc
aceforeverd: I mean you can write directly into the char* buf return from AllocManagedStringBuf. The string here will be freed after function exit, this make the returned StringRef invalid.

You can see https://github.com/4paradigm/OpenMLDB/blob/19cbda374f46d836a363b82ba25b9f50c85b0364/hybridse/src/udf/udf.cc#L269 for how to use AllocManagedStringBuf.

Again, pls be nice and patient

Ivyee17: ok

aceforeverd (approved):

aceforeverd:
Pls resolve merge conflicts

hybridse/src/udf/udf.cc
```diff
@@ -157,6 +157,16 @@ int32_t weekofyear(Date *date) {
     }
 }
 
+void int_to_char(int32_t val, StringRef* output) {
+    if (val > 256 || val < -256)
+        val = val % 256;
+    output->size_ = 1;
+    char v = static_cast<char>(val);
+    char *buffer = AllocManagedStringBuf(1);
+    output->size_ = 1;
```
zhanghaohit, line 164: duplicate assignment `output->size_ = 1`

hybridse/src/codegen/udf_ir_builder_test.cc
```diff
@@ -1007,6 +1007,13 @@ TEST_F(UdfIRBuilderTest, degrees) {
     CheckUdf<double, double>(udf_name, -90.0, -pi/2);
     CheckUdf<Nullable<double>, Nullable<double>>(udf_name, nullptr, nullptr);
 }
+TEST_F(UdfIRBuilderTest, charTest) {
+    auto udf_name = "char";
+    CheckUdf<StringRef, int32_t>(udf_name, StringRef("A"), 65);
+    CheckUdf<StringRef, int32_t>(udf_name, StringRef("B"), 322);
+    CheckUdf<StringRef, int32_t>(udf_name, StringRef("N"), -178);
```
zhanghaohit, line 1014: add a check for `256` and `-256`

hybridse/src/udf/default_udf_library.cc
```diff
@@ -895,6 +895,20 @@ void DefaultUdfLibrary::InitStringUdf() {
             @since 0.4.0)");
     RegisterAlias("lower", "lcase");
     RegisterAlias("upper", "ucase");
+    RegisterExternal("char")
+        .args<int32_t>(
+            static_cast<void (*)(int32_t, StringRef*)>(udf::v1::int_to_char))
+        .return_by_arg(true)
+        .doc(R"(
+            @brief Returns the ASCII character having the binary equivalent to expr. If n is larger than 256 the result is equivalent to char(n % 256).
```
zhanghaohit: if n == 256, what happens?

`If n is larger than 256 the result is equivalent to char(n % 256)` should be `If n >= 256 the result is equivalent to char(n % 256)`?

hybridse/src/udf/udf.cc
```diff
@@ -157,6 +157,16 @@ int32_t weekofyear(Date *date) {
     }
 }
 
+void int_to_char(int32_t val, StringRef* output) {
+    if (val > 256 || val < -256)
```
zhanghaohit: what's the defined behavior if val < 0? Is there a clear definition for val < 0? @aceforeverd

zhanghaohit (changes requested):

hybridse/src/udf/udf.cc
aceforeverd: no clear definition. I think it should always be `val % 256`. so `-1` is 255
This if check should be removed

hybridse/src/udf/udf.cc
Ivyee17: ok

hybridse/src/udf/udf.cc
Ivyee17: Cool idea! I'll remove this check.

hybridse/src/udf/default_udf_library.cc
Ivyee17: ok

hybridse/src/codegen/udf_ir_builder_test.cc
Ivyee17: ok

zhanghaohit (approved):
LGTM

Ivyee17:
cicd/cpp is down because the DBSDK is failed to init (for connection problem or sth. else?). may need to restart this CI? 

Log:
E0530 09:35:52.121150 13344 name_server_impl.cc:10790] ERROR: Failed to init DBSDK

hybridse/src/udf/udf.cc
```diff
@@ -157,6 +157,20 @@ int32_t weekofyear(Date *date) {
     }
 }
 
+void int_to_char(int32_t val, StringRef* output) {
+    val = val % 256;
+    int32_t strl = 1;
+    if (val == 0) {
+        strl = 0;
```
aceforeverd: I thought this impl is not accurate.
* the output str should always have constant length 1
* for the case `val = 0`, the output string is `\0` with size =1 .  Be aware the difference between C string and this. Here `\0` is not the delimiter.

hybridse/src/codegen/udf_ir_builder_test.cc
```diff
@@ -1007,6 +1007,15 @@ TEST_F(UdfIRBuilderTest, degrees) {
...
+    CheckUdf<StringRef, int32_t>(udf_name, StringRef("\0"), 256);
```
aceforeverd: use `StringRef(1, "\0")`

aceforeverd (changes requested):

aceforeverd:
> cicd/cpp is down because the DBSDK is failed to init (for connection problem or sth. else?). may need to restart this CI?
> 
> Log: E0530 09:35:52.121150 13344 name_server_impl.cc:10790] ERROR: Failed to init DBSDK

This was because of machine resource limitation, no worry

hybridse/src/udf/udf.cc
Ivyee17: I agree. The result of the expected value is "", and I match the case with "\0", which is not correct. Maybe I'm not quite familiar with the StringRef constructor. In this case, I should revise the expected value to "\0" instead of revising the return value to "". Very appreciated to your idea!

hybridse/src/codegen/udf_ir_builder_test.cc
Ivyee17: This is the constructor I want, thank you!

aceforeverd (approved):

PR #1901, opened 2022-05-30
feat: support iterating key from remote tablet with `Next` function in `DistributeWindowIterator`

### LINKED ISSUE #1887: Support getting remote data with `Next` function in `DistributeWindowIterator`
The `Next` function in `DistributeWindowIterator` can only iterator data in memory table currently.

dl239:


src/catalog/distribute_iterator.cc
```diff
@@ -182,18 +186,16 @@ void DistributeWindowIterator::Reset() {
 // seek to the pos where key = `key` on success
 // or do nothing on fail (preserve old stat)
 void DistributeWindowIterator::Seek(const std::string& key) {
+    Reset();
```
aceforeverd, line 189: seek failed should do nothing ?

src/catalog/distribute_iterator.h
```diff
@@ -130,20 +130,21 @@ class DistributeWindowIterator : public ::hybridse::codec::WindowIterator {
...
-        ItStat(int32_t p, ::hybridse::codec::WindowIterator* i, KV_IT kv) : pid(p), it(i), kv_it(kv) {}
+        ItStat(uint32_t id, ::hybridse::codec::WindowIterator* i, KV_IT kv) : pid(id), it(i), kv_it(kv) {}
     };
 
  private:
     void Reset();
 
     ItStat SeekByKey(const std::string& key) const;
 
+    ItStat SeekToFirstRemote();
```
aceforeverd: const

src/catalog/distribute_iterator.cc
```diff
@@ -284,11 +293,57 @@ void DistributeWindowIterator::Next() {
                 it_->SeekToFirst();
```
aceforeverd, line 293: `it_` should be only updated on success, and do nothing changed on fail.
In case if all `tablets_` left is empty, then try to seek from remote, the `it_` do not to be `nullptr`

src/catalog/distribute_iterator.cc
dl239: the comment has been updated

src/catalog/distribute_iterator.cc
dl239: will check `it_` is `nullptr` and validation in getkey/getvalue function

aceforeverd (approved):

vagetablechicken (approved):

PR #1938, opened 2022-06-09
feat: support zk log options & flags fix

### LINKED ISSUE #1500: sql_client: zk "exceeded deadline by N ms" floods logs
Ref https://issues.apache.org/jira/browse/ZOOKEEPER-642
We should support two feature:
1. set zk_session_timeout in sql_client(only sql_client haven't supported zk_session_timeout)
2. set log level of zk client in sql_client(servers don't need to remove the zk logs, only interactive client needs)

optional:
update zk c api version to 3.5.0, it'll reduce the amount of meaningless log messages.

vagetablechicken:
Closes #1500 
CLI: disable zk log by default(log level 0). use gflags zk_log_level & zk_log_file to modify `ClusterOptions`.
other sdk: use default `ClusterOptions ` now, info level and use stream stderr(default in zk).

make more flag desc.

### DIFF
--- a/src/zk/zk_client.h
+++ b/src/zk/zk_client.h
@@ -19,6 +19,7 @@
 
 #include <atomic>
 #include <condition_variable>  // NOLINT
+#include <cstdio>
 #include <map>
 #include <mutex>  // NOLINT
 #include <string>
@@ -27,7 +28,6 @@
 #include "boost/function.hpp"
 
 extern "C" {
-#include "stdio.h"
 #include "zookeeper/zookeeper.h"
 }
 
--- a/src/zk/zk_client.h
+++ b/src/zk/zk_client.h
@@ -27,6 +27,7 @@
 #include "boost/function.hpp"
 
 extern "C" {
+#include "stdio.h"
 #include "zookeeper/zookeeper.h"
 }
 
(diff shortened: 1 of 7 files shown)


tobegit3hub (approved):
LGTM

src/zk/zk_client.h
```diff
@@ -27,6 +27,7 @@
 #include "boost/function.hpp"
 
 extern "C" {
+#include "stdio.h"
```
dl239: please resolve all `cpplint` warnings

src/zk/zk_client.h
```diff
@@ -52,7 +53,11 @@ class ZkClient {
     ~ZkClient();
 
     // init zookeeper connections
-    bool Init();
+    // log level: DISABLE_LOGGING=0, ZOO_LOG_LEVEL_ERROR=1, ZOO_LOG_LEVEL_WARN=2, ZOO_LOG_LEVEL_INFO=3,
```
dl239, line 56: add enum

vagetablechicken: It's zk enum, integer is ok, no need to add more

dl239: it's not recommend to pass magic number when  call `init` function,

dl239: L44 in src/sdk/db_sdk.h and default parameter in init

dl239 (approved):

PR #1950, opened 2022-06-11
fix(#1787): wrong result from window open preceding query

### LINKED ISSUE #1787: window `OPEN PRECEDING` in frame start border do not take effect
**Bug Description**

```yaml
  - id: 19
    inputs:
      - name: t1
        columns:
          - id int
          - ts timestamp
          - g int
          - val int
        indexs:
          - idx:id:ts
        data: |
          1,  99, 111, 21
          2, 100, 111, 22
          3, 101, 111, 23
    sql: |
       select
          id, count(val) over w as cnt,
          max(val) over w as mv,
          min(val) over w as mi,
          lag(val, 1) over w as l1
       from t1 window w as(
          partition by `g` order by `ts`
          ROWS between 2 preceding and 0 OPEN PRECEDING);
    expect:
      columns:
        - id int
        - cnt int64
        - mv int
        - mi int
        - l1 int
      order: id
      data: |
         1, 0, NULL, NULL, NULL
         2, 1, 21, 21, 21
         3, 2, 22, 21, 22
```

the case do not pass because it still treat current row inside window when doing window aggregation. 
`OPEN PRECEDING` in end frame border seems work as expected though.

**Expected Behavior**

**Steps to Reproduce**

1. 
2. 
3. 
4.

aceforeverd:

close #1787 


aceforeverd:
[High-priority Bugs (0.5.x)](https://trello.com/c/iGid67sw/70-high-priority-bugs-05x)

hybridse/include/node/sql_node.h
```diff
@@ -1115,48 +1116,45 @@ class FrameBound : public SqlNode {
...
+    /// 2s following and 1s following -> (true, false)
+    int64_t GetSignedOffset(bool is_start_frame) const {
         switch (bound_type_) {
             case node::kCurrent:
                 return 0;
             case node::kFollowing:
                 return offset_;
             case node::kOpenFollowing:
-                return offset_ - 1;
+                return is_start_frame ? offset_ - 1 : offset_ + 1;
```
zhanghaohit: what if:
```
2 OPEN PRECEDING and 2 OPEN FOLLOWING
```
start frame is: `2 OPEN PRECEDING`
end frame is: `2 OPEN FOLLOWING`
correct range should be: [-1, 1]?

but here will generate [-1, 3]？

hybridse/include/node/sql_node.h
```diff
@@ -1115,48 +1116,45 @@ class FrameBound : public SqlNode {
...
+    /// usually, `offset_` is non-negative (think about the offset expr in SQL window frame:
+    ///         'ROWS BETWEEN 2 PRECEDING AND 0 OPEN PRECEDING')
+    /// * for 'PRECEDING' bound, output is negative
+    /// * for 'FOLLOWING' bound, output is positive
+    /// * for 'OPEN [ PRECEDING | FOLLOWING ]', output is one increasing or decreasing, based on
+    ///   whether it is start bound or end bound
+    ///
+    /// \param is_start_frame whether this bound is start or end of window frame
+    ///
+    /// by convention, the frame that is more close to the current row is the end frame
```
zhanghaohit: is it the earliest frame is start frame? e.g.,
`2 FOLLOWING and 5 FOLLOWING`
`2 FOLLOWING` is the start frame and `5 FOLLOWING` is the end frame?

zhanghaohit (changes requested):

hybridse/include/node/sql_node.h
aceforeverd: Should be, the lower offset is the start frame. And strictly, the `SignedOffset` of start frame should less or equal to end frame. see #1951

hybridse/include/node/sql_node.h
aceforeverd: you are right, I fixed the start frame definition

zhanghaohit (approved):
LGTM

dl239 (approved):

PR #1984, opened 2022-06-17
refactor: removed `Count` function with `ts_name` argument from src/tablet_client…

### LINKED ISSUE #1974: Remove `Count` function from tablet client
`Count` function with `ts_name` is unused now. We should delete it from src/client/tablet_client.h and src/client/tablet_client.cc
- delete declaration in [header file](https://github.com/4paradigm/OpenMLDB/blob/main/src/client/tablet_client.h#L101)
- delete the implement in [tablet_client.cc](https://github.com/4paradigm/OpenMLDB/blob/main/src/client/tablet_client.cc#L782)
- do not pass `ts_name` when calling `Count` function in /src/cmd/openmldb.cc

TAldridgeAO:
….h and src/tablet_client.cc. Removed ts_name parameter from count call in openmldb.cc

This fixes Issue "Remove Count function from tablet client #1974"
close #1974 

No Behavior changes.

No longer calls Count using ts_name in openmldb.cc


src/client/tablet_client.cc
```diff
@@ -779,28 +779,6 @@ bool TabletClient::Count(uint32_t tid, uint32_t pid, const std::string& pk, cons
     return Count(tid, pid, pk, idx_name, "", filter_expired_data, value, msg);
 }
 
-bool TabletClient::Count(uint32_t tid, uint32_t pid, const std::string& pk, const std::string& idx_name,
```
dl239, line 782: another `Count` function above will call this function. you can move this implement above.

TAldridgeAO: Can you please clarify what you are wanting me to do with this? I can't find another function calling Count within this section.

TAldridgeAO: Ignore that previous. I was looking at this pre-coffee. To clarify, it looks like the count function on line 777 should instead use the implementation from the removed Count function, correct?

dl239: yes,

TAldridgeAO: Okay. I've just pushed a new commit, remove the call to Count from Count, and adding back in the implementation. Let me know if anything else is needed!

dl239: Well done.  There is compiling errors https://github.com/4paradigm/OpenMLDB/runs/6959539945?check_suite_focus=true#step:7:587
It also need to remove `ts_name` from here https://github.com/4paradigm/OpenMLDB/blob/40d3cbc98eb4a5750078f4849a8f097f6c89a1a1/src/cmd/openmldb.cc#L3587. thank you

dl239 (approved):

src/client/tablet_client.cc
```diff
@@ -776,12 +776,7 @@ bool TabletClient::Get(uint32_t tid, uint32_t pid, const std::string& pk, uint64
 
 bool TabletClient::Count(uint32_t tid, uint32_t pid, const std::string& pk, const std::string& idx_name,
                          bool filter_expired_data, uint64_t& value, std::string& msg) {
-    return Count(tid, pid, pk, idx_name, "", filter_expired_data, value, msg);
-}
-
-bool TabletClient::Count(uint32_t tid, uint32_t pid, const std::string& pk, const std::string& idx_name,
-                         const std::string& ts_name, bool filter_expired_data, uint64_t& value, std::string& msg) {
-    ::openmldb::api::CountRequest request;
+     ::openmldb::api::CountRequest request;
```
vagetablechicken: intent, an extra space?

TAldridgeAO: Not sure how that happened, guessing it was just an issue with VSCODE not liking my paste. I've updated it and pushed a new commit to fix.

vagetablechicken (approved):

Issue #1586, opened 2022-04-06
default udf `sum/sum_where/avg/avg_where` should return NULL if no rows in range or all values is NULL


aceforeverd:
**Bug Description**

currently the `sum()` is defined with init value `0`, https://github.com/4paradigm/OpenMLDB/blob/f5fc988a38d0d9b8952155cdfe9d782133502e46/hybridse/src/udf/default_udf_library.cc#L99-L103, which should be `NULL` instead.

The same situation apply to `sum_where/avg/avg_where`. For `count/count_where`, it is correct to have `0` as init value 

a example test case:
```yaml
  - id: 15
    desc: sum correctness when no rows in range or value is all null
    inputs:
      - columns: [ "row_id int","ts timestamp","group1 string","val1 int" ]
        indexs: [ "index1:group1:ts" ]
        name: t1
        data: |
          1, 1612130400000, g1, NULL
          2, 1612130401000, g1, 2
          3, 1612130402000, g2, 3
          4, 1612130403000, g2, 4
          5, 1612130404000, g2, 5
    sql: |
      select
      `row_id`,
      `val1`,
      sum(val1) over w1 as agg1,
      from `t1` WINDOW
        w1 as (partition by `group1` order by `ts` rows_range between 5s preceding and 0s preceding MAXSIZE 10);
    expect:
      columns: ["row_id int", "val1 int", "agg1 int"]
      order: row_id
      data: |
        1, 1, NULL
        2, 2, 2
        3, 3, 3
        4, 4, 7
        5, 5, 12
```

**Expected Behavior**

**Steps to Reproduce**

1. 
2. 
3. 
4.

zhanghaohit:
pre-aggregation logics (in `src/storage/aggregation` and `hybridse/vm/aggregation`) should be updated too.

aceforeverd:
![](https://github.trello.services/images/mini-trello-icon.png) [High-priority Bugs (0.5.x)](https://trello.com/c/iGid67sw/70-high-priority-bugs-05x)

PR #1997, opened 2022-06-17
fix(codegen): loop combing should not reuse same count value for different columns

### LINKED ISSUE #1927: Get incorrect result for min/max/count with null data which may be related to loop-combining codegen
Now we may get incorrect result for methods of min/max/count with null data. Here is simple case which can be run in offline directly.

```
package com._4paradigm.openmldb.batchjob.tools.tobe.loaddata

import com._4paradigm.openmldb.batch.api.OpenmldbSession
import org.apache.spark.sql.{Row, SparkSession}
import org.apache.spark.sql.types.{IntegerType, StringType, StructField, StructType, TimestampType}
import java.sql.Timestamp

object SameWindowIssueLocal {

  def main(args: Array[String]): Unit = {

    val spark = SparkSession.builder().master("local")
      .getOrCreate()

    val sess = new OpenmldbSession(spark)

    val data = Seq(
      Row(1, "g1", Timestamp.valueOf("2021-02-01 13:59:58"), 1, -3),
      Row(2, "g1", Timestamp.valueOf("2021-02-01 13:59:59"), null, null),
      Row(3, "g1", Timestamp.valueOf("2021-02-01 14:00:00"), 2, -1),
      Row(4, "g2", Timestamp.valueOf("2021-02-01 14:00:00"), null, -1),
      Row(5, "g2", Timestamp.valueOf("2021-02-01 14:00:01"), null, -2),
      Row(6, "g3", Timestamp.valueOf("2021-02-01 13:54:02"), 1, 1),
    )
    val schema = StructType(List(
      StructField("row_id", IntegerType),
      StructField("key1", StringType),
      StructField("ts", TimestampType),
      StructField("l1", IntegerType),
      StructField("l2", IntegerType)
    ))
    val df = spark.createDataFrame(spark.sparkContext.makeRDD(data), schema)
    sess.registerTable("t1", df)

    val sqlText =
      """
        |select
        |    `row_id` as row_id_1,
        |    max(`l1`) over t6_2_key1_ts_0s_5s_10 as t6_2_l1_window_max_8,
        |    max(`l2`) over t6_2_key1_ts_0s_5s_10 as t6_2_l2_window_max_9
        |from
        |    `t1`
        |    window t6_2_key1_ts_0s_5s_10 as (partition by `key1` order by `ts` rows_range between 5s open preceding and 0s preceding MAXSIZE 10);
        |""".stripMargin

    val outputDf = sess.sql(sqlText)
    outputDf.show()

    val sqlText2 =
      """
        |select
        |    `row_id` as row_id_1,
        |    max(`l2`) over t6_2_key1_ts_0s_5s_10 as t6_2_l2_window_max_9
        |from
        |    `t1`
        |    window t6_2_key1_ts_0s_5s_10 as (partition by `key1` order by `ts` rows_range between 5s open preceding and 0s preceding MAXSIZE 10);
        |""".stripMargin

    val outputDf2 = sess.sql(sqlText2)
    outputDf2.show()
  }

}
``` 

If we only select the non-null column, the result is correct.

```
+--------+--------------------+
|row_id_1|t6_2_l2_window_max_9|
+--------+--------------------+
|       4|                  -1|
|       5|                  -1|
|       6|                   1|
|       1|                  -3|
|       2|                  -3|
|       3|                  -1|
+--------+--------------------+
```

If we select two columns and one of these columns has null data, the result is incorrect and make all result is null.

```
+--------+--------------------+--------------------+
|row_id_1|t6_2_l1_window_max_8|t6_2_l2_window_max_9|
+--------+--------------------+--------------------+
|       4|                null|                null|
|       5|                null|                null|
|       6|                   1|                   1|
|       1|                   1|                  -3|
|       2|                   1|                  -3|
|       3|                   2|                  -1|
+--------+--------------------+--------------------+
```

If we change the method from `max` to `add` which may not use loop-combining CodeGen, the result will be correct. So it may be the issue of handling null data in loop-combining CodeGen.

### LINKED ISSUE #1951: `window rows between 0 preceding and 2 preceding` should report compile error
**Bug Description**

the behavior of sql parser should be the same with spark-sql.
in spark, query like this should result compile time error
```sql
spark-sql> select x, sum(y) over (partition by z order by x rows between 0 preceding and 2 preceding) from values ((1, 1, 2)), ((2, 2, 2)) as tab(x, y, z);
Error in query: cannot resolve 'ROWS BETWEEN 0 PRECEDING AND 2 PRECEDING' due to data type mismatch: The lower bound of a window frame must be less than or equal to the upper bound; line 1 pos 22;
'Project [x#12, unresolvedalias(sum(y#13) windowspecdefinition(z#14, x#12 ASC NULLS FIRST, specifiedwindowframe(RowFrame, -0, -2)), None)]
+- SubqueryAlias tab
   +- LocalRelation [x#12, y#13, z#14]

spark-sql> select x, sum(y) over (partition by z order by x rows between 2 following and 0 following) from values ((1, 1, 2)), ((2, 2, 2)) as tab(x, y, z);
Error in query: cannot resolve 'ROWS BETWEEN 2 FOLLOWING AND 0 FOLLOWING' due to data type mismatch: The lower bound of a window frame must be less than or equal to the upper bound; line 1 pos 22;
'Project [x#20, unresolvedalias(sum(y#21) windowspecdefinition(z#22, x#20 ASC NULLS FIRST, specifiedwindowframe(RowFrame, 2, 0)), None)]
+- SubqueryAlias tab
   +- LocalRelation [x#20, y#21, z#22]
```
spark always expect `start_frame <= end_frame`

currently we just treat those window as window without data

**Expected Behavior**

**Steps to Reproduce**

1. 
2. 
3. 
4.

aceforeverd:

- fix: #1927 
- fix: #1951 
- partial #1586 


aceforeverd:
[bugfix: sql engine related bugs](https://trello.com/c/Gu5DKwue/104-bugfix-sql-engine-related-bugs)

tobegit3hub (approved):
LGTM

hybridse/src/codegen/aggregate_ir_builder.cc
```diff
@@ -236,34 +232,25 @@ class StatisticalAggGenerator {
...
     void GenInitState(::llvm::IRBuilder<>* builder) {
         for (size_t i = 0; i < col_num_; ++i) {
+            // any of min/max/avg/sum/count need the count info
+            count_state_[i] = GenCountInitState(builder);
+
             if (!sum_idxs_[i].empty()) {
                 sum_states_[i] = GenSumInitState(builder);
             }
             if (!avg_idxs_[i].empty()) {
                 if (col_type_ == ::hybridse::node::kDouble) {
```
dl239, line 242: why only judge `kDouble` here.

aceforeverd: the idea is generally:
- if column type is double, `avg` and `sum` can share the sum value

on output, avg requires `sum` and `count` to be float

zhanghaohit (approved):
LGTM

dl239 (approved):

PR #2012, opened 2022-06-21
fix: move in return stmt preventing copy elision

### LINKED ISSUE #2004: Return unique_ptr directly in `GetIterator`
Moving a local object in a return statement prevents copy elision.  It's better to return unique_ptr directly as below:
`return std::make_unique<ConstIterator<uint64_t, V>>(root_, this);`

https://github.com/4paradigm/OpenMLDB/blob/main/hybridse/include/codec/list_iterator_codec.h#L99

CuriousCorrelation:
* **What kind of change does this PR introduce?**
Fixes #2004
Follows suggestion from the issue desc.

* **What is the current behavior?** 
Move prevents copy elision
https://pvs-studio.com/en/docs/warnings/v828/
https://en.cppreference.com/w/cpp/language/copy_elision

* **What is the new behavior?**
Copy Elision is guaranteed when an object is returned directly (>C++17)

### DIFF
--- a/hybridse/include/codec/list_iterator_codec.h
+++ b/hybridse/include/codec/list_iterator_codec.h
@@ -94,9 +94,7 @@ class ColumnImpl : public WrapListImpl<V, Row> {
 
     // TODO(xxx): iterator of nullable V
     std::unique_ptr<ConstIterator<uint64_t, V>> GetIterator() override {
-        auto iter = std::unique_ptr<ConstIterator<uint64_t, V>>(
-            new ColumnIterator<V>(root_, this));
-        return std::move(iter);
+        return std::make_unique<ColumnIterator<V>>(root_, this);
     }
     ConstIterator<uint64_t, V> *GetRawIterator() override {
         return new ColumnIterator<V>(root_, this);


hybridse/include/codec/list_iterator_codec.h
```diff
@@ -94,9 +94,7 @@ class ColumnImpl : public WrapListImpl<V, Row> {
 
     // TODO(xxx): iterator of nullable V
     std::unique_ptr<ConstIterator<uint64_t, V>> GetIterator() override {
-        auto iter = std::unique_ptr<ConstIterator<uint64_t, V>>(
-            new ColumnIterator<V>(root_, this));
-        return std::move(iter);
+        return std::make_unique<ConstIterator<uint64_t, V>>(root_, this);
```
aceforeverd: guess that should be `make_unique<ColumnIterator>`

CuriousCorrelation: This `std::make_unique<ColumnIterator<V>>(root_, this);` is what I think the return statement should be, although I wasn't sure if I were to test the code mentioned in the issue or if I were to correct it. I'll try to force push that, let me know if there's anything I missed/should change.

dl239 (approved):

aceforeverd (approved):

hybridse/include/codec/list_iterator_codec.h
aceforeverd: you are correct, `ColumnIterator` is sub-class of `Constiterator`

PR #2024, opened 2022-06-22
fix(#1586): default value for `sum/avg/sum_where/avg_where`

### LINKED ISSUE #1586: default udf `sum/sum_where/avg/avg_where` should return NULL if no rows in range or all values is NULL

aceforeverd:
close #1586

### DIFF
--- a/cases/function/function/test_udaf_function.yaml
+++ b/cases/function/function/test_udaf_function.yaml
@@ -386,22 +386,28 @@ cases:
       -
         columns : ["id int","c1 string","c2 smallint","c3 int","c4 bigint","c5 float","c6 double","c7 timestamp","c8 date","c9 string","c10 bool"]
         indexs: ["index1:c1:c7"]
-        rows:
-          - [1,"aa",1,1,30,1.1,2.1,1590738990000,"2020-05-01","a",true]
-          - [2,"aa",4,4,33,1.4,2.4,1590738991000,"2020-05-03","c",false]
-          - [3,"aa",3,3,32,1.3,2.3,1590738992000,"2020-05-02","b",true]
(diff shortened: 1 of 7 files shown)


aceforeverd:
[bugfix: sql engine related bugs](https://trello.com/c/Gu5DKwue/104-bugfix-sql-engine-related-bugs)

dl239 (approved):

zhanghaohit (approved):

PR #2043, opened 2022-06-29
fix: use int to record string length instead of map

### LINKED ISSUE #1979: Use int value to record string length instead of Map in `InsertPreparedStatementImpl `
Use int value to record string length instead of Map in [`InsertPreparedStatementImpl `](https://github.com/4paradigm/OpenMLDB/blob/main/java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/sdk/impl/InsertPreparedStatementImpl.java#L55)

mangoGoForward:
Bug fix

Fixed #1979

None


mangoGoForward:
This PR contains other commit, I will close it and open a new pr

PR #2044, opened 2022-06-29
fix: use int to record string length instead of map

### LINKED ISSUE #1979: Use int value to record string length instead of Map in `InsertPreparedStatementImpl `

mangoGoForward:
Signed-off-by: mango <xu.weiKyrie@foxmail.com>

Bug fix

Fixed #1979

None

### DIFF
--- a/java/openmldb-jdbc/src/test/java/com/_4paradigm/openmldb/jdbc/SQLRouterSmokeTest.java
+++ b/java/openmldb-jdbc/src/test/java/com/_4paradigm/openmldb/jdbc/SQLRouterSmokeTest.java
@@ -367,7 +367,9 @@ public class SQLRouterSmokeTest {
             insert = "insert into tsql1010 values(?, ?, ?, ?, ?);";
             PreparedStatement impl3 = router.getInsertPreparedStmt(dbname, insert);
             impl3.setLong(1, 1003);
+            impl3.setString(3, "zhejiangxx");
             impl3.setString(3, "zhejiang");
+            impl3.setString(4, "xxhangzhou");
             impl3.setString(4, "hangzhou");
             impl3.setDate(2, d4);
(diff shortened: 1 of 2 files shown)


java/openmldb-jdbc/src/test/java/com/_4paradigm/openmldb/jdbc/SQLRouterSmokeTest.java
```diff
@@ -367,9 +367,7 @@ public void testInsertPreparedState(SqlExecutor router) {
             insert = "insert into tsql1010 values(?, ?, ?, ?, ?);";
             PreparedStatement impl3 = router.getInsertPreparedStmt(dbname, insert);
             impl3.setLong(1, 1003);
-            impl3.setString(3, "zhejiangxx");
```
vagetablechicken: Mind that do not delete these lines

mangoGoForward: Got, I think use `List` or `Map` more better instead of `int`, because we want to implement multiple set and save the last value. So this issue and PR may unnecessary

mangoGoForward: Done, reverted

dl239 (approved):

tobegit3hub (approved):
LGTM

aceforeverd (approved):

PR #2049, opened 2022-06-30
fix: `MakeResultSet` using wrong schema in projection

### LINKED ISSUE #2027: `MakeResultSet` using wrong schema  in projection
`MakeResultSet` using wrong schema  in projection
https://github.com/4paradigm/OpenMLDB/blob/main/src/sdk/result_set_sql.cc#L100-L108

dl239:
The result  returned by `Scan` function in `TableReader` is incorrect currently.


tobegit3hub (approved):
LGTM

aceforeverd (approved):

PR #2114, opened 2022-07-06
fix: resolved different signedness issue

Vijayitsover9000:
it was a label :"good first issue" 
 it resolves the different signedness at line 300 in 
[hybridse/examples/toydb/src/sdk/tablet_sdk.cc]{https://github.com/4paradigm/OpenMLDB/blob/main/hybridse/examples/toydb/src/sdk/tablet_sdk.cc#L300}

### DIFF
--- a/hybridse/examples/toydb/src/sdk/tablet_sdk.cc
+++ b/hybridse/examples/toydb/src/sdk/tablet_sdk.cc
@@ -297,7 +297,7 @@ void TabletSdkImpl::BuildInsertRequest(const std::string& db,
     request->set_db(db);
 
     std::unordered_set<std::string> column_set;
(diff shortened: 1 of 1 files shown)


hybridse/examples/toydb/src/sdk/tablet_sdk.cc
```diff
@@ -297,7 +297,7 @@ void TabletSdkImpl::BuildInsertRequest(const std::string& db,
     request->set_db(db);
 
     std::unordered_set<std::string> column_set;
-    for (size_t i = 0; i < schema.columns().size(); i++) {
+    for (size_t i = 0; i < (unsigned)schema.columns().size(); i++) {
```
dl239: it's better to update the type of i.  `int i = 0;`

dl239 (approved):

PR #2181, opened 2022-07-15
fix(#2177): avoid ts overflow in window

### LINKED ISSUE #2177: window `exclude current_time` report wrong result if request row/current row `ts = 0`, window end value overflow
**Bug Description**

**Expected Behavior**

**Steps to Reproduce**

1. 
2. 
3. 
4.

aceforeverd:
close #2177 .

here is a window `w (.... ROWS_RANGE BETWEEN .. AND 0 preceding EXCLUDE CURRENT_TIME)`, 
and if e.g, there is a row with `ts = 0`,  this row is not correctly handled in window.

- for batch mode: it is added to effective window two times
- for request mode: other rows before current row that `ts = 0` is not excluded


cases/function/window/window_attributes.yaml
```diff
@@ -478,3 +481,53 @@ cases:
          4, 3, 23, 21, 23
          5, 0, NULL, NULL, NULL
          6, 1, 56, 56, 56
+  - id: 9
+    desc: |
+      ROWS Window with exclude current_row and exclude current_row
```
aceforeverd: exclude current_time

aceforeverd:
patch long window, move to #2202

wuyou10206 (approved):
checked

zhanghaohit (approved):

PR #2220, opened 2022-07-22
fix: create with zero partitionnum

### LINKED ISSUE #2217: When partitionnum = 0 and distribution is not specified, the creation of the table should fail
**Bug Description**
When partitionnum = 0 and distribution is not specified, the creation of the table should fail, but now it succeeds
```sql
create table t4 (c1 string NOT NULL,c2 int,c3 timestamp,c4 timestamp,index(key=(c1),ts=c4,ttl=0m))
options (
partitionnum = 0,
replicanum = 3
);
```

**Expected Behavior**
An error message should be prompted

**Steps to Reproduce**

1. 
2. 
3. 
4.

dl239:
- fix #2217 
- support mutiple distributions
https://github.com/4paradigm/OpenMLDB/blob/v0.5.3/hybridse/src/planv2/ast_node_converter.cc#L1680-L1681


vagetablechicken (approved):

tobegit3hub (approved):
LGTM

aceforeverd:
pls fix tests in ast_node_converter_test

PR #2236, opened 2022-07-25
fix: create index core dump

### LINKED ISSUE #475: create index should avoid coredump
If we use sql_client, run `create index i1 on t1(c1);` with no options, `ASTCreateIndexStatement::options_list()` will be nullptr. It will core in
https://github.com/4paradigm/OpenMLDB/blob/d1edb4a26e651400bff29cf281fe2698c85e7bc9/hybridse/src/planv2/ast_node_converter.cc#L1701

It's better to return error if no options? Or we can accept `create index` without ts settings? 
P.S. we can set index without ts settings in `create table` .

### LINKED ISSUE #2233: CLI core dump if not set option in `CREATE INDEX` statement
```
127.0.0.1:30100/test> create index index1 on t4 (c2);
Segmentation fault
```

dl239:

### DIFF
--- a/hybridse/src/planv2/ast_node_converter.cc
+++ b/hybridse/src/planv2/ast_node_converter.cc
@@ -1994,12 +1994,15 @@ base::Status ConvertCreateIndexStatement(const zetasql::ASTCreateIndexStatement*
 
     node::SqlNode* index_key_node = node_manager->MakeIndexKeyNode(keys);
     index_node_list->PushBack(index_key_node);
-    for (const auto option : root->options_list()->options_entries()) {
-        node::SqlNode* node = nullptr;
-        CHECK_STATUS(ConvertIndexOption(option, node_manager, &node));
-        if (node != nullptr) {
-            // NOTE: unhandled option will return OK, but node is not set
-            index_node_list->PushBack(node);
+    if (root->options_list() != nullptr) {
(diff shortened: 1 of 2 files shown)


hybridse/src/planv2/ast_node_converter.cc
```diff
@@ -1994,12 +1994,15 @@ base::Status ConvertCreateIndexStatement(const zetasql::ASTCreateIndexStatement*
...
     node::SqlNode* index_key_node = node_manager->MakeIndexKeyNode(keys);
     index_node_list->PushBack(index_key_node);
-    for (const auto option : root->options_list()->options_entries()) {
-        node::SqlNode* node = nullptr;
-        CHECK_STATUS(ConvertIndexOption(option, node_manager, &node));
-        if (node != nullptr) {
-            // NOTE: unhandled option will return OK, but node is not set
-            index_node_list->PushBack(node);
+    if (root->options_list() != nullptr) {
+        CHECK_TRUE(!root->options_list()->options_entries().empty(), common::kSqlAstError, "no item in options")
```
aceforeverd: `option () ` should be fine ?

dl239: removed

vagetablechicken (approved):

aceforeverd (approved):

PR #2241, opened 2022-07-25
docs: improved en version of sql/dql/group_by

michelle-qinqin:
docs

change the format of example from title formatting to bold
update the boundary table

### DIFF
--- a/docs/en/reference/sql/dql/GROUP_BY_CLAUSE.md
+++ b/docs/en/reference/sql/dql/GROUP_BY_CLAUSE.md
@@ -1,7 +1,5 @@
 # GROUP BY Clause
 
-All -group by- currently only has supports in batch mode (that is, console debugging SQL support, offline mode is still under development)
-
 ## Syntax
 
 ```SQL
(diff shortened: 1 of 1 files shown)


lumianph (approved):

docs/en/reference/sql/dql/GROUP_BY_CLAUSE.md
```diff
@@ -17,25 +15,23 @@ SELECT select_expr [,select_expr...] FROM ... GROUP BY ...
 
 ## Boundary Description
```
lumianph, line 16: @michelle-qinqin I just notice this`boundary description` that appears many times in different docs. This also seems a bit weird (like a Chinglish ...). Maybe we just change it to `Description`?

michelle-qinqin: that's ok with me~
But can I create a new branch and a new PR to change this statement in all related files later?
Since there are many unmerged branches that might have this problem, and there are both Chinese and English docs using `boundary description`, it would be more convenient and safer for me to check all docs for once specifically for this problem.

lumianph: yes, you can fix this issue later in a single PR

lumianph: yes, you can fix this issue later in a single PR

PR #2255, opened 2022-07-27
feat(long window): support *_where in sql engine

### LINKED ISSUE #1296: Support long time window with filters

aceforeverd:

five op supported:
- count_where
- sum_where
- avg_where
- min_where
- max_where

close #1296


src/cmd/sql_cmd_test.cc
```diff
@@ -644,6 +644,12 @@ void CreateDBTableForLongWindow(const std::string& base_db, const std::string& b
     ASSERT_EQ(tables.size(), 1) << msg;
 }
 
+// -----------------------------------------------------------------------------------
+// col1 col2 col3 i64_col i16_col i32_col f_col d_col t_col  s_col  date_col   filter
+// str1 str2  i     i       i       i       i     i     i      i    1900-01-i  i % 2
+//
+// where i in [i .. 11]
```
zhanghaohit: i in [1 .. 11]

src/cmd/sql_cmd_test.cc
```diff
@@ -1514,22 +1522,26 @@ TEST_P(DBSDKTest, DeployLongWindowsExecuteCountWhere) {
...
+        count_where(i64_col, filter<1) over w1 as w1_count_where_i64_col_filter,
+        count_where(i64_col, col1='str1') over w1 as w1_count_where_i64_col_col1,
+        count_where(i16_col, filter>1) over w1 as w1_count_where_i16_col,
+        count_where(i32_col, 1<filter) over w1 as w1_count_where_i32_col,
+        count_where(f_col, 0=filter) over w1 as w1_count_where_f_col,
+        count_where(d_col, 1=filter) over w1 as w1_count_where_d_col,
+        count_where(t_col, 1>=filter) over w1 as w1_count_where_t_col,
+        count_where(s_col, 2<filter) over w1 as w1_count_where_s_col,
+        count_where(date_col, 2>filter) over w1 as w1_count_where_date_col,
+        count_where(col3, 0>=filter) over w2 as w2_count_where_col3 from )" +
```
zhanghaohit, line 1709: possible to add ut for other filter type except the `int`?

zhanghaohit (dismissed):

zhanghaohit:
Did u verify for  those no-filter test (e..g, `DeployLongWindowsExecuteCount`) if the `long window` opt is really taking effect rather than reverting back to normal cases?

src/cmd/sql_cmd_test.cc
aceforeverd: added in next test case

aceforeverd:
> Did u verify for those no-filter test (e..g, `DeployLongWindowsExecuteCount`) if the `long window` opt is really taking effect rather than reverting back to normal cases?

For debug mode build, it prints physical plan and runner infos which did runs on pre-agg mode
Screen Shot 2022-08-02 at 6 55 07 PM

Anyway there is no tests in UT code itself. I think we should be able to test long window result in sql engine independently, so it will be much easier to 
- test correctness of long window query by a yaml case, as well as correctness for each compile phase
- much easier to debug (single UT, great for debugger)

vagetablechicken (approved):

src/storage/aggregator.cc
```diff
@@ -123,16 +123,25 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
     }
 
     AggrBufferLocked* aggr_buffer_lock;
+    // If filter key not empty, ts_end of AggrBuffer should aligned
+    // with AggrBuffer with same key but different filter key.
+    // Designed alignment only for range window, left TODO for rows window
+    int64_t aligned_end = -1;
```
zhanghaohit: `ts_start` also has to be aligned.

aceforeverd:
[pre-aggregate support xxx_where](https://trello.com/c/qFWsyu8b/66-pre-aggregate-support-xxxwhere)

hybridse/src/vm/runner.cc
```diff
@@ -2985,47 +3035,118 @@ std::shared_ptr<TableHandler> RequestAggUnionRunner::RequestUnionWindow(
...
+                    agg_it->Next();
+                    continue;
+                }
+                if (0 != agg_row_parser->GetString(drow, "filter_key", &filter_val)) {
+                    LOG(ERROR) << "failed to get value of filter_key";
+                    agg_it->Next();
+                    continue;
+                }
+
+                if (prev_ts_start == ts_start && filter_val == prev_filter_val) {
```
zhanghaohit: here we are assuming the rows with same `filter_val` comes together. 

However it may not be true. e.g.,

ts_start | ts_end | filter_val
|-----|-------|------|
0 | 10 | 0
0 | 10 | 1
0 | 10 | 0
0 | 10 | 1

It doesn't work in this case.

src/cmd/sql_cmd_test.cc
```diff
@@ -2310,6 +2323,67 @@ TEST_P(DBSDKTest, LongWindowAnyWhereUnsupportTimeFilter) {
...
+            ASSERT_FALSE(status.IsOK());
+            EXPECT_EQ(status.msg, "avg_where only support over memory base table")
+                << "code=" << status.code << ", msg=" << status.msg << "\n"
+                << status.trace;
+        }
+
+        void TearDownPreAggTables() override {}
+    };
+
+    // unsupport: deploy any_where with rows bucket
```
zhanghaohit: with hdd table

hybridse/src/vm/runner.cc
aceforeverd: yes, using `filter_val` as index key should fix this. 

Or we can use a set to workaroud this, just not performant.

zhanghaohit: as a quickfix, maybe first we use a `set` to check the occurrence of `filter_val`

aceforeverd: fixed, this case should cover https://github.com/4paradigm/OpenMLDB/pull/2255/files#diff-3f6b69529b323ca236ac25ae70c301269579a5ce1442d569066786bf009e6023R2219

aceforeverd: fixed, this case should cover https://github.com/4paradigm/OpenMLDB/pull/2255/files#diff-3f6b69529b323ca236ac25ae70c301269579a5ce1442d569066786bf009e6023R2219

zhanghaohit (approved):
LGTM

PR #2277, opened 2022-07-30
feat(api-server): add query parameters to the sql query api

### LINKED ISSUE #2128: feat(api-server): add query parameters to the sql query api

enhance #2091 

**Additional context**

qsliu2017:
fix #2128 

add query parameters to the sql query api

### DIFF
--- a/docs/en/quickstart/rest_api.md
+++ b/docs/en/quickstart/rest_api.md
@@ -79,16 +79,14 @@ The request URL: http://ip:port/dbs/{db_name}
 
 HTTP method: POST
 
-The request body example: 
+**Request Body Example**
+
+The query without parameter: 
 
 ```json
 {
     "mode": "online",
-    "sql": "SELECT c1, c2, c3 FROM demo WHERE c1 = ? AND c2 = ?",
-    "parameter": {
-      "schema": ["Int32", "String"],
-      "data": [1, "aaa"]
-    }
+    "sql": "select 1"
 }
 ```
 
@@ -99,8 +97,33 @@ The response:
 ```json
 {
     "code":0,
+    "msg":"ok"
+}
+```
+
+The query with parameters:
+
+```json
+{
+    "mode": "online",
+    "sql": "SELECT c1, c2, c3 FROM demo WHERE c1 = ? AND c2 = ?",
+    "input": {
+      "schema": ["Int32", "String"],
+      "data": [1, "aaa"]
+    }
+}
+```
+
+all supported types (case-insensitive):
+`Bool`, `Int16`, `Int32`, `Int64`, `Float`, `Double`, `String`, `Date` and `Timestamp`.
+
+The response:
+
+```json
+{
+    "code":0,
     "msg":"ok",
-    "result": {
+    "data": {
       "schema": ["Int32", "String", "Float"],
       "data": [[1, "aaa", 1.2], [1, "aaa", 3.4]]
     }
--- a/docs/en/quickstart/rest_api.md
+++ b/docs/en/quickstart/rest_api.md
@@ -84,7 +84,11 @@ The request body example:
 ```json
 {
     "mode": "online",
-    "sql": "select 1"
+    "sql": "SELECT c1, c2, c3 FROM demo WHERE c1 = ? AND c2 = ?",
+    "parameter": {
+      "schema": ["Int32", "String"],
+      "data": [1, "aaa"]
+    }
 }
 ```
 
@@ -95,7 +99,11 @@ The response:
 ```json
 {
     "code":0,
-    "msg":"ok"
+    "msg":"ok",
+    "result": {
+      "schema": ["Int32", "String", "Float"],
+      "data": [[1, "aaa", 1.2], [1, "aaa", 3.4]]
+    }
 }
 ```
 
--- a/src/apiserver/api_server_impl.cc
+++ b/src/apiserver/api_server_impl.cc
@@ -104,51 +104,52 @@ std::map<std::string, ExecContext> mode_map{
     {"offsync", {false, true}}, {"offasync", {false, false}}, {"online", {true, false}}};
 
 void APIServerImpl::RegisterQuery() {
-    provider_.post("/dbs/:db_name",
-                   [this](const InterfaceProvider::Params& param, const butil::IOBuf& req_body, JsonWriter& writer) {
-                       auto resp = GeneralResp();
-                       auto db_it = param.find("db_name");
-                       if (db_it == param.end()) {
-                           writer << resp.Set("url has no db_name");
-                           return;
-                       }
-                       auto db = db_it->second;
+    provider_.post("/dbs/:db_name", [this](const InterfaceProvider::Params& param, const butil::IOBuf& req_body,
+                                           JsonWriter& writer) {
+        auto resp = GeneralResp();
+        auto db_it = param.find("db_name");
+        if (db_it == param.end()) {
+            writer << resp.Set("url has no db_name");
+            return;
+        }
+        auto db = db_it->second;
 
-                       // default mode is offsync
-                       QueryReq req;
-                       JsonReader query_reader(req_body.to_string().c_str());
-                       query_reader >> req;
-                       if (!query_reader) {
-                           writer << resp.Set("Json parse failed, " + req_body.to_string());
-                           return;
-                       }
-                       auto mode = boost::to_lower_copy(req.mode);
-                       auto it = mode_map.find(mode);
-                       if (it == mode_map.end()) {
-                           writer << resp.Set("Invalid mode " + mode);
-                           return;
-                       }
-                       ExecContext ctx = it->second;
+        // default mode is offsync
+        QueryReq req;
+        JsonReader query_reader(req_body.to_string().c_str());
+        query_reader >> req;
+        if (!query_reader) {
+            writer << resp.Set("Json parse failed, " + req_body.to_string());
+            return;
+        }
+        auto mode = boost::to_lower_copy(req.mode);
+        auto it = mode_map.find(mode);
+        if (it == mode_map.end()) {
+            writer << resp.Set("Invalid mode " + mode);
+            return;
+        }
+        ExecContext ctx = it->second;
 
-                       const auto& sql = req.sql;
-                       VLOG(1) << "post [" << ctx.ToString() << "] query on db [" << db << "], sql: " << sql;
-                       // TODO(hw): if api server supports standalone, we should check if cluster mode here
+        const auto& sql = req.sql;
+        const auto parameter = req.parameter;
+        VLOG(1) << "post [" << ctx.ToString() << "] query on db [" << db << "], sql: " << sql;
+        // TODO(hw): if api server supports standalone, we should check if cluster mode here
 
-                       hybridse::sdk::Status status;
-                       // TODO(hw): if sql is not a query, it may be a ddl, we use ExecuteSQL to execute it before we
-                       //  supports ddl http api. It's useful for api server tests(We can create table when we only
-                       //  connect to the api server).
-                       auto rs = sql_router_->ExecuteSQL(db, sql, ctx.is_online, ctx.is_sync, ctx.job_timeout, &status);
-                       if (!status.IsOK()) {
-                           writer << resp.Set(status.code, status.msg);
-                           LOG(WARNING) << "failed at: code " << status.code << ", msg " << status.msg;
-                           return;
-                       }
+        hybridse::sdk::Status status;
+        // TODO(hw): if sql is not a query, it may be a ddl, we use ExecuteSQL to execute it before we
+        //  supports ddl http api. It's useful for api server tests(We can create table when we only
+        //  connect to the api server).
+        auto rs = sql_router_->ExecuteSQL(db, sql, parameter, ctx.is_online, ctx.is_sync, ctx.job_timeout, &status);
+        if (!status.IsOK()) {
+            writer << resp.Set(status.code, status.msg);
+            LOG(WARNING) << "failed at: code " << status.code << ", msg " << status.msg;
+            return;
+        }
 
-                       QueryResp query_resp;
-                       query_resp.rs = rs;
-                       writer << query_resp;
-                   });
+        QueryResp query_resp;
+        query_resp.rs = rs;
+        writer << query_resp;
+    });
 }
 
 bool APIServerImpl::Json2SQLRequestRow(const butil::rapidjson::Value& non_common_cols_v,
--- a/src/apiserver/api_server_impl.cc
+++ b/src/apiserver/api_server_impl.cc
@@ -633,6 +633,148 @@ std::string APIServerImpl::InnerTypeTransform(const std::string& s) {
     return out;
 }
 
+JsonReader& operator&(JsonReader& ar, QueryReq& s) {  // NOLINT
+    ar.StartObject();
+    // mode is not optional
+    ar.Member("mode") & s.mode;
+    ar.Member("sql") & s.sql;
+    if (ar.HasMember("parameters")) {
+        ar.Member("parameters") & s.parameter;
+    }
+    return ar.EndObject();
+}
+
+JsonReader& operator&(JsonReader& ar, std::shared_ptr<openmldb::sdk::SQLRequestRow>& parameter) {  // NOLINT
+    ar.StartObject();                                                                              // start "parameter"
+    if (!ar.HasMember("schema") || !ar.HasMember("data")) return ar;
+
+    ::hybridse::vm::Schema schema;
+    {
+        ar.Member("schema");
+        size_t size;
+        ar.StartArray(&size);  // start "schema"
+        for (auto i = 0; i < size; i++) {
+            std::string type;
+            ar& type;
+            // uppercase
+            std::transform(type.begin(), type.end(), type.begin(), [](unsigned char c) { return std::toupper(c); });
+
+            auto col = schema.Add();
+            if (type == "BOOL") {
+                col->set_type(::hybridse::type::kBool);
+            } else if (type == "SMALLINT" || type == "INT16") {
+                col->set_type(::hybridse::type::kInt16);
+            } else if (type == "INT" || type == "INT32") {
+                col->set_type(::hybridse::type::kInt32);
+            } else if (type == "BIGINT" || type == "INT64") {
+                col->set_type(::hybridse::type::kInt64);
+            } else if (type == "FLOAT") {
+                col->set_type(::hybridse::type::kFloat);
+            } else if (type == "DOUBLE") {
+                col->set_type(::hybridse::type::kDouble);
+            } else if (type == "STRING") {
+                col->set_type(::hybridse::type::kVarchar);
+            } else if (type == "DATE") {
+                col->set_type(::hybridse::type::kDate);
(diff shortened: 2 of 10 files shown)


aceforeverd:
[ospp - Golang SDK](https://trello.com/c/7aquDamb/118-ospp-golang-sdk)

src/apiserver/api_server_impl.cc
```diff
@@ -104,51 +104,103 @@ std::map<std::string, ExecContext> mode_map{
...
+                if (it->IsBool()) {
+                    types->AddColumnType(hybridse::sdk::kTypeBool);
+                } else if (it->IsDouble()) {
+                    types->AddColumnType(hybridse::sdk::kTypeDouble);
+                } else if (it->IsInt64()) {
+                    types->AddColumnType(hybridse::sdk::kTypeInt64);
+                } else if (it->IsString()) {
+                    types->AddColumnType(hybridse::sdk::kTypeString);
+                } else {
+                    writer << resp.Set("Invalid parameter type, " + it->GetType());
```
aceforeverd: rhs is `Type` not string, can't append

src/apiserver/api_server_impl.cc
```diff
@@ -104,51 +104,103 @@ std::map<std::string, ExecContext> mode_map{
...
+        auto db_it = param.find("db_name");
+        if (db_it == param.end()) {
+            writer << resp.Set("url has no db_name");
+            return;
+        }
+        auto db = db_it->second;
+
+        // default mode is offsync
+        Document document;
+        if (document.Parse(req_body.to_string().c_str()).HasParseError()) {
```
aceforeverd: how about just change `QueryReq` struct

src/apiserver/api_server_impl.cc
```diff
@@ -104,51 +104,103 @@ std::map<std::string, ExecContext> mode_map{
...
+
+            auto types = std::make_shared<hybridse::sdk::ColumnTypes>();
+            for (auto it = _parameter->value.Begin(); it != _parameter->value.End(); it++) {
+                if (it->IsBool()) {
+                    types->AddColumnType(hybridse::sdk::kTypeBool);
+                } else if (it->IsDouble()) {
+                    types->AddColumnType(hybridse::sdk::kTypeDouble);
+                } else if (it->IsInt64()) {
+                    types->AddColumnType(hybridse::sdk::kTypeInt64);
+                } else if (it->IsString()) {
```
aceforeverd: other types ?

src/apiserver/api_server_impl.cc
```diff
@@ -104,51 +104,103 @@ std::map<std::string, ExecContext> mode_map{
...
+                    parameter->AppendDouble(it->GetDouble());
+                } else if (it->IsInt64()) {
+                    parameter->AppendInt64(it->GetInt64());
+                } else if (it->IsString()) {
+                    parameter->AppendString(it->GetString());
+                }
+            }
+        }
+
+        parameter->Build();
```
aceforeverd: check success

src/apiserver/json_helper.cc
```diff
@@ -200,6 +203,75 @@ JsonReader& JsonReader::operator&(std::string& s) {  // NOLINT
...
+            } else {
+                error_ = true;
+            }
+            Next();
+        }
+    }
+    this->EndArray();
+
+    ::hybridse::sdk::SchemaImpl* schema_impl = new ::hybridse::sdk::SchemaImpl(schema);
+    std::shared_ptr<::hybridse::sdk::Schema> schema_shared(schema_impl);
```
aceforeverd: fix the three cpplint report

src/apiserver/json_helper.cc
```diff
@@ -200,6 +203,75 @@ JsonReader& JsonReader::operator&(std::string& s) {  // NOLINT
...
+    std::set<std::string> _rs{};
```
aceforeverd: same here

src/apiserver/json_helper.cc
```diff
@@ -200,6 +203,75 @@ JsonReader& JsonReader::operator&(std::string& s) {  // NOLINT
...
+            case ::hybridse::type::kDouble:
+                ok = parameter->AppendDouble(val.GetDouble());
+                break;
+            case ::hybridse::type::kVarchar:
+                ok = parameter->AppendString(val.GetString());
+                break;
+            default:
+                break;
+        }
+        if (!ok) goto fast_fail;
```
aceforeverd: For maintenance it's better not using `goto`

src/apiserver/json_helper.cc
```diff
@@ -200,6 +203,75 @@ JsonReader& JsonReader::operator&(std::string& s) {  // NOLINT
...
+    if (!parameter->Init(str_length)) goto fast_fail;
+    for (auto i = 0; i < size; i++) {
+        auto col = schema.Get(i);
+        auto& val = values.at(i);
+        bool ok = false;
+        switch (col.type()) {
+            case ::hybridse::type::kBool:
+                ok = parameter->AppendBool(val.GetBool());
+                break;
+            case ::hybridse::type::kInt64:
```
aceforeverd: more type need supported, include all numeric and string type, date and timestamp can be omitted currently

qsliu2017: these types are parsed from JSON ([qsliu2017/OpenMLDB@093e2ce/src/apiserver/json_helper.cc#L217](https://github.com/qsliu2017/OpenMLDB/blob/093e2ce560fb2a106ac7cbe3e2305e62fcdea4da/src/apiserver/json_helper.cc#L217)). since JSON only have several types, we only cover these types.

src/apiserver/json_helper.cc
```diff
@@ -200,6 +207,137 @@ JsonReader& JsonReader::operator&(std::string& s) {  // NOLINT
     return *this;
 }
 
+JsonReader& JsonReader::operator&(std::shared_ptr<openmldb::sdk::SQLRequestRow>& parameter) {  // NOLINT
+
```
aceforeverd: cpplint

src/apiserver/json_helper.cc
```diff
@@ -200,6 +207,137 @@ JsonReader& JsonReader::operator&(std::string& s) {  // NOLINT
...
+    {
+        this->Member("schema");
+        size_t size;
+        this->StartArray(&size);  // start "schema"
+        for (auto i = 0; i < size; i++) {
+            if (!error_) {
+                auto col = schema.Add();
+                auto type = std::string(CURRENT.GetString());
+
+                // lowercase
```
aceforeverd: upper

src/apiserver/json_helper.cc
```diff
@@ -200,6 +207,137 @@ JsonReader& JsonReader::operator&(std::string& s) {  // NOLINT
...
+                    break;
+                case ::hybridse::type::kVarchar:
+                    ok = parameter->AppendString(CURRENT.GetString(), CURRENT.GetStringLength());
+                    break;
+                case ::hybridse::type::kDate:
+                    ok = parameter->AppendDate(CURRENT.GetInt());
+                    break;
+                case ::hybridse::type::kTimestamp:
+                    ok = parameter->AppendTimestamp(CURRENT.GetInt64());
+                    break;
```
aceforeverd: no default branch

src/apiserver/json_helper.cc
```diff
@@ -200,6 +207,137 @@ JsonReader& JsonReader::operator&(std::string& s) {  // NOLINT
...
+                str_length += CURRENT.GetStringLength();
+            }
+            Next();
+        }
+        this->EndArray();  // end first iter "data"
+    }
+    {
+        ::hybridse::sdk::SchemaImpl* schema_impl = new ::hybridse::sdk::SchemaImpl(schema);
+        std::shared_ptr<::hybridse::sdk::Schema> schema_shared(schema_impl);
+        std::set<std::string> _rs{};
```
dl239: remove `_` prefix

src/apiserver/api_server_impl.h
```diff
@@ -92,6 +94,9 @@ Archiver& operator&(Archiver& ar, QueryReq& s) {  // NOLINT
     // mode is not optional
     ar.Member("mode") & s.mode;
     ar.Member("sql") & s.sql;
+    if (ar.HasMember("parameters")) {
```
vagetablechicken: JsonWriter HasMember is always false. If you don't want to support write `QueryReq ` to json, delete the template, just impl JsonReader reads QueryReq

src/apiserver/api_server_test.cc
```diff
@@ -184,6 +184,81 @@ TEST_F(APIServerTest, query) {
...
+    ASSERT_TRUE(env->cluster_sdk->Refresh());
+    ASSERT_TRUE(env->cluster_remote->ExecuteInsert(env->db, insert_sql, &status));
+
+    {
+        brpc::Controller cntl;
+        cntl.http_request().set_method(brpc::HTTP_METHOD_POST);
+        cntl.http_request().uri() = "http://127.0.0.1:8010/dbs/" + env->db;
+        cntl.request_attachment().append(R"({
+            "sql": "select c1, c2 from demo where c2 = ?;", "mode": "online", "parameters": {
+                "schema": ["STRING"],
```
vagetablechicken, line 223: Why need to input schema?

src/apiserver/json_helper.cc
```diff
@@ -200,6 +207,137 @@ JsonReader& JsonReader::operator&(std::string& s) {  // NOLINT
     return *this;
 }
 
+JsonReader& JsonReader::operator&(std::shared_ptr<openmldb::sdk::SQLRequestRow>& parameter) {  // NOLINT
```
vagetablechicken: json_helper is the util, do not do impl read/write methods in it. Move it to api_server_impl, or create a new file for all req&resp class.

src/apiserver/json_helper.h
```diff
@@ -64,11 +66,12 @@ class JsonReader {
...
-    JsonReader& operator&(unsigned& u);     // NOLINT
-    JsonReader& operator&(int& i);          // NOLINT
-    JsonReader& operator&(double& d);       // NOLINT
-    JsonReader& operator&(std::string& s);  // NOLINT
+    JsonReader& operator&(bool& b);                                                   // NOLINT
+    JsonReader& operator&(unsigned& u);                                               // NOLINT
+    JsonReader& operator&(int& i);                                                    // NOLINT
+    JsonReader& operator&(double& d);                                                 // NOLINT
+    JsonReader& operator&(std::string& s);                                            // NOLINT
+    JsonReader& operator&(std::shared_ptr<openmldb::sdk::SQLRequestRow>& parameter);  // NOLINT
```
vagetablechicken: SQLRequestRow is not a basic type, don't impl in here

src/sdk/sql_cluster_router.cc
```diff
@@ -2438,6 +2438,12 @@ std::shared_ptr<hybridse::sdk::ResultSet> SQLClusterRouter::ExecuteSQL(const std
                                                                        bool is_online_mode, bool is_sync_job,
                                                                        int offline_job_timeout,
                                                                        hybridse::sdk::Status* status) {
+    return ExecuteSQL(db, sql, {}, is_online_mode, is_sync_job, offline_job_timeout, status);
+}
+
+std::shared_ptr<hybridse::sdk::ResultSet> SQLClusterRouter::ExecuteSQL(
```
vagetablechicken, line 2444: So it should be ExecuteSQLParameterized...

src/apiserver/api_server_test.cc
qsliu2017: because OpenMLDB support `int[16|32|64]` etc., while JSON format only has several types.

src/apiserver/api_server_test.cc
```diff
@@ -184,6 +184,81 @@ TEST_F(APIServerTest, query) {
...
+
+        butil::rapidjson::Document document;
+        if (document.Parse(cntl.response_attachment().to_string().c_str()).HasParseError()) {
+            ASSERT_TRUE(false) << "response parse failed with code " << document.GetParseError()
+                               << ", raw resp: " << cntl.response_attachment().to_string();
+        }
+        ASSERT_EQ(0, document["code"].GetInt());
+        ASSERT_STREQ("ok", document["msg"].GetString());
+        ASSERT_EQ(2, document["data"].Size());
+        ASSERT_EQ(2, document["data"][0].Size());
```
aceforeverd: return schema for response rows

aceforeverd (changes requested):

src/apiserver/api_server_test.cc
vagetablechicken: You can get the exact type by request row, user doesn't have to write the type

vagetablechicken: ref Json2SQLRequestRow

vagetablechicken (changes requested):
Add doc about it. `docs/zh/quickstart/rest_api.md`

src/apiserver/api_server_test.cc
qsliu2017: we do NOT have a SCHEMA for a query.
e.g. `select c1, c2 from demo where c2 = ?;` needs params of schema `["STRING"]` while `select c1, c2 from demo where c1 = ? AND c2 = ?;` needs `["INT", "STRING"]`

vagetablechicken: We CAN have it, users don't need to do this. see ExecuteProcedure. Plz read it before you stick to your vision

qsliu2017: https://github.com/4paradigm/OpenMLDB/blob/5ac4667d0e71dc803ddbd04928bc4681b5ba84b6/src/apiserver/api_server_impl.cc#L400-L410

If I understand right, `ExecuteProcedure` gets parameter schema from the procedure already stored in the cluster, so the user doesn't need to input it again.

But for query, we do not have such stored info.

vagetablechicken: You CAN get it in cache too, but you don't try to.
see GetSQLCache, RouterSQLCache::GetParameterSchema. We can't get schema by sql complier, and get it from cache?

docs/en/quickstart/rest_api.md
```diff
@@ -95,7 +99,11 @@ The response:
 ```json
 {
     "code":0,
-    "msg":"ok"
+    "msg":"ok",
+    "result": {
```
aceforeverd: we should keep it consistent with [feature extract api](https://github.com/4paradigm/OpenMLDB/blob/44ddfde5e9a0edae6c23b46a286ea477091b3a67/docs/en/quickstart/rest_api.md#real-time-feature-extraction), use `"data"` as filed name

src/apiserver/api_server_test.cc
qsliu2017: I've tried `SQLRouter::GetRequestRow` to extract schema from query, but it turns out `Non-support kFilterPlan Op in online serving`

<details>
<summary>Details</summary>

```
W0912 14:00:26.360630 12917 sql_compiler.cc:329] Fail create sql plan: Non-support kFilterPlan Op in online serving
    (At /workspace/OpenMLDB/hybridse/src/planv2/planner_v2.cc:44)
    (At /workspace/OpenMLDB/hybridse/src/plan/planner.cc:673)
    (At /workspace/OpenMLDB/hybridse/src/plan/planner.cc:510)
    (At /workspace/OpenMLDB/hybridse/src/plan/planner.cc:510)
    (At /workspace/OpenMLDB/hybridse/src/plan/planner.cc:504)
    (Caused by) Non-support kFilterPlan Op in online serving
W0912 14:00:26.361176 12917 sql_cluster_router.cc:1488] fail to explain sql select c1, c2 from demo where c2 = ?;
```

</details>

aceforeverd: where clause is not supported in online serving mode, try other sql like last join

aceforeverd: `GetRequestRow` is for request row, not same as parameter row. @vagetablechicken how is parameter row constructed in java sdk ?

vagetablechicken: > `GetRequestRow` is for request row, not same as parameter row. @vagetablechicken how is parameter row constructed in java sdk ?

PreparedStatementImpl, `CreateSQLRequestRowFromColumnTypes` and `ExecuteSQLParameterized` too. What I expect is 
getting parameter schema by sql compiler, like `GetInsertInfo`, not by the request row we sent.
Does it difficult? Or we can do it later. This pr send request row with schema

docs/en/quickstart/rest_api.md
```diff
@@ -84,7 +84,11 @@ The request body example:
 ```json
 {
     "mode": "online",
-    "sql": "select 1"
+    "sql": "SELECT c1, c2, c3 FROM demo WHERE c1 = ? AND c2 = ?",
```
vagetablechicken, line 109: add a new example. Query can do normal sql and parameterized sql, do not mislead users to do only parameterized sql

docs/en/quickstart/rest_api.md
```diff
@@ -84,7 +84,11 @@ The request body example:
 ```json
 {
     "mode": "online",
-    "sql": "select 1"
+    "sql": "SELECT c1, c2, c3 FROM demo WHERE c1 = ? AND c2 = ?",
+    "parameter": {
+      "schema": ["Int32", "String"],
```
vagetablechicken, line 111: give all types in schema, `Int32`(the first letter uppercase) is uncommon. If case-insensitive, let us know

src/apiserver/api_server_test.cc
qsliu2017: we can keep `schema` field for now. after removing `schema` and the SDK that sends this field will still work.

src/apiserver/api_server_test.cc
```diff
@@ -178,21 +178,19 @@ TEST_F(APIServerTest, query) {
         {
             "code": 0,
             "msg": "ok",
-            "result": {
-                "schema": ["Int32", "String"],
-                "data": [[1, "bb"]]
-            }
+            "schema": ["Int32", "String"],
```
aceforeverd:
```json
data: {
  schema
   data
}
```
should be more reasonable, a sql stmt not always has return rows

src/apiserver/api_server_test.cc
vagetablechicken: > we can keep `schema` field for now. after removing `schema` and the SDK that sends this field will still work.

@qsliu2017 what do you mean "still work after removing schema"?

src/apiserver/api_server_test.cc
qsliu2017: query that does not return rows will not have `schema` and `data` field

<details>
<summary>Details</summary>

The request body example: 

```json
{
    "mode": "online",
    "sql": "select 1"
}
```
mode: "offsync", "offasync", "online"
The response:
```json
{
    "code":0,
    "msg":"ok"
}
```

</details>

aceforeverd: you might to check both `schema` and `data` to ensure whether rows returned. 

Another reason to use later one is to be consistent with feature extract api: https://github.com/4paradigm/OpenMLDB/blob/44ddfde5e9a0edae6c23b46a286ea477091b3a67/docs/en/quickstart/rest_api.md#real-time-feature-extraction

aceforeverd (approved):
LGTM

vagetablechicken (approved):

Issue #2148, opened 2022-07-12
log: if CLI remains open for a long time, get so many sampler warn log


vagetablechicken:
```
W0712 10:15:22.106148 129511424 sampler.cpp:189] bvar is busy at sampling for 2 seconds!
```
brpc log, maybe we should avoid to print it on terminal too, redirect to log file.

aceforeverd:
this seems important same as #1500 ?

aceforeverd:
@dl239 will you kindly fix this together with #2117 ?  Both issues may have similar solution in end user side.

aceforeverd:
another place may produce error log in cli:
https://github.com/aceforeverd/fedb/blob/25166e447b2b7185cdf589ebc18ae6a51fbc610e/hybridse/src/vm/sql_compiler.cc#L329

PR #2292, opened 2022-08-03
fix: #2235 enable trace

### LINKED ISSUE #2235: `enable_trace` is broken for cmd like sql query
refer to original pr https://github.com/4paradigm/OpenMLDB/pull/1130/files#

aceforeverd:
close #2235 

when `set @@enable_trace = 'true';`, should work for 
- sql query 
- deploy
- select into

For any other cmd, interface is ready, whether it prints trace depends on implementation, fix it just by putting info into `status.trace`

### DIFF
--- a/hybridse/include/sdk/base.h
+++ b/hybridse/include/sdk/base.h
@@ -18,10 +18,14 @@
 #define HYBRIDSE_INCLUDE_SDK_BASE_H_
 
 #include <stdint.h>
+
 #include <memory>
+#include <ostream>
 #include <string>
-#include <vector>
 #include <unordered_map>
+#include <vector>
+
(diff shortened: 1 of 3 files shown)


zhanghaohit (approved):

wuyou10206 (approved):
checked，deploy/select/select...into when an error occurs, CLI prints trace back, but at the same time, a log appears, and the problem has been issued

aceforeverd:
> checked，deploy/select/select...into when an error occurs, CLI prints trace back, but at the same time, a log appears, and the problem has been issued

issued here https://github.com/4paradigm/OpenMLDB/issues/2148#issuecomment-1204703195

PR #2300, opened 2022-08-04
fix: delete data from associated pre-aggr table

### LINKED ISSUE #1673: delete key in pre-aggr table when delete occur in base table
**Bug Description**
aggregator is not supported `delete` yet, base and aggregate tables should be kept consistent.

**Expected Behavior**
when delete on base table, pre-aggr table and aggregator should invalid these rows.

**Steps to Reproduce**

1. 
2. 
3. 
4.

zhanghaohit:
Fix #1673

#2307 has to be fixed before this PR work correctly.

### DIFF
--- a/src/storage/aggregator.cc
+++ b/src/storage/aggregator.cc
@@ -355,16 +355,14 @@ bool Aggregator::Init(std::shared_ptr<LogReplicator> base_replicator) {
         if (cur_offset >= entry.log_index()) {
             continue;
         }
-
-        // TODO(nauta): When the base table key is deleted, the pre-aggr table needs to be deleted at the same time.
-        if (entry.has_method_type() && entry.method_type() == ::openmldb::api::MethodType::kDelete) {
-            PDLOG(WARNING, "unsupport delete method for pre-aggr table");
-            continue;
-        }
         for (int i = 0; i < entry.dimensions_size(); i++) {
             const auto& dimension = entry.dimensions(i);
             if (dimension.idx() == index_pos_) {
-                Update(dimension.key(), entry.value(), entry.log_index(), true);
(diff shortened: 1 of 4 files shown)


nautaa:
https://github.com/4paradigm/OpenMLDB/blob/309dc04cdc16f5529ea09122c4e0d2849d1df90a/src/storage/aggregator.cc#L359-L363

There seems to be nothing to do, you can remove the warning log.

zhanghaohit:
> 

here if we found a `kDelete` entry, we shall do `Delete(key)` again, right? in case during the recover, we recover the key in the `aggr_buffer_map_`

zhanghaohit:
> https://github.com/4paradigm/OpenMLDB/blob/309dc04cdc16f5529ea09122c4e0d2849d1df90a/src/storage/aggregator.cc#L359-L363
> 
> There seems to be nothing to do, you can remove the warning log.

here if we found a `kDelete` entry, we shall do `Delete(key)` again, right? in case during the recover, we recover the key in the `aggr_buffer_map_`

nautaa:
> > https://github.com/4paradigm/OpenMLDB/blob/309dc04cdc16f5529ea09122c4e0d2849d1df90a/src/storage/aggregator.cc#L359-L363
> > 
> > There seems to be nothing to do, you can remove the warning log.
> 
> here if we found a `kDelete` entry, we shall do `Delete(key)` again, right? in case during the recover, we recover the key in the `aggr_buffer_map_`

yep, you are right. aggregator maybe update in recovery phase.

zhanghaohit:
merged after 060 release

dl239 (approved):

aceforeverd (approved):

wuyou10206 (approved):
checked

PR #2320, opened 2022-08-08
Add the delete interface with the specified timestamp to the tablet

XLC-2:
Bug fix

In the distributed scenario, write failure will lead to inconsistent data on different tablets

All tablets can roll back the data that failed to be written to ensure that the data of each tablet is consistent

### DIFF
--- a/src/client/tablet_client.h
+++ b/src/client/tablet_client.h
@@ -95,6 +95,10 @@ class TabletClient : public Client {
     bool Delete(uint32_t tid, uint32_t pid, const std::string& pk, const std::string& idx_name,
                 std::string& msg);  // NOLINT
 
+    bool Delete(uint32_t tid, uint32_t pid, const std::string& pk, const std::string& idx_name,
+                uint64_t& ts,           // NOLINT
+                std::string& msg);      // NOLINT
+
     bool Count(uint32_t tid, uint32_t pid, const std::string& pk, const std::string& idx_name, bool filter_expired_data,
                uint64_t& value, std::string& msg);  // NOLINT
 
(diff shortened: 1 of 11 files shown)


src/client/tablet_client.h
```diff
@@ -95,6 +95,10 @@ class TabletClient : public Client {
     bool Delete(uint32_t tid, uint32_t pid, const std::string& pk, const std::string& idx_name,
                 std::string& msg);  // NOLINT
 
+    bool Delete(uint32_t tid, uint32_t pid, const std::string& pk, const std::string& idx_name,
+                uint64_t& ts,           // NOLINT
```
dl239, line 99: `uint64_t ts`

src/tablet/tablet_impl.cc
```diff
@@ -1546,7 +1546,7 @@ void TabletImpl::Delete(RpcController* controller, const ::openmldb::api::Delete
         }
         idx = index_def->GetId();
     }
-    if (table->Delete(request->key(), idx)) {
+    if (table->Delete(request->key(), idx, request->ts())) {
```
dl239, line 1549: `if request->has_ts()` use new delete interface, else call the old

src/storage/segment.cc
```diff
@@ -317,6 +317,45 @@ bool Segment::Delete(const Slice& key) {
...
+}
+
+bool Segment::Delete(const Slice& key, uint32_t idx, const uint64_t time) {
+    ::openmldb::base::Node<uint64_t, DataBlock*>* entry_ts_node = NULL;
+    DataBlock* block = NULL;
+    {
+        auto pos = ts_idx_map_.find(idx);
+        if (pos == ts_idx_map_.end()) {
+            return false;
+        }
```
dl239, line 359: delete in lock scope

PR #2321, opened 2022-08-09
fix: pre-aggr table indexing schema

### LINKED ISSUE #2307: filter_key is not added into the dimension key
**Bug Description**
The index of pre-aggr table is <key, filter_key>. However, in `aggregator`, the dimension key is `key` only, which is not consistent with the index definition.

Two possible solutions:
- add the `filter_key` to the dimension key (problem: cannot easily delete if base table deletes)
- change the index to <key> (problem: there will be multiple entries for the  same <key, ts>. It is not feasible for disk table).

zhanghaohit:
- the pre-aggr table index definition is not consistent with the table access
   Close #2307
- aligned start_ts for range pre-aggr batch

### DIFF
--- a/src/storage/aggregator.cc
+++ b/src/storage/aggregator.cc
@@ -410,8 +410,7 @@ bool Aggregator::FlushAggrBuffer(const std::string& key, const std::string& filt
         return false;
     }
     ::openmldb::api::LogEntry entry;
-    std::string pk = absl::StrCat(key, "|", filter_key);
-    entry.set_pk(pk);
+    entry.set_pk(key);
     entry.set_ts(time);
     entry.set_value(encoded_row);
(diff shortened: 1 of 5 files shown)


dl239 (approved):

src/storage/aggregator.cc
```diff
@@ -157,10 +159,10 @@ bool Aggregator::Update(const std::string& key, const std::string& row, const ui
         int64_t latest_ts = aggr_buffer.ts_end_ + 1;
```
aceforeverd: unused variable

aceforeverd:
one test failed `TabletImplTest.AggregatorRecovery`

zhanghaohit:
> one test failed `TabletImplTest.AggregatorRecovery`

fixed

aceforeverd (approved):

PR #2323, opened 2022-08-09
docs: improved both en and zh version of ddl/create_index

michelle-qinqin:
docs

move the overall description to the front of the doc
added note block format on the note on line 49 of the original doc

### DIFF
--- a/docs/en/reference/sql/ddl/CREATE_INDEX_STATEMENT.md
+++ b/docs/en/reference/sql/ddl/CREATE_INDEX_STATEMENT.md
@@ -48,7 +48,7 @@ CREATE INDEX index2 ON t5 (col2);
 -- SUCCEED
 ```
 ```{note}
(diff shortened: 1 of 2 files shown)


docs/en/reference/sql/ddl/CREATE_INDEX_STATEMENT.md
```diff
@@ -37,18 +40,17 @@ OptionEntry ::=
...
 ## **Example**
 ```SQL
 CREATE INDEX index2 ON t5 (col2);
 -- SUCCEED
 ```
-**Note**: If `OPTIONS` is not specified, the created index cannot be used in the SQL to serving online as deployment.
-
-We can also set ts column as below:
+```{note}
+If `OPTIONS` is not provided, the created index cannot be used in the SQL that will be deployed online, since the index doesn't have TS (timestamp).
```
lumianph: the SQL with the created index cannot be deployed online

lumianph (approved):

PR #2337, opened 2022-08-10
fix: taskmanager start failed

dl239:
`TaskManager` start failed because there is no `logs` dirs when execute startup command


aceforeverd (approved):

tobegit3hub (approved):
LGTM

PR #2366, opened 2022-08-17
feat: add 'AddIndex' and 'DeleteIndex' for DiskTable

### LINKED ISSUE #2018: support AddIndex/DeleteIndex for disk table

Add support of AddIndex/DeleteIndex for `DiskTable`

jiang1997:
close https://github.com/4paradigm/OpenMLDB/issues/2018

# Development
- [x] 1. table.h: Add AddIndex and DeleteIndex as  virtual member functions
- [x] 2. disk_table.cc
  - [ ] implement AddIndex
    - [ ] concurrent?
  - [x] implement DeleteIndex
- [x] 3. tablet_impl.cc
  - [x] enable DiskTable::AddIndex 
  - [x] enable DiskTable::DeleteIndex

 # Test
- [x] 1. disk_table_test.cc
  - [x] add a test for AddIndex
- [x] 2. tablet_impl_test.cc
  - [x] enable test for DiskTable::AddIndex
  - [x]  enable test for DiskTable::DeleteIndex


src/storage/table.h
```diff
@@ -168,6 +168,7 @@ class Table {
 
     bool CheckFieldExist(const std::string& name);
 
+    virtual bool AddIndex(const ::openmldb::common::ColumnKey& column_key) = 0;
```
dl239, line 171: add `override` to `AddIndex` in `MemTable` class

src/storage/disk_table.cc
```diff
@@ -1097,8 +1097,128 @@ void DiskTableRowIterator::SeekToFirst() {
...
+        DEBUGLOG("add cf_name %s. tid %u pid %u", index_def->GetName().c_str(), id_, pid_);
+
+        rocksdb::Status s = db_->CreateColumnFamily(cfo, index_def->GetName(), &handle);
+        if (!s.ok()) {
+            PDLOG(WARNING, "rocksdb failed to create ColumnFamily. tid %u pid %u error %s",
+                  id_, pid_, s.ToString().c_str());
+            return false;
+        }
+
+        cf_hs_.push_back(handle);
```
dl239: `Put` and `Iterator` will read `cf_hs_`.  It need to guarantee the thread safe

src/storage/disk_table.cc
```diff
@@ -1097,8 +1097,128 @@ void DiskTableRowIterator::SeekToFirst() {
...
+        PDLOG(WARNING, "index %s can't delete. tid %u pid %u", idx_name.c_str(), id_, pid_);
+        return false;
+    }
+    auto table_meta = GetTableMeta();
+    auto new_table_meta = std::make_shared<::openmldb::api::TableMeta>(*table_meta);
+    if (index_def->GetId() < (uint32_t)table_meta->column_key_size()) {
+        new_table_meta->mutable_column_key(index_def->GetId())->set_flag(1);
+    }
+    std::atomic_store_explicit(&table_meta_, new_table_meta, std::memory_order_release);
+    index_def->SetStatus(IndexStatus::kWaiting);
```
dl239, line 1296: The data in these index should be delete

src/storage/disk_table.cc
jiang1997: Will there be multiple requests require to add new indexes at the same time?

zhanghaohit: yes. `AddIndex` is a RPC call. It can be executed by multiple threads at the same time.

Can learn from the memory table:
- copy the origin `cf_hs_` and update the copied `cf_hs`
- use `std::atomic_store_explicit` to set the `cf_hs_` atomically

src/storage/disk_table.cc
```diff
@@ -1097,8 +1097,128 @@ void DiskTableRowIterator::SeekToFirst() {
...
+
+        uint32_t inner_id = table_index_.GetAllInnerIndex()->size();
+
+        index_def = std::make_shared<IndexDef>(column_key.index_name(), table_index_.GetMaxIndexId() + 1,
+                IndexStatus::kReady, ::openmldb::type::IndexType::kTimeSerise, col_vec);
+
+
+        if (!column_key.ts_name().empty()) {
+            auto ts_iter = schema.find(column_key.ts_name());
+            index_def->SetTsColumn(std::make_shared<ColumnDef>(ts_iter->second));
```
zhanghaohit, line 1213: if `ts_iter == schema.end()`, log error and return `false`

src/storage/disk_table.cc
```diff
@@ -1097,8 +1097,128 @@ void DiskTableRowIterator::SeekToFirst() {
...
+        for (const auto& col_name : column_key.col_name()) {
+            auto it = schema.find(col_name);
+            if (it == schema.end()) {
+                PDLOG(WARNING, "not found col_name[%s]. tid %u pid %u", col_name.c_str(), id_, pid_);
+                return false;
+            }
+            col_vec.push_back(it->second);
+        }
+
+        uint32_t inner_id = table_index_.GetAllInnerIndex()->size();
```
zhanghaohit, line 1201: @dl239 do we need to check if there is any existing inner index whose `column_key` is the same as the new index? If so, we can merge it with one `InnerIndex`.

src/storage/disk_table.cc
zhanghaohit: @jiang1997 in mem table, it is deleted in `SchedGc`

zhanghaohit (changes requested):

src/storage/disk_table.cc
dl239: In memory table, index in the same `InnerIndex` is in the same segment. If there are multi ts columns in one segment, the value of node in  first level `skiplist` is an array.  we cannot allocate more memory for reserved pos. This is different with `disktable`.  I think we can merge it in `disktable` if there are benefits in the same column family.

jiang1997: Got it. I will deal with it.

src/storage/disk_table.cc
jiang1997: Since calls from multiple threads may get the same ``inner_id``, both ``mem_table`` and ``disk_table``  may suffer from concurrent calls to ``Addindex``.
https://github.com/4paradigm/OpenMLDB/blob/f50df38b238fa96342d634ac327f7897443587c9/src/storage/mem_table.cc#L591
https://github.com/4paradigm/OpenMLDB/blob/a6e1a8d7d3f558b50ba538616c4c56b6c815b246/src/storage/disk_table.cc#L1191

jiang1997:
There are multiple functions that either change the element of cf_hs_ or cause cf_hs_to to reallocate. For thread safety, I simply use std::mutex to prevent accessing cf_hs_ simultaneously.
Is there any more canonical way to do so, please?

src/storage/disk_table.cc
zhanghaohit: Yes. I think it is a potential thread-safe issue for memory table too. @dl239 please help verify?

src/storage/disk_table.cc
```diff
@@ -1097,8 +1158,138 @@ void DiskTableRowIterator::SeekToFirst() {
...
+        if (!s.ok()) {
+            PDLOG(WARNING, "rocksdb failed to create ColumnFamily. tid %u pid %u error %s",
+                  id_, pid_, s.ToString().c_str());
+            return false;
+        }
+
+        // may change cf_hs_->data
+        {
+            std::lock_guard<std::mutex> guard(cf_hs_mutex);
+            auto new_cf_hs_ = std::make_shared<std::vector<rocksdb::ColumnFamilyHandle*>>(GetCF_DS_());
```
zhanghaohit: if use `mutex lock`, then no need to copy and reset atomically.

src/storage/disk_table.cc
```diff
@@ -1097,8 +1158,138 @@ void DiskTableRowIterator::SeekToFirst() {
...
+        cfo.comparator = &cmp_;
+        cfo.prefix_extractor.reset(new KeyTsPrefixTransform());
+
+        if (index_def->GetTTLType() == ::openmldb::storage::TTLType::kAbsoluteTime ||
+            index_def->GetTTLType() == ::openmldb::storage::TTLType::kAbsOrLat) {
+            cfo.compaction_filter_factory = std::make_shared<AbsoluteTTLFilterFactory>(inner_index_st);
+        }
+
+        rocksdb::ColumnFamilyHandle* handle;
+        cf_ds_.push_back(rocksdb::ColumnFamilyDescriptor(index_def->GetName(), cfo));
```
zhanghaohit: this one also need the `lock` to protect?

src/storage/disk_table.h
```diff
@@ -409,27 +409,33 @@ class DiskTable : public Table {
...
 
+    bool AddIndex(const ::openmldb::common::ColumnKey& column_key) override;
     bool DeleteIndex(const std::string& idx_name) override;
     uint64_t GetRecordIdxCnt() override;
     bool GetRecordIdxCnt(uint32_t idx, uint64_t** stat, uint32_t* size) override;
     uint64_t GetRecordPkCnt() override;
     uint64_t GetRecordByteSize() const override { return 0; }
     uint64_t GetRecordIdxByteSize() override;
 
+    std::vector<rocksdb::ColumnFamilyHandle*> GetCF_DS_() {
```
zhanghaohit: naming convention: `GetCfDs()`.

here it is more like `CloneCf`?

src/storage/disk_table.h
```diff
@@ -409,27 +409,33 @@ class DiskTable : public Table {
...
+
     int GetCount(uint32_t index, const std::string& pk, uint64_t& count) override; // NOLINT
 
  private:
     rocksdb::DB* db_;
     rocksdb::WriteOptions write_opts_;
     std::vector<rocksdb::ColumnFamilyDescriptor> cf_ds_;
-    std::vector<rocksdb::ColumnFamilyHandle*> cf_hs_;
+    std::shared_ptr<std::vector<rocksdb::ColumnFamilyHandle*>> cf_hs_;
+    std::mutex cf_hs_mutex;
```
zhanghaohit: for `private` member, it is named with suffix `_`, e.g., `cf_hs_mutex_`

can use the same `mutex` for both `cf_hs_` and `cf_ds_`

zhanghaohit (changes requested):

src/storage/disk_table.cc
jiang1997: As I comment here, `push_back` may trigger reallocation and modify `cf_hs_->data`. It may cause errors when functions like `NewWindowIterator` and `NewTraverseIterator` accessing elements of `cf_hs_`  by 'old' `cf_hs_->data`. So atomical copy and reset are needed.


> Data races
The container is modified.
If a reallocation happens, all contained elements are modified.
Otherwise, no existing element is accessed, and concurrently accessing or modifying them is safe.

https://cplusplus.com/reference/vector/vector/push_back/

src/storage/disk_table.h
jiang1997: Yes, it looks more reasonable.

jiang1997: My fault. Here should have be `GetCF_HS_` and actually, `GetCF_DS_` returns a copy of `cf_hs`.

src/storage/disk_table.cc
jiang1997: Because except for some initialization functions, only here appears `cf_ds_`. If we don't consider concurrent calls to `Addindex`, it looks fine not to protect `cf_ds_`.
If we have to, yes, definitely and more than so.

src/storage/disk_table.h
jiang1997: Get it.

src/storage/disk_table.cc
zhanghaohit: oh. Yes. You're right. 

In this case, you may need to atomically load the `cf_hs_` in other `Reading` cases.

src/storage/disk_table.cc
```diff
@@ -1135,7 +1326,7 @@ int DiskTable::GetCount(uint32_t index, const std::string& pk, uint64_t& count)
     ro.snapshot = snapshot;
     // ro.prefix_same_as_start = true;
     ro.pin_data = true;
-    rocksdb::Iterator* it = db_->NewIterator(ro, cf_hs_[inner_pos + 1]);
+    rocksdb::Iterator* it = db_->NewIterator(ro, cf_hs_->at(inner_pos + 1));
```
zhanghaohit: like here. `cf_hs = std::atomic_load_explicit(&cf_hs_, std::memory_order_acquire)`. And then use `cf_hs->at`

zhanghaohit (changes requested):

src/storage/disk_table.cc
jiang1997: > Yes, after `atomic_store_explicit`, previous `cf_hs_` is destoried.

Yes, after atomic_store_explicit, previous cf_hs_ is destroyed.
Thanks for reminding.

jiang1997: > oh. Yes. You're right.
> 
> In this case, you may need to atomically load the `cf_hs_` in other `Reading` cases.

Yes, previous `cf_hs_` is destroyed after `atomic_store_explicit` is done.
Thanks for reminding.

src/storage/disk_table.cc
zhanghaohit: `Addindex` may be called concurrently. We can use same lock to protect it.

src/storage/disk_table.cc
```diff
@@ -191,7 +193,8 @@ bool DiskTable::Init() {
     options_.create_if_missing = true;
     options_.error_if_exists = false;
     options_.create_missing_column_families = true;
-    rocksdb::Status s = rocksdb::DB::Open(options_, path, cf_ds_, &cf_hs_, &db_);
+    // may change cf_hs_-.data
```
zhanghaohit: `cf_hs_->data`?

src/storage/disk_table.cc
zhanghaohit: For disk table, we keep the same design as memory table. That is, do not merge the new index with same `column_key` with other existing ones. If doing the merge, we have to update all the existing `key-value` pairs in rocksdb as the `key` format may change from `pk+ts` to `pk+ts+ts_col`.

jiang1997: > For disk table, we keep the same design as memory table. That is, do not merge the new index with same `column_key` with other existing ones. If doing the merge, we have to update all the existing `key-value` pairs in rocksdb as the `key` format may change from `pk+ts` to `pk+ts+ts_col`.

Thank you for the reminder. 
I find the code that may be causing this error.
https://github.com/4paradigm/OpenMLDB/blob/fee7fa6dfd36dbe60c6868b74faadabc179e1062/src/storage/disk_table.cc#L255-L259

src/storage/disk_table.cc
```diff
@@ -206,7 +208,8 @@ bool DiskTable::Put(const std::string& pk, uint64_t time, const char* data, uint
     rocksdb::Status s;
     std::string combine_key = CombineKeyTs(pk, time);
     rocksdb::Slice spk = rocksdb::Slice(combine_key);
-    s = db_->Put(write_opts_, cf_hs_[1], spk, rocksdb::Slice(data, size));
+    auto cf_hs(cf_hs_);
```
zhanghaohit: I think the constructor `shared_ptr( const shared_ptr& r ) noexcept` is not thread-safe? Under the hood, a shared pointer is a pair of two pointers, one to the object and one to a control block (holding the ref counter).

If you assign a new object to a std::shared_ptr while another thread uses it, it might end up with the new object pointer but still using a pointer to the control block of the old object

src/storage/disk_table.cc
```diff
@@ -329,16 +337,69 @@ void DiskTable::SchedGc() {
...
+                        db_->DestroyColumnFamilyHandle(cf_hs->at(idx + 1));
+                        db_->CreateColumnFamily(cfo, cur_index->GetName(), &handle);
+                        // modify element of cf_hs_
+                        {
+                            std::lock_guard<std::mutex> guard(cf_hs_mutex_);
+                            cf_hs_->at(idx + 1) = handle;
+                        }
+                        deleted_num++;
+                    } else {
+                        // do nothing following mem_table
```
zhanghaohit, line 378: when are the data of these indexes deleted after deleting of all the indexes?

src/storage/disk_table.cc
```diff
@@ -1103,8 +1170,138 @@ void DiskTableRowIterator::SeekToFirst() {
...
+            PDLOG(WARNING, "rocksdb failed to create ColumnFamily. tid %u pid %u error %s",
+                  id_, pid_, s.ToString().c_str());
+            return false;
+        }
+
+        {
+            std::lock_guard<std::mutex> guard(cf_hs_mutex_);
+
+            cf_ds_.push_back(rocksdb::ColumnFamilyDescriptor(index_def->GetName(), cfo));
+            auto new_cf_hs_ = std::make_shared<std::vector<rocksdb::ColumnFamilyHandle*>>(CloneCf());
```
zhanghaohit: atomic load the old `cf_hs_`

zhanghaohit (changes requested):

zhanghaohit:
rebase or merge from `main` branch

src/storage/disk_table.cc
jiang1997: As we use `std::atomic_store_explicit` to update `cf_hs_`, we don't need to consider the data race like this.
`cf_hs` is either constructed from the old `cf_hs_` or the new `cf_hs_` and both will work as expected.

src/storage/disk_table.cc
jiang1997: See here https://github.com/4paradigm/OpenMLDB/blob/35a418eccdaa6021bf2e4b808ce737f0b0e1984a/src/storage/disk_table.cc#L375-L377
After dropping the column family, the data inside it will be deleted by `rockdb` at some point. Say [here](http://mysql.taobao.org/monthly/2018/06/09/) the "some point" is when the reference count to the column family decreases to 0.

https://github.com/facebook/rocksdb/issues/1295
https://github.com/facebook/rocksdb/issues/5006

src/storage/disk_table.cc
jiang1997: As I have mentioned, we use `std::atomic_store_explicit` to update `cf_hs_`. I think it's fine to do so.

src/storage/disk_table.cc
zhanghaohit: No. It's not like this. `std::atomic_store_explicit` must be used tother with `std::atomic_load_explicit` to guarantee the `atomicity`. If it is not true that you can use normal load if you use `std::atomic_store_explicit`.

See the note from cppreference:
`
To avoid data races, once a shared pointer is passed to any of these functions, it cannot be accessed non-atomically. In particular, you cannot dereference such a shared_ptr without first atomically loading it into another shared_ptr object, and then dereferencing through the second object.
`

https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic

src/storage/disk_table.cc
zhanghaohit: No. Here I mean when the condition `if (indexs.size() == 1)` will be true. I didn't find any code you remove or change the size of inner index.

src/storage/disk_table.cc
jiang1997: Got it. My bad.
Thanks for the explanation

src/storage/disk_table.cc
jiang1997: If I get this right, `MemTable` do not delete the `inner_index` either and reserve it for future reuse.

https://github.com/4paradigm/OpenMLDB/blob/35a418eccdaa6021bf2e4b808ce737f0b0e1984a/src/storage/disk_table.cc#L1177-L1184

jiang1997: If I get this right, `MemTable` do not delete elements of the `inner_index` either and keep it for future reuse.

https://github.com/4paradigm/OpenMLDB/blob/35a418eccdaa6021bf2e4b808ce737f0b0e1984a/src/storage/mem_table.cc#L556-L563

zhanghaohit: Yes. `MemTable` has a bug for this case #2680

There is a PR to fix the bug: #2723, but still not complete. 

For `DiskTable`, it is not easy to only delete the data in one single index if the size of its parent inner index > 1. Maybe you can put an issue to take note of this bug. We can fix it in another PR.

src/storage/disk_table.h
```diff
@@ -409,27 +409,37 @@ class DiskTable : public Table {
...
     uint64_t GetRecordIdxByteSize() override;
 
+    std::shared_ptr<std::vector<rocksdb::ColumnFamilyHandle*>> GetCf() {
+        return std::atomic_load_explicit(&cf_hs_, std::memory_order_relaxed);
+    }
+
+    std::shared_ptr<std::vector<rocksdb::ColumnFamilyHandle*>> CloneCf() {
+        return std::make_shared<std::vector<rocksdb::ColumnFamilyHandle*>>(*(GetCf()));
+    }
+
```
zhanghaohit: make these above two methods as `private`

src/storage/disk_table.cc
jiang1997: Got it.

zhanghaohit: #2748

src/storage/disk_table.h
```diff
@@ -380,10 +380,10 @@ class DiskTable : public Table {
 
     uint64_t GetRecordCnt() const override {
         uint64_t count = 0;
-        if (cf_hs_.size() == 1) {
-            db_->GetIntProperty(cf_hs_[0], "rocksdb.estimate-num-keys", &count);
+        if (cf_hs_->size() == 1) {
```
dl239: get cf by `GetCf`

src/storage/disk_table.cc
```diff
@@ -324,16 +329,69 @@ void DiskTable::SchedGc() {
...
+                        }
+
+                        rocksdb::ColumnFamilyHandle* handle;
+                        db_->DropColumnFamily(cf_hs->at(idx + 1));
+                        db_->DestroyColumnFamilyHandle(cf_hs->at(idx + 1));
+                        db_->CreateColumnFamily(cfo, cur_index->GetName(), &handle);
+                        // modify element of cf_hs_
+                        {
+                            std::lock_guard<std::mutex> guard(cf_hs_mutex_);
+                            cf_hs_->at(idx + 1) = handle;
```
dl239, line 374: use `atomic_store_explicit` and remove the `cf_hs_mutex_`

src/storage/disk_table.h
```diff
@@ -409,13 +409,14 @@ class DiskTable : public Table {
     bool IsExpire(const ::openmldb::api::LogEntry& entry) override;
 
     void CompactDB() {
-        for (rocksdb::ColumnFamilyHandle* cf : cf_hs_) {
+        for (rocksdb::ColumnFamilyHandle* cf : *cf_hs_) {
```
jiang1997: @dl239 
Do we need to use ``GetCf`` here?

dl239 (approved):

PR #2380, opened 2022-08-20
test: cicd add mail

wuyou10206:
Add email notification function when integration test fails

No notice

### DIFF
--- a/.github/workflows/integration-test-src.yml
+++ b/.github/workflows/integration-test-src.yml
@@ -143,8 +143,8 @@ jobs:
           server_port: 587
           username: ${{ secrets.MAIL_USERNAME }}
           password: ${{ secrets.MAIL_PASSWORD }}
-          subject: OpenMLDB Memory Test
-          body: OpenMLDB Memory Test Failed
+          subject: OpenMLDB Memory 1 Test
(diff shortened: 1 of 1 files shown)


vagetablechicken (approved):

.github/workflows/integration-test-src.yml
```diff
@@ -89,6 +89,23 @@ jobs:
           comment_mode: "create new"
           check_name: "SRC java-sdk-cluster-memory-0 Report"
           comment_title: "SRC java-sdk-cluster-memory-0 Report"
+      - name: tar test report
+        if: ${{ failure() }}
+        run: tar -zcvf allure-results.tar.gz test/integration-test/openmldb-test-java/openmldb-sdk-test/target/allure-results
+      - name: Send Email
```
aceforeverd, line 95: GItHub will send fail email, config via https://github.com/settings/notifications ?

aceforeverd (approved):

.github/workflows/integration-test-src.yml
wuyou10206: This can't send the desired attachment, and can't send the failed mail according to the job

PR #2431, opened 2022-09-03
feat: add the implement of unhex build in function #822

### LINKED ISSUE #822: built-in function: Support `unhex()` function

### unhex(expr)
Converts hexadecimal expr to binary.

### Examples:
```sql
> SELECT decode(unhex('537061726B2053514C'), 'UTF-8');
 Spark SQL
```
### reference:
https://spark.apache.org/docs/2.3.0/api/sql/#unhex

We do provide a built-in function development guide for you. 
Please check [OpenMLDB Build-In Function Develop Guide](https://github.com/4paradigm/OpenMLDB/blob/main/docs/en/built_in_function_develop_guide.md) or the [Chinese version](https://github.com/4paradigm/OpenMLDB/blob/main/docs/cn/built_in_function_develop_guide.md) for help.

Please make sure you:
- [ ] Add built-in C++ function in src/udf/udf.h and src/udf/udf.cc
- [ ] Register function to default library from function `void DefaultUdfLibrary::IniStringUdf()` in src/udf/default_udf_library.cc.
- [ ] Add related unit test in src/codegen/udf_ir_builder_test.cc
- [ ] Documenting function 

If you have any questions or concerns, please let us know. We appreciate every piece of code, comment.

**Additional context**

aucker:

  feature: add the implement of `unhex()` function. #822 

  Now in SQL queries, you can use the feature like following:
  ```shell
  select unhex("4F70656E4D4C4442");
  OpenMLDB
  ```
  ```shell
  select unhex(hex("OpenMLDB"));
  OpenMLDB
  ```
  When your input contains characters that are not even number, it will resolve the first character, then resolver the leftover.
  ```shell
  select unhex("4");
  

  select unhex("7B");
  {
  
  select unhex("47B");
  {
  ```
  And when your input contains the invalid characters, you will get a `NULL` result like mysql and spark sql:
  ```shell
  select unhex("zfk");
  NULL
  ```


hybridse/src/udf/udf.cc
```diff
@@ -64,6 +64,28 @@ void hex(StringRef *str, StringRef *output) {
     output->data_ = buffer;
 }
 
+void unhex(StringRef *str, StringRef *output) {
+    std::ostringstream ss;
+    uint32_t arr[str->size_];
```
aceforeverd: resolve the cpplint warnings

hybridse/src/udf/udf.cc
```diff
@@ -64,6 +64,28 @@ void hex(StringRef *str, StringRef *output) {
     output->data_ = buffer;
 }
 
+void unhex(StringRef *str, StringRef *output) {
```
aceforeverd: consider invalid inputs:
- string input contains character other than numerics
 `null` should returned.

If input string size % 2 != 0. How to handle it ? 
e.g. input is `667`, whether you should take `66` -> `f` or take `67` -> `g` ? 
I personally not sure but you can have a reference to spark sql or mysql :)

hybridse/src/udf/udf.cc
```diff
@@ -64,6 +64,28 @@ void hex(StringRef *str, StringRef *output) {
...
+        if (str->data_[i] <= 'F' && str->data_[i] >= 'A') {
+            tmp = str->data_[i] - 'A' + 10;
+        } else if (str->data_[i] <= 'z' && str->data_[i] >= 'a') {
+            tmp = str->data_[i] - 'a' + 10;
+        } else if (str->data_[i] <= '9' && str->data_[i] >= '0') {
+            tmp = str->data_[i] - '0';
+        }
+        arr[i] = tmp;
+    }
+    for (uint32_t i=0; i < str->size_; i+=2) {
```
aceforeverd: two level inner array is used, which can be avoided. 

I think you can take first two characters from input and parse it as hex number. Then write into output directly, since output size is known from input size.

aceforeverd (changes requested):

hybridse/src/udf/udf.cc
aucker: will fix that soon, thanks for the review :)

hybridse/src/udf/udf.cc
aucker: just resolved the cpplint warnings <3

hybridse/src/udf/udf.cc
aucker: Hi, @aceforeverd. I just update my code and add the error handler for the 'null' case and the size is not even case, now my tests all passed, but my code structure is not as good as you wish, hope you can take several minutes give it a review, thanks :). And happy mid-autumn festival 🥮

hybridse/src/udf/default_udf_library.cc
```diff
@@ -585,6 +585,26 @@ void DefaultUdfLibrary::InitStringUdf() {
...
+                select unhex("537061726B2053514C");
+                --output "Spark SQL"
+
+                select unhex("zfk");
+                --output "NULL"
+
+                select unhex("7B");
+                --output "{"
+            @endcode
+            @since 0.6.0)");
```
aceforeverd: so this should be 0.7.0

hybridse/src/udf/default_udf_library.cc
```diff
@@ -585,6 +585,26 @@ void DefaultUdfLibrary::InitStringUdf() {
             @endcode
             @since 0.6.0)");
 
+    RegisterExternal("unhex")
+        .args<StringRef>(static_cast<void (*)(StringRef*, StringRef*)>(udf::v1::unhex))
+        .return_by_arg(true)
```
aceforeverd, line 590: so it should be `false` as above

hybridse/src/udf/default_udf_library.cc
```diff
@@ -585,6 +585,26 @@ void DefaultUdfLibrary::InitStringUdf() {
             @endcode
             @since 0.6.0)");
 
+    RegisterExternal("unhex")
+        .args<StringRef>(static_cast<void (*)(StringRef*, StringRef*)>(udf::v1::unhex))
```
aceforeverd: unhex might return `NULL` so the signature should be `StringRef*, StringRef*, bool*`, where the third parameter is whether output is `NULL`. You can learn how to register it through other functions, like dayofyear https://github.com/4paradigm/OpenMLDB/blob/75e3ec231009c55a8bd4c00ac79efa106df0324d/hybridse/src/udf/default_udf_library.cc#L2180. 

`NULL` is a special value which is different from "NULL" string literal :)

hybridse/src/udf/udf.cc
aceforeverd: no worry, happy coding

hybridse/src/udf/udf.cc
```diff
@@ -64,6 +64,45 @@ void hex(StringRef *str, StringRef *output) {
...
+                ss << char(arr[i] << 4 | arr[i+1]);
+            }
+        }
+        output->size_ = ss.str().size();
+        char *buffer = AllocManagedStringBuf(output->size_);
+        memcpy(buffer, ss.str().data(), output->size_);
+        output->data_ = buffer;
+    } else {
+        output->size_ = 4;
+        output->data_ = "NULL";
```
aceforeverd: it is the `NULL` value as stated above

hybridse/src/udf/udf.cc
```diff
@@ -64,6 +64,45 @@ void hex(StringRef *str, StringRef *output) {
...
+        } else if (str->data_[i] <= 'f' && str->data_[i] >= 'a') {
+            arr[i] = str->data_[i] - 'a' + 10;
+        } else if (str->data_[i] <= '9' && str->data_[i] >= '0') {
+            arr[i] = str->data_[i] - '0';
+        } else {
+            flag = 1;
+            break;
+        }
+    }
+    if (flag == 0) {    // every character is valid hex character
```
aceforeverd: here is the behavior of spark sql:
```
spark-sql> select unhex(667), char_length(unhex(667)), hex(unhex(667));
g	2	6700
Time taken: 0.049 seconds, Fetched 1 row(s)
```
strange behavior but it seem interrupt it as `67`, `00`. But I think our behavior is fine too

hybridse/src/udf/udf.cc
```diff
@@ -64,6 +64,45 @@ void hex(StringRef *str, StringRef *output) {
     output->data_ = buffer;
 }
 
+void unhex(StringRef *str, StringRef *output) {
+    std::ostringstream ss;
+    uint8_t* arr = new uint8_t[str->size_];
```
aceforeverd: since there will be one more function parameter: `is_null`, it's fine to alloc buffer at very begining, then just update the allocated string content in place.

hybridse/src/udf/udf.cc
aceforeverd: also to metion, mysql has more similar behavior to us:
```
MariaDB [(none)]> select char_length(unhex(667)), unhex(667), hex(unhex(667));
+-------------------------+------------+-----------------+
| char_length(unhex(667)) | unhex(667) | hex(unhex(667)) |
+-------------------------+------------+-----------------+
|                       2 | g         | 0667            |
+-------------------------+------------+-----------------+
```

aucker: yeah, I just tried in MySQL 5.7, looks like it unhex in ASCII table, The char of hex '6' is [ACKNOWLEDGE], so in terminal, it returns a character.
image

```shell
mysql> select unhex(667), char_length(unhex(667)), hex(unhex(667));
+------------+-------------------------+-----------------+
| unhex(667) | char_length(unhex(667)) | hex(unhex(667)) |
+------------+-------------------------+-----------------+
| g         |                       2 | 0667            |
+------------+-------------------------+-----------------+
1 row in set (0.00 sec)
```

hybridse/src/udf/default_udf_library.cc
aucker: ok, just updated to 0.7.0

hybridse/src/udf/default_udf_library.cc
aucker: added the `bool* is_null` parameter

hybridse/src/udf/default_udf_library.cc
aucker: emm, I don't quite understand, according to the `dayofyear` function, I think it should be `true`

hybridse/src/udf/udf.cc
aucker: according to spark sql, when I input the invalid characters, it will return the following:
```shell
spark-sql> select unhex("zzz");
NULL
Time taken: 2.967 seconds, Fetched 1 row(s)
spark-sql> select char_length(unhex("zzz"));
NULL
Time taken: 0.087 seconds, Fetched 1 row(s)
```
So in my code, I write the following:
```cpp
output->size_ = NULL;
output->data_ = NULL;
```
in `udf_ir_builder_test.cc` file, 
```cpp
CheckUdf<StringRef, StringRef>("unhex", nullptr, StringRef("Zzzz"));
```
And I can pass all the unit tests.
Is that all right?

aceforeverd: in this case (invalid input), set the third parameter (is_bool) to true is the only step, resulting the output as `NULL`. Generally the `output` object here is not accessed afterwards.

hybridse/src/udf/udf.cc
```diff
@@ -64,6 +64,44 @@ void hex(StringRef *str, StringRef *output) {
...
+            for (uint32_t i=1; i < str->size_; i+=2) {
+                ss << char(arr[i] << 4 | arr[i+1]);
+            }
+        }
+        output->size_ = ss.str().size();
+        char *buffer = AllocManagedStringBuf(output->size_);
+        memcpy(buffer, ss.str().data(), output->size_);
+        output->data_ = buffer;
+    } else {
+        output->size_ = NULL;
```
aceforeverd: rm the else branch

hybridse/src/udf/udf.cc
aceforeverd: here is a simplified steps:
1. alloc the managed buffer with size `origin_size / 2 + origin_size % 2` and memset
2. iterate through the input and 
   - any invalid character will set `is_null` to true and break
   - update the pos of buffer with value `arr[i] << 4 | arr[i+1]`
3. write `output` if not null

aucker: `buffer[i/2] = char(arr[i] << 4 | arr[i+1]);`, it works!
I thought there is no way to update the corresponde position of buffer but using `stringstream`, it's amazing, thanks for the kindly hint.:)

But there is one weird question: 
```shell
[ RUN      ] UdfIRBuilderTest.UnhexTest
/home/aucker/mldb/OpenMLDB/hybridse/src/udf/udf_test.h:172: Failure
Expected equality of these values:
  expect
    Which is: 7&�5
  output
    Which is: 7&�5
[  FAILED  ] UdfIRBuilderTest.UnhexTest (119 ms)
```
after I update the code, I test it again, all tests passed except this one, but in the log, they are all same, I can't see the difference between of `expect` and the `output`.
And the corresponde test is this one, I just remove the last `C`:
```shell
CheckUdf<StringRef, StringRef>("unhex", "7&�5", StringRef("537061726B2053514"));
```

hybridse/src/codegen/udf_ir_builder_test.cc
```diff
@@ -123,6 +123,32 @@ TEST_F(UdfIRBuilderTest, HexStringUdfTest) {
...
+    CheckUdf<StringRef, StringRef>("unhex", "Spark SQLSpark", StringRef("537061726B2053514C537061726B"));
+    CheckUdf<StringRef, StringRef>("unhex", "OpenMLDB", StringRef("4F70656E4D4C4442"));
+    CheckUdf<StringRef, StringRef>("unhex", "OpenMLDB", StringRef("4f70656e4d4c4442"));
+    // The following are valid character but not string unhex tests and the length of
+    // some tests cases are odd.
+    CheckUdf<StringRef, StringRef>("unhex", "", StringRef("4"));
+    CheckUdf<StringRef, StringRef>("unhex", "{", StringRef("7B"));
+    CheckUdf<StringRef, StringRef>("unhex", "{", StringRef("47B"));
+    CheckUdf<StringRef, StringRef>("unhex", "7&", StringRef("537061726"));
+    CheckUdf<StringRef, StringRef>("unhex", "7&�5", StringRef("537061726B2053514"));
```
aucker: Here is the only test fails

hybridse/src/udf/default_udf_library.cc
aceforeverd: My mistake

hybridse/src/udf/udf.cc
aceforeverd: `7` is `0x55`, `%` is `0x37`

aceforeverd: `b2` is wrong

hybridse/src/udf/udf.cc
```diff
@@ -64,6 +64,39 @@ void hex(StringRef *str, StringRef *output) {
...
+            arr[i] = str->data_[i] - '0';
+        } else {
+            *is_null = true;
+            break;
+        }
+    }
+    if (!*is_null) {    // every character is valid hex character
+        if (str->size_ % 2 == 0) {
+            for (uint32_t i=0; i < str->size_; i+=2) {
+                buffer[i/2] = static_cast<char>(arr[i] << 4 | arr[i+1]);
```
aceforeverd: rm the `arr` list and just access `str->data_`

hybridse/src/codegen/udf_ir_builder_test.cc
zhanghaohit: can use the `// NOLINT` added at the end of line to avoid the reviewdog warning

hybridse/src/udf/udf.cc
aucker: Should I add another case to handle this error? I copy the unhex result from MySQL 5.7 as the `expect` result in test, so I guess when handle the hex result like `0xa5`, `0xd1` etc. Though each character is valid, but they are not in ASCII table, it will give the same error. Not so sure, I will have a try at 8am, my computer is not around.

hybridse/src/codegen/udf_ir_builder_test.cc
aucker: Will fix that, thanks for the review:)

hybridse/src/udf/udf.cc
aucker: I use the `arr` list to store the decimal result of each character, and `(arr[i] << 4 | arr[i+1])` works when it's decimal, will it work when it's hex?

hybridse/src/udf/default_udf_library.cc
```diff
@@ -585,6 +585,27 @@ void DefaultUdfLibrary::InitStringUdf() {
...
+
+            @code{.sql}
+                select unhex("537061726B2053514C");
+                --output "Spark SQL"
+
+                select unhex("7B");
+                --output "{"
+
+                select unhex("zfk");
+                --output "NULL"
```
zhanghaohit: --output `NULL` rather than "NULL" string?

hybridse/src/udf/udf.cc
aucker: I tried in mysql and sparksql, for hex that are not in ascii table, it all returns the `�` character.
```shell
mysql> select unhex("80");
+-------------+
| unhex("80") |
+-------------+
| �           |
+-------------+
1 row in set (0.00 sec)

mysql> select unhex("88");
+-------------+
| unhex("88") |
+-------------+
| �           |
+-------------+
1 row in set (0.00 sec)
```
In sparksql: 
```shell
spark-sql> select unhex("80");
�
Time taken: 0.057 seconds, Fetched 1 row(s)
spark-sql> select unhex("81");
�
Time taken: 0.051 seconds, Fetched 1 row(s)
```

Github don't render the special characters, actually my test is 
image, so just `B2` is not valid character, `B2` is like `80`, `81` etc. When hex value is greater than `7f`, it all return `�`.
I add two tests,
```shell
CheckUdf<StringRef, StringRef>("unhex", "�", StringRef("a2")); 
CheckUdf<StringRef, StringRef>("unhex", "�", StringRef("8a"));
```
and here is the log:
```shell
/home/aucker/mldb/OpenMLDB/hybridse/src/udf/udf_test.h:172: Failure
Expected equality of these values:
  expect
    Which is: �
  output
    Which is: �
/home/aucker/mldb/OpenMLDB/hybridse/src/udf/udf_test.h:172: Failure
Expected equality of these values:
  expect
    Which is: �
  output
    Which is: �
```

So I think now that this function behavior is like mysql and spark(return `�` when hex greater than `7F`),  and I cannot test which character is equal to the hex `�`, can I just remove this test?

hybridse/src/udf/udf.cc
aceforeverd: no it won't work. But you can them into a single loop, as the logic `2char` -> `2 u8int ` -> `char`. Then arr is not needed

hybridse/src/udf/udf.cc
aceforeverd: do not copy the character with your system clipboard. chars from 128 - 255 is what called the [extended ascii characters](https://www.ascii-code.com/). It has serval standards which result in different chars. You get a different char possibly because of the encoding setting of the terminal. 

So I think just write string result as `"\x80"` in tests is better. ?

aceforeverd (approved):

hybridse/src/udf/udf.cc
```diff
@@ -64,6 +64,38 @@ void hex(StringRef *str, StringRef *output) {
     output->data_ = buffer;
 }
 
+void unhex(StringRef *str, StringRef *output, bool* is_null) {
+    char *buffer = AllocManagedStringBuf(str->size_ / 2 + str->size_ % 2);
+    for (uint32_t i = 0; i < str->size_; ++i) {
+        if ((str->data_[i] >= 'A' && str->data_[i] <= 'F') || (str->data_[i] >= 'a' && str->data_[i] <= 'f') || (str->data_[i] >= '0' && str->data_[i] <= '9')) {
```
zhanghaohit: lint issue

zhanghaohit (changes requested):
some lint issue

zhanghaohit (approved):

PR #2435, opened 2022-09-05
fix: ignore `interactive` if `cmd` is specified

### LINKED ISSUE #2429: Ns_client stuck in infinite loop while Interactive set false
**Bug Description**
while using ns_client or sql_client to execute a  command with Interactive set as true(default), it will stuck in an infinite loop

**Expected Behavior**
cmd result
end

**Steps to Reproduce**
```
/openmldb --role=ns_client  --cmd show components
```
it will cause an infinite loop
```
/openmldb --role=ns_client --interactive=false --cmd show components
```
it works well

dl239:
fix #2429

### DIFF
--- a/src/cmd/sql_cmd.h
+++ b/src/cmd/sql_cmd.h
@@ -139,6 +139,9 @@ void HandleSQL(const std::string& sql) {
 // cluster mode: if zk_cluster is not empty,
 // standalone mode:
 void Shell() {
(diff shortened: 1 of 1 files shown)


vagetablechicken (approved):

aceforeverd:
@wuyou10206 @lqy222 can you test if problem has resolved ?

aceforeverd (approved):

PR #2503, opened 2022-09-20
docs: update openmldb sql capabilities 

### LINKED ISSUE #2552: docs: improve `reference/sql/functions_and_operators/operators.md`
- logic AND: only `AND`, `&&` is not supported
- logic OR: only `OR`, `||` is not supported
- add `RLIKE` keyword
- other rechecks

aceforeverd:
**What's changed**

- limit clause is not supported in online serving mode
- revise SQL operators, close #2552 
- zookeeper constant: version between 3.4 to 3.6


dl239 (approved):

docs/en/reference/sql/dql/LIMIT_CLAUSE.md
```diff
@@ -16,11 +16,12 @@ SELECT ... LIMIT ...
 ```
 
 ## Description
-For the standalone version, `LIMIT` is supported in all conditions. For the cluster version, the execution modes, which support this clause, are shown below.
+
+Limit clause is not supported in online preview mode. By practise, limit clause should used together with order by clause, to constrains the result rows into a unique order. OpenMLDB do not support order by clause yet, it is exepcted to get inconsistent results for the same query.
```
zhanghaohit: not supported in online request mode?

zhanghaohit: `Limit` clause is not supported in online request mode. By practice, `limit` clause should be used together with `order by` clause, to constrain the result rows in a unique order. OpenMLDB does not support `order by `clause yet; without it it may get inconsistent results for the same query.


some typos fixed

zhanghaohit (changes requested):

docs/en/reference/sql/dql/LIMIT_CLAUSE.md
aceforeverd: since request mode always return one row. It is useless for limit appear in the end of sql string. So it is turned off in case confusing. 

Using limit inside a subquery has its meaning for request mode, but we do not support it yet, physical node is not optimized for limit subquery

zhanghaohit: then here is a typo, right? `online preview mode`

aceforeverd: Thx. 

But
- ~~online preview mode is what the request mode called in the docs elsewhere~~
- `to constrains the result rows into a unique order` is picked up from psql: https://www.postgresql.org/docs/15/queries-limit.html. Though should be fine too

zhanghaohit (approved):

PR #2539, opened 2022-09-22
fix: client and ut glog settings

vagetablechicken:
client: role == ""(java/python sdk) or "sql_client", use glog_dir
server: others, use openmldb_log_dir
ut use original log_dir

If get emtpy log dir(all roles), set logtostderr=true


src/base/glog_wrapper_test.cc
```diff
@@ -36,7 +36,7 @@ TEST_F(GlogWrapperTest, UnprotectedSetGlog) {
     ::openmldb::base::SetLogLevel(DEBUG);
     std::string path = "hello";
     FLAGS_openmldb_log_dir = "/tmp/glog_wrapper_test1";
-    FLAGS_role = "tester1";
+    FLAGS_role = "tester1"; // role != empty/sql_client, use openmldb_log_dir
```
zhanghaohit: lint

zhanghaohit (approved):

zhanghaohit:
now the log output is not directed to the screen in the tests? #2619

aceforeverd (approved):

vagetablechicken:
> now the log output is not directed to the screen in the tests? #2619

If tests SetupGlog(true), use FLAGS_log_dir and FLAGS_minloglevel(the original setting of glog), and then if log_dir is empty, we will set `FLAGS_logtostderr = true;`, so it will print to screen. If you don't want, change FLAGS_log_dir.

Other tests(without SetupGlog), may set by the cluster router or server(contains SetupGlog in component init) in tests, it's a lottery. If you want, add `SetupGlog(true)` in test.

Issue #2507, opened 2022-09-20
add sha256sum for release pkg


vagetablechicken:
Users may download incomplete released pkgs(low download speed or interrupt ...), we should add sha256sum for them. Users can check the pkgs easily.

aceforeverd:
ref this job https://github.com/4paradigm/hybridsql-asserts/blob/main/.github/workflows/build.yml#L162-L184

try-agaaain:
Is it to create a shell file to check the integrity of the file?

try-agaaain:
I want to give it a try, please assign this task to me.

PR #2560, opened 2022-09-25
build: add sha256sum for release pkg

try-agaaain:
build: add sha256sum for release pkg [#2507](https://github.com/4paradigm/OpenMLDB/issues/2507)

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -118,6 +118,7 @@ jobs:
         if: ${{ github.event_name == 'push' }}
         run: |
           tar czf ${{ env.OPENMLDB_PREFIX }}.tar.gz ${{ env.OPENMLDB_PREFIX }}/
+          sha256sum ${{ env.OPENMLDB_PREFIX }}.tar.gz > ${{ env.OPENMLDB_PREFIX }}-sha256sum
 
       - name: upload artifacts
         if: ${{ github.event_name == 'push' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/cicd.yaml
```diff
@@ -194,6 +195,7 @@ jobs:
         if: ${{ github.event_name == 'push' }}
         run: |
           tar czf ${{ env.OPENMLDB_PREFIX }}.tar.gz ${{ env.OPENMLDB_PREFIX }}/
+          sha256sum ${{ env.OPENMLDB_PREFIX }}.tar.gz > ${{ env.OPENMLDB_PREFIX }}-sha256sum
```
aceforeverd: pls ref https://github.com/4paradigm/hybridsql-asserts/blob/main/.github/workflows/build.yml#L162-L184. 

changes only needed in `release` job. And a single `SHA256SUM` file is required for extra release

aceforeverd (changes requested):

.github/workflows/cicd.yaml
try-agaaain: Does that mean putting the sha256sum of the two installation packages for Linux and drawin in the same file?

Like this:
```yaml
      - name: generate signature
        run: |
          sha256sum openmldb-*.tar.gz > SHA256SUM

      - name: Release
        if: ${{ startsWith(github.ref, 'refs/tags/v') }}
        uses: softprops/action-gh-release@v1
        with:
          files: |
            openmldb-*.tar.gz
            SHA256SUM
```

aceforeverd: yes!

try-agaaain:
Here's a check failure, how can I fix it?

aceforeverd:
> Here's a check failure, how can I fix it?

Should be fine to ignore

aceforeverd (approved):

vagetablechicken (approved):

PR #2592, opened 2022-09-29
fix: show curl error in start.sh

vagetablechicken:
If curl server status failed, we can't get the error. So show it.


release/bin/start.sh
```diff
@@ -79,7 +79,9 @@ case $OP in
             COUNT=1
             while [ $COUNT -lt 12 ]
             do
-                if ! curl "http://$ENDPOINT/status" > /dev/null 2>&1; then
+                if ! curl "http://$ENDPOINT/status" > /tmp/server_status.log 2>&1; then
```
aceforeverd: `curl -S` ?

vagetablechicken: OK, I'll try it.

zhanghaohit (approved):

tobegit3hub (approved):
LGTM

Issue #2042, opened 2022-06-29
ci: reuse the publish test reports workflow


aceforeverd:
**Describe the feature you'd like**

the cpp, hybridse and sdk test report actually follow the same logic but are written multiple times, like here:

- https://github.com/4paradigm/OpenMLDB/blob/3dfa6ceb0c2ca786258ef35afced9ed874a3131f/.github/workflows/hybridse-ci.yml#L88.

Idealy, they can avoid by reusing the same custom workflow, refer:

- https://docs.github.com/en/actions/using-workflows/reusing-workflows

There are six place where we can reusing

- in `cicd.yaml`, `hybridse-ci.yml` and `sdk.yml` for the `publish-test-reports` jobs
- and the three `publish-*-test-result-from-fork.yml`

**Additional context**

mangoGoForward:
I want to tackle it, please assign to me

lumianph:
> I want to tackle it, please assign to me

thanks~ I have assigned this task to you.

mangoGoForward:
I had some problems when process on this issue. seems the workflow reused is a whole job, but our `publish-test-reports` jobs have a little difference, such as:
```
      - name: Publish Linux UT Results
        uses: EnricoMi/publish-unit-test-result-action@v1
        with:
          files: linux-ut-result-*/**/*.xml
          check_name: Linux Test Report
          comment_title: Linux Test Report
```
in `cicd.yaml`, and 
```
      - name: Publish Linux UT Results
        uses: EnricoMi/publish-unit-test-result-action@v1
        with:
          files: |
            linux-ut-result-*/**/*.xml
            mac-ut-result-*/**/*.xml
          check_name: SDK Test Report
          comment_title: SDK Test Report
```
in `sdk.yml`, the value of files is different. Also the `needs` is different in their own jobs

PR #2601, opened 2022-09-30
ci: reuse workflow for three publish-*-test-result-from-fork.yml 

try-agaaain:
This PR is related to issue: "ci: reuse the publish test reports workflow [#2042](https://github.com/4paradigm/OpenMLDB/issues/2042)", and I created the file `publish-test-result-from-fork-template.yml` to avoid duplicating the code. 

In the file publish-hybridse-test-result-from-fork.yml, I divided the original job `test-results` into two: `test-results` and `public-mac-ut-results`. Does this approach meet the requirements? 

>If this way is reasonable, I will continue to modify publish-test-reports jobs in `cicd.yaml`, `hybridse-ci.yml` and `sdk.yml` to reduce code duplication


.github/workflows/publish-hybridse-test-result-from-fork.yml
```diff
@@ -9,61 +9,21 @@ on:
 
 jobs:
   test-results:
+    uses: ./.github/workflows/publish-test-result-from-fork-template.yml
+    with:
+      name: Publish Linux UT Results
+      check_name: Linux Test Report
+      comment_title: Linux Test Report
+
+  public-mac-ut-results:
```
aceforeverd: I think they can be in the same job as two steps, what's the problem ?

.github/workflows/publish-test-result-from-fork-template.yml
```diff
@@ -0,0 +1,68 @@
...
+        type: string
+      comment_title:
+        required: true
+        type: string
+
+jobs:
+  test-results:
+    name: Test Results
+    runs-on: ubuntu-latest
+    if: >
```
aceforeverd: the if condition should be outside of this workflow definition, instead give the condition from each step calling

.github/workflows/publish-test-result-from-fork-template.yml
```diff
@@ -0,0 +1,68 @@
...
+            then
+              dir="${file/%.zip/}"
+              mkdir -p "$dir"
+              unzip -d "$dir" "$file"
+            fi
+          done
+      - name: ${{ inputs.name }}
+        uses: EnricoMi/publish-unit-test-result-action@v1
+        with:
+          commit: ${{ github.event.workflow_run.head_sha }}
```
aceforeverd: commit and files should be available as input parameter so caller can custom test reports

.github/workflows/publish-test-result-from-fork-template.yml
```diff
@@ -0,0 +1,68 @@
...
+          for file in artifacts/*.zip
+          do
+            if [ -f "$file" ]
+            then
+              dir="${file/%.zip/}"
+              mkdir -p "$dir"
+              unzip -d "$dir" "$file"
+            fi
+          done
+      - name: ${{ inputs.name }}
```
aceforeverd: does name parameter necessary? Or a const name just work ?

aceforeverd (commented):
Nice to see somebody pick this issue up :)

.github/workflows/publish-test-result-from-fork-template.yml
try-agaaain: Thank you for your advice. I revised the code and added a new submission. Does it meet the requirements?

try-agaaain:
Any suggestions would be greatly appreciated.

src/sdk/sql_cluster_router.cc
```diff
@@ -1893,7 +1893,7 @@ std::shared_ptr<hybridse::sdk::ResultSet> SQLClusterRouter::HandleSQLCmd(const h
             exit(0);
         }
         case hybridse::node::kCmdShowJobs: {
-            std::string db = "__INTERNAL_DB";
+            std::string db = openmldb::nameserver::INTERNAL_DB;
```
aceforeverd: try merge/rebase upstream changes, so those code changes won't appear here

.github/workflows/publish-test-result-from-fork.yml
```diff
@@ -16,46 +16,9 @@ jobs:
...
-            fi
-          done
-      - name: Publish Linux UT Results
-        uses: EnricoMi/publish-unit-test-result-action@v1
-        with:
-          commit: ${{ github.event.workflow_run.head_sha }}
-          files: artifacts/linux-ut-result-*/**/*.xml
-          check_name: Linux Test Report
-          comment_title: Linux Test Report
+    uses: ./.github/workflows/publish-test-result-from-fork-action.yml
```
aceforeverd, line 18: seem not work because it can't find the file
see https://github.com/aceforeverd/fedb/actions/runs/3224277329/jobs/5275245948

Note those changes are bit complex to test, because workflow from `workflow_dispatch` only apply changes after changes merged into main branch. You may test over your own repo first. Here is the steps:

1. merge changes of this PR into main (or any default branch you set) into your own repo
2. create a pull request into your repo that triggers `cicd` , `hybridse` and `sdk` workflows
3. see if the three test report will appear in PR comment

here is my PR you can checkout https://github.com/aceforeverd/fedb/pull/2, I merged your commits into default branch before PR

aceforeverd (changes requested):

src/sdk/sql_cluster_router.cc
try-agaaain: Maybe I got it wrong again, and instead of making it go away, I introduced more changes after rebase. Is there a way to fix it?

.github/workflows/publish-test-result-from-fork.yml
try-agaaain, line 18: @aceforeverd,  Can you give me some tips? Although I googled the relevant materials, I was still not sure how to solve this problem:

```
Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/OpenMLDB/OpenMLDB/.github/workflows/publish-test-results-action.yml'. Did you forget to run actions/checkout before running your local action?
```
Any suggestions would be greatly appreciated.

try-agaaain:
@aceforeverd , This time the workflows are up and running. I tested the workflows in my repository. Although there were some check failures, but they didn't seem to be due to my changes. Here are some more details: https://github.com/team-317/OpenMLDB/actions

.github/workflows/publish-test-result-from-fork.yml
aceforeverd: I think use checkout action should fix this

try-agaaain: I've fixed this problem by use the full path such as `team-317/OpenMLDB/.github/workflows/publish-test-results-action.yml@reuse_workflow`. BTW, If this PR would to be merged into the main branch, maybe it should be modified to `4paradigm/OpenMLDB/.github/workflows/publish-test-results-action.yml@main`. 

Then I tested the workflows in my repository, and it seems to be working fine.

Is it possible to merge my PR?

aceforeverd:
Hi @team-317 

I merged your changes in my repo and give a few patch over that. I'm also testing it in https://github.com/aceforeverd/fedb/pull/5. The major problem here is we should not calls reusable workflow with the syntax
```
uses: team-317/OpenMLDB/.github/workflows/publish-test-results-action.yml@reuse_workflow
```
instead, it can be 
```
uses: ./.github/workflows/publish-test-results-action.yml
```
this should work after the PR merged into default branch.

Here is the patch: https://github.com/aceforeverd/fedb/commit/5b98f628f5a79084f0f7db004ed7415728116849

Hope we can figure it out !

.github/workflows/publish-test-result-from-fork.yml
aceforeverd: Yes, the PR is mostly fine for me. I'll try merge it this week, but before that fix the `uses` patch problem.

.github/workflows/publish-hybridse-test-result-from-fork.yml
```diff
@@ -8,66 +8,32 @@ on:
...
       github.event.workflow_run.conclusion != 'skipped' && (
          github.event.sender.login == 'dependabot[bot]' ||
          github.event.workflow_run.head_repository.full_name != github.repository
       )
+    name: Publish Linux UT Results
+    uses: team-317/OpenMLDB/.github/workflows/publish-test-result-from-fork-action.yml@reuse_workflow
+    with:
+      commit: ${{ github.event.workflow_run.head_sha }}
+      files: artifacts/linux-ut-result-*/**/*.xml
+      check_name: Linux Test Report
```
aceforeverd: `check_name` and `comment_title` should be `HybridSE Linux Test Report`

try-agaaain: Ok, thank you for your advice. I will revise it

.github/workflows/hybridse-ci.yml
```diff
@@ -85,29 +85,31 @@ jobs:
...
 
+  publish-test-results-mac:
+    needs: ["macos-build"]
+    # the action will only run on 4paradigm/HybridSE's context, not for fork repo or dependabot
+    if: >
+      always() && github.event_name == 'push' || (
+      github.event.pull_request.head.repo.full_name == github.repository &&
+      github.event.sender.login != 'dependabot[bot]' )
+    name: Publish Mac UT Results
+    uses: EnricoMi/publish-unit-test-result-action@v1
```
try-agaaain: Referring to your patch, I removed the `steps` and `runs-on` in `hybridse-ci.yml`. Then when I pushed this changes to my remote repository, I got the following error:

image

Do you encounter the same problem?

aceforeverd: Note it is not same 'uses' value as above line 97

try-agaaain: emmmmm, I don't quite understand.

try-agaaain: > Note it is not same 'uses' value as above line 97

Yes, I konw, reusable workflows should be referenced at the top-level `jobs.*.uses` key. So should I add `runs-on` and `steps` as before when use `EnricoMi/publish-unit-test-result-action@v1`?

aceforeverd: see https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow. There is no filename in the end. `EnricoMi/publish-unit-test-result-action@v1` is an action, not reusable workflow.

aceforeverd (approved):

vagetablechicken (approved):

PR #2625, opened 2022-10-09
fix: run `start.sh` failed if has not installed `curl`

### LINKED ISSUE #2624: run `start.sh` failed if has not installed `curl`
WX20221009-110044

dl239:
- run `start.sh` failed if has not installed `curl`
- update `sh` to `bash` in documents

### DIFF
--- a/release/bin/start.sh
+++ b/release/bin/start.sh
@@ -75,21 +75,30 @@ case $OP in
             ./bin/openmldb --flagfile=./conf/"$COMPONENT".flags --enable_status_service=true >>  "$LOG_DIR"."$COMPONENT".log 2>&1 &
             PID=$!
             sleep 3
-            ENDPOINT=$(grep '\--endpoint' ./conf/"$COMPONENT".flags | awk -F '=' '{print $2}')
-            COUNT=1
-            while [ $COUNT -lt 12 ]
-            do
-                if ! curl "http://$ENDPOINT/status" > /dev/null 2>&1; then
-                    sleep 1
-                    (( COUNT+=1 ))
-                elif kill -0 "$PID" > /dev/null 2>&1; then
+            if [ -x "$(command -v curl)" ]; then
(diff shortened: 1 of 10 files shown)


release/bin/start.sh
```diff
@@ -75,21 +75,30 @@ case $OP in
...
+                    elif kill -0 "$PID" > /dev/null 2>&1; then
+                        echo $PID > "$OPENMLDB_PID_FILE"
+                        echo "Start ${COMPONENT} success"
+                        exit 0
+                    else
+                        break
+                    fi
+                done
+            else
+                sleep 2
```
vagetablechicken: plz add some echo comments, e.g. "no curl, sleep 2 then check the process running status", in curl branch, " max wait time is 12s"

dl239: no curl message has added.  i think the max wait time is not necessary in startup script.

release/bin/start.sh
```diff
@@ -72,24 +72,34 @@ case $OP in
         fi
 
         if [ "$COMPONENT" != "taskmanager" ]; then
-            ./bin/openmldb --flagfile=./conf/"$COMPONENT".flags --enable_status_service=true >>  "$LOG_DIR"."$COMPONENT".log 2>&1 &
+            ./bin/openmldb --flagfile=./conf/"$COMPONENT".flags --enable_status_service=true >> "$LOG_DIR"."$COMPONENT".log 2>&1 &
```
vagetablechicken: If openmldb is incompatible, e.g. released bin runs on ubuntu22, will the error msg write to this log file?
If start failed, we'd better to print hint, like 'check log in "$LOG_DIR"."$COMPONENT".log and $LOG_DIR"/"$COMPONENT".INFO'

vagetablechicken (approved):

aceforeverd (approved):

PR #2626, opened 2022-10-09
feat: support validatesql

vagetablechicken:
use `ValidateSQL` to check if sql is valid and return error hint in batch mode and request mode.


vagetablechicken:
- [ ] long window
- [ ] engine options cluster_optimized_?

dl239 (approved):

aceforeverd (approved):

PR #2637, opened 2022-10-12
fix: traverse with same ts

### LINKED ISSUE #1696: `Traverse` may get duplicate data if there are same ts records on one pk
* `Traverse` may get duplicate data if there are same ts records on one pk
*  `Traverse` may run in endless loop if there are many same ts records on one pk

dl239:
close #1696

### DIFF
--- a/src/catalog/distribute_iterator.cc
+++ b/src/catalog/distribute_iterator.cc
@@ -131,7 +131,7 @@ bool FullTableIterator::NextFromRemote() {
                 kv_it_ = iter->second->Traverse(tid_, cur_pid_, "", last_pk_, last_ts_,
                             FLAGS_traverse_cnt_limit, false, kv_it_->GetTSPos(), count);
                 DLOG(INFO) << "pid " << cur_pid_ << " last pk " << last_pk_ <<
-                    " key " << last_ts_ << " count " << count;
+                    " key " << last_ts_ << " ts_pos " << kv_it_->GetTSPos() << " count " << count;
             } else {
                 iter++;
                 kv_it_.reset();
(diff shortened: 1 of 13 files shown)


vagetablechicken (approved):

src/catalog/distribute_iterator.cc
```diff
@@ -302,19 +302,22 @@ void DistributeWindowIterator::Next() {
         std::string cur_pk = kv_it_->GetPK();
         auto traverse_it = std::dynamic_pointer_cast<openmldb::base::TraverseKvIterator>(kv_it_);
         uint64_t last_ts = 0;
+        uint32_t ts_pos = 1;
```
zhanghaohit, line 305: should be 0 or 1 if `traverse_it` is not valid?

src/proto/tablet.proto
```diff
@@ -257,6 +257,7 @@ message TraverseRequest {
     optional uint64 ts = 6;
     optional bool enable_remove_duplicated_record = 7 [default = false];
     optional bool skip_current_pk = 8 [default = false];
+    optional uint64 ts_pos = 9;
```
zhanghaohit: `uint32_t`?

src/catalog/distribute_iterator.cc
dl239: should be 0 if traverse from another tablet

src/proto/tablet.proto
```diff
@@ -268,6 +269,7 @@ message TraverseResponse {
     optional uint64 ts = 6;
     optional bool is_finish = 7;
     optional uint64 snapshot_id = 8;
+    optional uint64 ts_pos = 9;
```
zhanghaohit: same here. `uint32`

zhanghaohit (approved):
LGTM

PR #2669, opened 2022-10-17
fix: iterator zero

### LINKED ISSUE #2631: rows with timestamp == 0 will be ignored in cluster mode
**Bug Description**
If there are rows with timestamp == 0, those rows is not read back in the cluster mode (multiple tablets).

There may be some bugs in the `DistributeWindowIterator`

**Steps to Reproduce**

1. load tables with ts == 0 data
2. run online batch window sql

dl239:
Use `Traverse` instead of `Scan` to get data from remote tablet in `DistributeIterator`


zhanghaohit (approved):

tobegit3hub (approved):
LGTM

PR #2745, opened 2022-11-01
feat: table put now returns false upon negative ts

jkpjkpjkp:


src/codec/codec.cc
```diff
@@ -346,6 +346,9 @@ bool RowBuilder::SetInt64(int8_t* buf, uint32_t index, int64_t val) {
 }
 
 bool RowBuilder::AppendTimestamp(int64_t val) {
```
aceforeverd, line 348: bad news is caller usually not check the return code. 

add the `ABSL_MUST_USE_RESULT` annotation

dl239:
done in #2776

PR #2754, opened 2022-11-04
feat(udf): hash64 function

aceforeverd:
## Brief

The function is called `hash64`, or `farm_fingerprint` same as BigQuery. `hash` is not used because it is a reserved keyword in parser.

## Detail

This is a non-cryptographic hash function, so do not use in cryptographic way. 

The underlaying hash algorithm uses googles [farmhash](https://github.com/google/farmhash), also borrowed from BigQuery. Same result is guaranteed to produced as long as both the data type and value is the same, no matter when to execute the query, or where the executing machine is. 

`hash64` accept a single parameter of type bool, int16, int32, int64, float, double, timestamp, string, date.

## Know Limitations

-  `hash64(*)` or hash multiple parameters does not supported yet. Maybe left as an enhancement later. 
- `hash(NULL)` produce `NULL` currently, which might not as expected behavior compared to spark sql etc. Hash function is defined as 
   > never returns NULL, even for NULL inputs.

   ref https://docs.snowflake.com/en/sql-reference/functions/hash.html.  This is the limitation of udf registry framework, and might handle it later #2755 .


hybridse/src/udf/literal_traits.h
```diff
@@ -408,26 +408,50 @@ struct DataTypeTrait<Tuple<T...>> {
...
 struct CCallDataTypeTrait<V*> {
     using LiteralTag = Opaque<V>;
 };
 template <>
 struct CCallDataTypeTrait<openmldb::base::Timestamp*> {
     using LiteralTag = openmldb::base::Timestamp;
+
+    static absl::string_view to_bytes_ref(openmldb::base::Timestamp** data) {
+        auto& ts = (*data)->ts_;
+        return CCallDataTypeTrait<decltype(openmldb::base::Timestamp::ts_)>::to_bytes_ref(&ts);
```
aceforeverd, line 434: this may produce exact same result to the same int64 input of `ts_`, the same apply to `Date` below. Discussion may necessary.

hybridse/src/udf/udf.h
```diff
@@ -203,6 +205,18 @@ struct IncOne {
     V operator()(V i) { return i + 1; }
 };
 
+int64_t FarmFingerprint(absl::string_view input);
+template <typename T>
+struct Hash64 {
+    using Args = std::tuple<T>;
```
dl239, line 211: what's the `using Args` for

aceforeverd, line 211: just required by udf registry framework

dl239 (approved):

zhanghaohit (approved):

PR #2817, opened 2022-11-23
feat: support array data type in SQL query and UDF system

aceforeverd:
## SQL

```sql
select [1, 2], ARRAY [1, 2], ARRAY<INT64>[4, 5]
```

- array of array is not supported : `ARRAY<ARRAY<INT>>`
- array is nullable
- array type is not supported in storage, so select array as a column won't allowed
   - *currently, array type only appears as UDF parameter and return values*

### Type inference

If array type is specified in SQL : `ARRAY<T>[...]`, take the SQL type; otherwise, element type is inferred, e.g `INT + INT64 -> INT64`

## UDF

```cpp
template <typename T, typename TT = typename DataTypeTrait<T>::CCallArgType>
ArrayRef {
   TT* raw;               // array of raw elements
   bool* nullables;   // array of null infos of each element
   uint64_t size;      // array size
}
```

### UDF registry type

`ArrayRef<T>`, where `T` is `bool/intxx/float/double/StringRef/Timestamp/Date`

### C Calling type

`ArrayRef<T>*`, `T` is same the as in UDF registry type

### Develop UDF with array type

**array as input param**

read the `ArrayRef<T>::raw/nullables/size`, have a example `array_contains`

**array as output value**

- make `ArrayRef<T>*` as last param in external udf define
- when array size if determined, use `hybridse::udf:v1::AllocManagedArray` to alloc space for the array type

*Other register way e.g with Codegen UDF is not tested, not guaranteed to work


hybridse/src/codegen/array_ir_builder.cc
```diff
@@ -0,0 +1,117 @@
...
+
+    ::llvm::Value* array_sz = builder.CreateLoad(idx_val_ptr);
+    CHECK_TRUE(Set(bb, array_alloca, 2, array_sz), common::kCodegenError);
+
+    *output = NativeValue::Create(array_alloca);
+    return base::Status::OK();
+}
+
+
+base::Status ArrayIRBuilder::NexEmptyArray(llvm::BasicBlock* bb, NativeValue* output) const {
```
zhanghaohit: typo `New`

hybridse/src/codegen/array_ir_builder.h
```diff
@@ -0,0 +1,66 @@
...
+
+#include "absl/base/attributes.h"
+#include "codegen/struct_ir_builder.h"
+
+namespace hybridse {
+namespace codegen {
+
+// Array Struct, consist of two fields
+// - Array of raw values
+// - Array of nullable values
```
aceforeverd: comment

hybridse/src/node/type_node.cc
```diff
@@ -103,5 +121,13 @@ OpaqueTypeNode *OpaqueTypeNode::ShadowCopy(NodeManager *nm) const {
     return nm->MakeOpaqueType(bytes_);
 }
 
+// {base_name}<element_type>(num_elements)
+std::string FixedArrayType::DebugString() const { return absl::StrCat(TypeNode::DebugString(), "(", num_elements_, ")"); }
```
aceforeverd: lint

hybridse/src/udf/udf.h
```diff
@@ -38,6 +38,99 @@ using openmldb::base::Timestamp;
 
 namespace v1 {
 
+// ============================================= //
+//       Object Lieftime Management
+//           Helpter Classes
```
aceforeverd: typo

hybridse/src/udf/default_defs/array_def.cc
```diff
@@ -0,0 +1,116 @@
...
+
+             @code{.sql}
+                 select array_contains([2,2], 2) as c0;
+                 -- output true
+             @endcode
+
+             @since 0.7.0
+             )");
+
+    RegisterExternal("split_array")
```
zhanghaohit, line 101: it has the same functionality as `split`?

can I do `select split_array("2,1", ",")`?

aceforeverd: No, array type is not supported in storage. 

The two has same purpose, leaves later to handle the difference between array and list.

zhanghaohit (approved):

dl239 (approved):

PR #2864, opened 2022-12-12
feat: size udf

zhanghaohit:
support `size` like `size(split("a b c", " "))`

### DIFF
--- a/cases/function/test_feature_zero_function.yaml
+++ b/cases/function/test_feature_zero_function.yaml
@@ -174,3 +174,30 @@ cases:
         - [3, "", "", "", "", "", ""]
         - [4, "", "", "", "", "", ""]
         - [5, "", "", "", "", "", ""]
+  - id: 6
+    desc: size function of split result on single row
+    inputs:
+      - name: main
+        columns: ["id int64", "pk int64", "c1 string"]
+        indexs: ["index1:pk:id"]
+        rows:
+          - [1, 0, "k1:v1,k2:v2"]
+          - [2, 0, "k3:v3"]
+          - [3, 0, "???,,k4:v4"]
+          - [4, 0, NULL]
+          - [5, 0, "k5:v5,k3:v3"]
+    sql: |
+      SELECT id,
+        size(fz_split(c1, ",")) AS split_and_size,
+        size(fz_split_by_key(c1, ",", ":")) AS split_key_and_size,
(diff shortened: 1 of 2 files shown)


hybridse/src/udf/default_defs/feature_zero_def.cc
```diff
@@ -379,6 +379,11 @@ struct FZStringOpsDef {
         output->size_ = bytes;
         output->data_ = buf;
     }
+
+    static void ListSize(hybridse::codec::ListRef<StringRef>* list, int32_t* size) {
+        auto list_v = reinterpret_cast<hybridse::codec::ListV<StringRef> *>(list->list);
+        *size = list_v->GetCount();
```
dl239: if the result value is not-null,  return by the function return value? refer [build-in function docs] (https://github.com/4paradigm/OpenMLDB/blob/main/docs/zh/developer/built_in_function_develop_guide.md?plain=1#L91)

dl239 (approved):

cases/function/test_feature_zero_function.yaml
```diff
@@ -174,3 +174,30 @@ cases:
...
+        indexs: ["index1:pk:id"]
+        rows:
+          - [1, 0, "k1:v1,k2:v2"]
+          - [2, 0, "k3:v3"]
+          - [3, 0, "???,,k4:v4"]
+          - [4, 0, NULL]
+          - [5, 0, "k5:v5,k3:v3"]
+    sql: |
+      SELECT id,
+        size(fz_split(c1, ",")) AS split_and_size,
```
aceforeverd: use split， fz split is deprecated

cases/function/test_feature_zero_function.yaml
```diff
@@ -174,3 +174,30 @@ cases:
...
+        indexs: ["index1:pk:id"]
+        rows:
+          - [1, 0, "k1:v1,k2:v2"]
+          - [2, 0, "k3:v3"]
+          - [3, 0, "???,,k4:v4"]
+          - [4, 0, NULL]
+          - [5, 0, "k5:v5,k3:v3"]
+    sql: |
+      SELECT id,
+        size(split(c1, ",")) AS split_and_size,
```
zhanghaohit, line 191: - check `size` over one single `column`

zhanghaohit: `type` not match

aceforeverd (approved):

PR #2875, opened 2022-12-13
refactor: delete root_path_dir after run case

dl239:

### DIFF
--- a/src/test/util.h
+++ b/src/test/util.h
@@ -18,12 +18,14 @@
 #define SRC_TEST_UTIL_H_
 
 #include <algorithm>
+#include <filesystem>
 #include <map>
 #include <memory>
 #include <string>
 #include <utility>
 #include <vector>
 
+#include "absl/strings/str_cat.h"
 #include "brpc/server.h"
 #include "codec/sdk_codec.h"
 #include "gflags/gflags.h"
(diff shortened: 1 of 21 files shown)


src/test/util.h
```diff
@@ -39,11 +41,34 @@ namespace openmldb {
...
-
 inline std::string GenRand() {
     return std::to_string(rand() % 10000000 + 1);  // NOLINT
 }
 
+class TempPath {
+ public:
+    TempPath() {
+        std::filesystem::path tmp_path = std::filesystem::temp_directory_path();
+        base_path_ = absl::StrCat(tmp_path.string(), "/openmldb_test", ::openmldb::test::GenRand());
```
aceforeverd: filesystem::path supports append with `operator/`

src/test/util.h
```diff
@@ -39,11 +41,34 @@ namespace openmldb {
...
+    }
+    std::string GetTempPath() {
+        return absl::StrCat(base_path_, "/", ::openmldb::test::GenRand());
+    }
+    std::string GetTempPath(const std::string& prefix) {
+        return absl::StrCat(base_path_, "/", prefix, ::openmldb::test::GenRand());
+    }
+    std::string CreateTempPath(const std::string& prefix) {
+        std::string path = GetTempPath(prefix);
+        std::filesystem::create_directory(path);
```
aceforeverd: check ret value

tobegit3hub (approved):
LGTM

src/test/util.h
dl239: no need to check, program will core dump if create failed(not catch exception).

aceforeverd (approved):

PR #2876, opened 2022-12-13
docs: add debug hints in pulsar demo

vagetablechicken:


aceforeverd (approved):

docs/en/use_case/pulsar_connector_demo.md
```diff
@@ -79,7 +79,7 @@ bin/pulsar-daemon start standalone --zookeeper-port 5181
 ```{note}
 OpenMLDB want to use the port 2181, so we should change the zk port here. We will use zk port 2181 to connect OpenMLDB, but zk port in Pulsar standalone won’t affect anything.
 ```
-You can check if the pulsar runs well, `ps` or check the log.
+You can check if the pulsar runs well, `ps` or check the standalone server log `logs/pulsar-standalone-....log`.
```
zhanghaohit: the `cn` version should also update?

tobegit3hub (approved):
LGTM

PR #2894, opened 2022-12-21
fix: deploy failed if the column name of main table is keyword

dl239:
- print table name if ttl_type mismatch
- deploy failed if the column name of main table is keyword like `index`


tobegit3hub (approved):
LGTM

zhanghaohit (approved):

PR #2926, opened 2023-01-03
fix: fix tablet coredump

dl239:
- tablet core dump
WX20221230-163951

the value of rocksdb iterator will invalid if call the next/seek/seektofirst
https://github.com/facebook/rocksdb/blob/main/include/rocksdb/iterator.h#L87-L89
- `ShowMemPool` does not print memory stat


zhanghaohit (approved):

tobegit3hub (approved):
LGTM

PR #2936, opened 2023-01-05
fix: set guava version for openmldb-batch

### LINKED ISSUE #2935: Offline jobs fail because of curator API is incorrect
Now we will get the following error for docker-compose demo case.

Screen Shot 2023-01-05 at 16 22 55

It may be the issue of guava versions which are different in openmldb-batch and spark distribution.

tobegit3hub:
* Resolve https://github.com/4paradigm/OpenMLDB/issues/2935 .

### DIFF
--- a/java/openmldb-batch/pom.xml
+++ b/java/openmldb-batch/pom.xml
@@ -67,6 +67,12 @@
 			<version>${scala.version}</version>
 		</dependency>
 
+		<dependency>
+			<groupId>com.google.guava</groupId>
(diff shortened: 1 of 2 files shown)


tobegit3hub:
Verified with docker-compose test case. Now all the offline jobs work with latest openmldb-batch jar.

Screen Shot 2023-01-06 at 17 12 58

dl239 (approved):

vagetablechicken (approved):

PR #2943, opened 2023-01-10
fix: disk table memory leak

dl239:


src/storage/disk_table.cc
```diff
@@ -436,9 +439,8 @@ void DiskTable::GcHead() {
                 }
             }
         }
-        delete it;
-        db_->ReleaseSnapshot(snapshot);
     }
+    db_->ReleaseSnapshot(snapshot);
```
aceforeverd: use `absl::Cleanup` or smart pointer in case accidentally returned inside loop

zhanghaohit (approved):

aceforeverd (approved):

PR #2999, opened 2023-01-30
refactor: rename pk and remove the traverse limit by default

### LINKED ISSUE #1412: Get table row count of online table
Users may try `select * from t1;` to get the count of table `t1`. But we may truncate the scan response in tablet server(by `FLAGS_scan_max_bytes_size`).
So we should:
1. add `preview` or other cmd to get the count(sum of TableInfo), note that the count may be not accurate and why.
2. add doc about `select *` does not work well if table is big.

dl239:
close #1412


release/conf/tablet.flags.template
```diff
@@ -81,10 +81,10 @@
...
 
 # query conf
-# max table traverse iteration（full table scan/aggregation）,default: 50000
-#--max_traverse_cnt=50000
-# max table traverse pk number（batch query）, default: 5000
-#--max_traverse_pk_cnt=5000
+# max table traverse iteration(full table scan/aggregation),default: 0
+#--max_traverse_cnt=0
+# max table traverse unique key number(batch query), default: 0
+#--max_traverse_key_cnt=0
```
aceforeverd, line 87: must notice in changelog

lumianph: will it cause any issues if doing the upgrade? will the users keep their old config file?

dl239: > will it cause any issues if doing the upgrade? will the users keep their old config file?

if using the old config file,  the result of online batch computing will be truncated as before

dl239: > must notice in changelog

ok

docs/zh/deploy/conf.md
```diff
@@ -182,10 +182,10 @@
...
 
 # 查询配置
-# 最大扫描条数（全表扫描/全表聚合），默认：50000
-#--max_traverse_cnt=50000
-# 最大扫描pk数（批处理），默认：5000
-#--max_traverse_pk_cnt=5000
+# 最大扫描条数(全表扫描/全表聚合)，默认：0
+#--max_traverse_cnt=0
+# 最大扫描不同key的个数(批处理)，默认：0
+#--max_traverse_pk_cnt=0
```
zhanghaohit: --max_traverse_key_cnt

src/catalog/distribute_iterator.cc
```diff
@@ -38,14 +38,17 @@ void FullTableIterator::SeekToFirst() {
 }
 
 bool FullTableIterator::Valid() const {
-    return (cnt_ <= FLAGS_max_traverse_cnt) && ((it_ && it_->Valid()) || (kv_it_ && kv_it_->Valid()));
```
zhanghaohit, line 41: this is not right in the first place. should be `cnt_ < FLAGS_max_traverse_cnt`?

see
```
bool DiskTableTraverseIterator::Valid() {
    if (traverse_cnt_ >= FLAGS_max_traverse_cnt) {
        return false;
    }
```

src/tablet/tablet_impl.cc
```diff
@@ -1504,14 +1504,14 @@ void TabletImpl::Traverse(RpcController* controller, const ::openmldb::api::Trav
...
         scount++;
-        if (it->GetCount() >= FLAGS_max_traverse_cnt) {
+        if (FLAGS_max_traverse_cnt > 0 && it->GetCount() >= FLAGS_max_traverse_cnt) {
             DEBUGLOG("traverse cnt %lu max %lu, key %s ts %lu", it->GetCount(), FLAGS_max_traverse_cnt, last_pk.c_str(),
                      last_time);
             break;
         }
     }
     bool is_finish = false;
-    if (it->GetCount() >= FLAGS_max_traverse_cnt) {
```
zhanghaohit, line 1514: here should be `it->GetCount() + 1 >= FLAGS_max_traverse_cnt`? as it->GetCount() is from 0, and here the `it->GetCount()` is not increased before the `it->Next()`.

zhanghaohit (changes requested):

src/catalog/distribute_iterator.cc
dl239: yes, it's better to be consistent

src/tablet/tablet_impl.cc
```diff
@@ -1504,14 +1504,14 @@ void TabletImpl::Traverse(RpcController* controller, const ::openmldb::api::Trav
         map_it->second.emplace_back(it->GetKey(), value);
         total_block_size += last_pk.length() + value.size();
         scount++;
-        if (it->GetCount() >= FLAGS_max_traverse_cnt) {
+        if (FLAGS_max_traverse_cnt > 0 && it->GetCount() >= FLAGS_max_traverse_cnt) {
```
zhanghaohit, line 1507: same for here `it->GetCount() + 1 >= FLAGS_max_traverse_cnt`

zhanghaohit (approved):

src/tablet/tablet_impl.cc
dl239: traverse_cnt will increase in `SeekToFirst` refer  [here](https://github.com/4paradigm/OpenMLDB/blob/v0.7.1/src/storage/mem_table.cc#L999)

tobegit3hub (approved):
LGTM

lumianph (approved):

PR #3025, opened 2023-02-07
feat: add std built-in function

### LINKED ISSUE #1961: feat: the new aggregate function `std`
You may follow the built-in aggregate function development doc (Chinese)

https://openmldb.ai/docs/zh/main/developer/built_in_function_develop_guide.html

zhanghaohit:
Fixes #1961

Support two versions of std functions:
- `std`/`stddev`/`stddev_samp`: sample standard devision 
image
- `stddev_pop`: population standard devision
image

where `u` is the avg and `N` is the number of values. It has the same semantics as the SparkSQL (however different from MySQL)

### DIFF
--- a/hybridse/src/udf/default_udf_library.cc
+++ b/hybridse/src/udf/default_udf_library.cc
@@ -273,14 +273,14 @@ struct AvgUdafDef {
 };
 
 template <typename T>
-struct StdUdafDef {
+struct StdPopUdafDef {
     using ContainerT = std::pair<std::vector<T>, double>;
     void operator()(UdafRegistryHelper& helper) {  // NOLINT
         std::string suffix = absl::StrCat(".opaque_std_pair_std_vector_double_", DataTypeTrait<T>::to_string());
         helper.templates<Nullable<double>, Opaque<ContainerT>, Nullable<T>>()
-            .init("std_init" + suffix, Init)
-            .update("std_update" + suffix, Update)
-            .output("std_output" + suffix, reinterpret_cast<void *>(Output), true);
+            .init("std_pop_init" + suffix, Init)
+            .update("std_pop_update" + suffix, Update)
+            .output("std_pop_output" + suffix, reinterpret_cast<void *>(Output), true);
     }
 
     static void Init(ContainerT* ptr) {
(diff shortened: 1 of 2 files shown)


aceforeverd (approved):

dl239 (approved):

hybridse/src/udf/default_udf_library.cc
```diff
@@ -272,6 +272,47 @@ struct AvgUdafDef {
...
+        if (ptr->first.size() == 0) {
+            *is_null = true;
+        } else {
+            size_t cnt = ptr->first.size();
+            double avg = ptr->second / cnt;
+            double stddev = 0;
+            for (size_t i = 0; i < cnt; i++) {
+                stddev += std::pow(ptr->first[i] - avg, 2);
+            }
+            stddev = std::sqrt(stddev / cnt);
```
aceforeverd: Sample Standard Deviation should divide by `cnt -1` ?

hybridse/src/udf/default_udf_library.cc
```diff
@@ -2679,6 +2720,28 @@ void DefaultUdfLibrary::InitUdaf() {
         )")
         .args_in<int16_t, int32_t, int64_t, float, double>();
 
+    RegisterUdafTemplate<StdUdafDef>("std")
```
aceforeverd: should we should support population standard deviation too ?

aceforeverd (changes requested):

hybridse/src/udf/default_udf_library.cc
zhanghaohit: oh. the formula I previously gave in the PR text is wrong. The correct one should be 
image
 (population standard deviation), which is also the definition in MySQL.

hybridse/src/udf/default_udf_library.cc
zhanghaohit: support both `std` and `std_samp`

hybridse/src/udf/default_udf_library.cc
```diff
@@ -2679,6 +2750,58 @@ void DefaultUdfLibrary::InitUdaf() {
...
```
aceforeverd: let's be compatible with spark sql: https://spark.apache.org/docs/latest/api/sql/index.html#stddev
- `std`, `stddev`, `stddev_samp` are sample standard deviation
- `stddev_pop` is population standard deviation

zhanghaohit: ok. it's different from the `mysql` conventions.

aceforeverd: we're more relevant to spark, offline mode is based on spark.

aceforeverd (approved):

PR #3029, opened 2023-02-08
feat: support `drawdown` built-in func

zhanghaohit:
```
Drawdown is defined as the max decline percentage from a historical peak to a subsequent valley.
It is commonly used as an indicator of risk in quant-trading to measure the max loss.
```

`drawdown=mdd(T)`

image
image
image

### DIFF
--- a/hybridse/src/udf/default_udf_library.cc
+++ b/hybridse/src/udf/default_udf_library.cc
@@ -16,6 +16,7 @@
 
 #include "udf/default_udf_library.h"
 
+#include <algorithm>
 #include <string>
 #include <tuple>
 #include <unordered_set>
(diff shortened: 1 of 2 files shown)


hybridse/src/udf/default_udf_library.cc
```diff
@@ -552,6 +553,53 @@ struct TopKDef {
...
+    }
+
+    static void Init(ContainerT* ptr) {
+        new (ptr) ContainerT(-1, 0);
+    }
+
+    static ContainerT* Update(ContainerT* ptr, T t, bool is_null) {
+        if (!is_null) {
+            if (t < 0) {
+                LOG(ERROR) << "drawdown only supports positive values";
```
aceforeverd: there may too many error logs

hybridse/src/udf/default_udf_library.cc
```diff
@@ -2873,6 +2921,34 @@ void DefaultUdfLibrary::InitUdaf() {
         )")
         .args_in<int16_t, int32_t, int64_t, float, double>();
 
+    RegisterUdafTemplate<DrawdownUdafDef>("drawdown")
+        .doc(R"(
+            @brief Compute drawdown of values.
+
+            Drawdown is defined as the max decline percentage from a historical peak to a subsequent valley.
```
aceforeverd, line 2931: so the base is the history peek value ?

aceforeverd (approved):

hybridse/src/udf/default_udf_library.cc
dl239: use `LOG_FIRST_N` ?

hybridse/src/udf/default_udf_library.cc
zhanghaohit: yes. the base is the history peak value.

zhanghaohit:
current implementation has some problems, as the iterator is in the reversing order.

hybridse/src/udf/default_udf_library.cc
zhanghaohit: noted. will change to `LOG_FIRST_N`

hybridse/src/udf/default_udf_library.cc
```diff
@@ -2873,6 +2921,34 @@ void DefaultUdfLibrary::InitUdaf() {
...
+            |--|
+            |1|
+            |8|
+            |5|
+            |2|
+            |10|
+            |4|
+            @code{.sql}
+                SELECT drawdown(value) OVER w;
+                -- output 0.75 (decline from 8 to 2)
```
zhanghaohit, line 2953: the decline percentage from 10 to 4 is 0.6.

we are trying to find the max percentage rather than find the `max decline` / `max peak value`

zhanghaohit:
> current implementation has some problems, as the iterator is in the reversing order.

fixed. now assuming we are `update` the data in the reversing order of timestamp, i.e., from largest to smallest.

dl239 (approved):

hybridse/src/udf/default_udf_library.cc
```diff
@@ -552,6 +554,55 @@ struct TopKDef {
     }
 };
 
+template <typename T>
+struct DrawdownUdafDef {
+    // <drawdown, current peak>
```
zhanghaohit: comment not correct

tobegit3hub (approved):
LGTM

PR #3046, opened 2023-02-14
feat: new builtin functions

aceforeverd:
- `var_samp`, `var_pop`
- `entropy`:  same as https://rosettacode.org/wiki/Entropy
- earth_distance: same as https://rosettacode.org/wiki/Haversine_formula
- nth_value_where


hybridse/src/udf/default_defs/window_functions_def.cc
```diff
@@ -91,6 +92,74 @@ node::ExprNode* BuildAt(UdfResolveContext* ctx, ExprNode* input, ExprNode* idx,
...
+        helper.library()
+            ->RegisterUdafTemplate<Impl>(helper.name())
+            .doc(helper.GetDoc())
+            .template args_in<int16_t, int32_t, int64_t>();
+    }
+
+    template <typename NthType>
+    struct Impl {
+        // (nth idx, [(value, value_is_null)])
+        using ContainerT = std::pair<NthType, std::vector<std::pair<T, bool>>>;
```
aceforeverd: avoid vector overhead when `nth < 0` ?

hybridse/src/udf/default_defs/aggregate_def.cc
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    static void Init(ContainerT* addr) { new (addr) ContainerT({}, 0); }
+
+    static ContainerT* Update(ContainerT* ctr, CType value, bool is_null) {
+        if (is_null) {
+            return ctr;
+        }
+        auto val = container::ContainerStorageTypeTrait<T>::to_stored_value(value);
+        if (ctr->first.count(val) == 0) {
```
dl239: prefer to use `find` to avoid querying twice

cases/function/function/test_udaf_function.yaml
```diff
@@ -2596,3 +2596,87 @@ cases:
         - [4, NULL, "g1:2", "g2:3", "g2:4.000000,g1:4.000000", "55:10", "4:4.000000,2:2.000000,1:2.000000"]
         - [5, 5,    "g1:2", "g2:3", "g2:4.000000,g1:4.000000", "55:15", "5:4.000000,4:4.000000,2:2.000000"]
         - [6, 5,    "g1:2", "g2:3", "g2:4.000000,g1:4.000000", "55:15", "4:4.000000,5:2.500000,2:2.000000"]
+
+  - id: 63
+    desc: entropy
+    mode: request-unsupport
+    inputs:
+      - name: t1
+        schema: id:int32, col1:int32, col2:int16, col3:string, col4 double, std_ts:timestamp
```
dl239: add `bool` and `date`

hybridse/src/udf/default_defs/window_functions_def.cc
```diff
@@ -91,6 +92,74 @@ node::ExprNode* BuildAt(UdfResolveContext* ctx, ExprNode* input, ExprNode* idx,
...
+            if (ctr->first > 0) {
+                // count from window start to window end
+                *value = ctr->second[sz - ctr->first].first;
+                *is_null = ctr->second[sz - ctr->first].second;
+                return;
+            }
+
+            // count from window end
+            *value = ctr->second[-ctr->first - 1].first;
+            *is_null = ctr->second[-ctr->first - 1].second;
```
aceforeverd: free ctr

hybridse/src/udf/default_defs/window_functions_def.cc
```diff
@@ -90,6 +92,105 @@ node::ExprNode* BuildAt(UdfResolveContext* ctx, ExprNode* input, ExprNode* idx,
...
+                                  bool cond_is_null) {
+            // by default, update is iterated from window end to window start
+            if (ctr->nth == 0) {
+                // update only once, we treat nth as constant
+                ctr->nth = nth;
+            }
+            if (cond && !cond_is_null) {
+                if (ctr->nth > 0) {
+                    // nth from window start
+                    ctr->data.emplace_back(container::ContainerStorageTypeTrait<T>::to_stored_value(value),
```
dl239: a fixed-size buffer like dequeue or list (push_back and pop_front) may work well

aceforeverd: sounds good, I impl a simple ring buffer for this

hybridse/src/udf/default_defs/window_functions_def.cc
```diff
@@ -90,6 +92,131 @@ node::ExprNode* BuildAt(UdfResolveContext* ctx, ExprNode* input, ExprNode* idx,
...
 
+
+template <typename T>
+struct RingBuffer {
+    explicit RingBuffer(size_t sz, const T& default_val) : idx(0), sz(sz), data(sz, default_val) {}
+
+    template <typename... Arg>
+    void Emplace(Arg&&... val) {
+        data[idx % sz] = T(std::forward<Arg>(val)...);
+        idx++;
```
dl239: > - the position of 0 or nth is not the real nth? 

does not matter, for `nth > 0`, next value in the array always refer to nth value, since array is size of `nth`

> - consider the case that n is much larger than the number of records(
extreme case or incorrect input).  it is not recommended to allocate the `n` size buffer at the begging

so maybe `deque` is generally better ?

dl239: if use vector, one optional solution is keep nth as the member. if the size of vector is less than nth, push_back directly. otherwise compute the real idx as current implement. Absolutely, it need additional judgement  in `At` or `output` function if the size less than nth

aceforeverd: Dynamically sized vector issues many relocation if nth is huge. In that case, deque is better.

aceforeverd: using deque has disadvantage too, according https://en.cppreference.com/w/cpp/container/deque
> deques typically have large minimal memory cost; a deque holding just one element has to allocate its full internal array (e.g. 8 times the object size on 64-bit libstdc++; 16 times the object size or 4096 bytes, whichever is larger, on 64-bit libc++)

aceforeverd: I choose the `std::queue<Container = std::list>`, which is out of relocation , not so huge startup space, and *O(1)*  push and pop. General memory space should a little higher compared to `std::vector`. 

Comment below if there is a better container.

hybridse/src/udf/default_defs/window_functions_def.cc
```diff
@@ -90,6 +94,112 @@ node::ExprNode* BuildAt(UdfResolveContext* ctx, ExprNode* input, ExprNode* idx,
...
+                    *is_null = true;
+                    return;
+                }
+                *value = ctr->data.front().first;
+                *is_null = ctr->data.front().second;
+                return;
+            }
+
+            // count from window end
+            if (ctr->data.empty()) {
```
dl239: it may be more simply as below: 
```
if (ctr->nth == 0 || ctr->data.empty() || (ctr->nth > 0 && ctr->data.size() < ctr->nth)) {
...
}
*value = ctr->data.front().first;
*is_null = ctr->data.front().second;
```

dl239 (approved):

vagetablechicken (approved):

PR #3061, opened 2023-02-15
feat: auto set timeout of offline sync job and add onsync mode in apiserver

### LINKED ISSUE #3031: use static flag sync_job_timeout for offline sync job, no need to set job timeout in runtime
offline async jobs and other jobs still use variable `job_timeout`.

So if set sync_job to true, we will wait for sync_job_timeout. 
Doc about it on sync_job:
if met timeout, it may be the EOF of Socket, ...(copy from FAQ)

vagetablechicken:
closes #3031


tobegit3hub (approved):
LGTM

dl239 (approved):

PR #2846, opened 2022-12-07
feat: support with clause

aceforeverd:
1. WITH clause
    Syntax and CTE rules follows Bigquery: https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#with_clause
2. fix doc of `*_cate_where`, close #2882

PR #3064, opened 2023-02-16
fix: openmldb_api_test compile failed

dl239:
WX20230216-205403

### DIFF
--- a/src/sdk/CMakeLists.txt
+++ b/src/sdk/CMakeLists.txt
@@ -298,14 +298,8 @@ COMMENT "pack static cxx sdk and all depend static libs into one"
 )
 
 if(TESTING_ENABLE)
-    add_library(openmldb_cxx_sdk STATIC IMPORTED GLOBAL)
-    add_dependencies(openmldb_cxx_sdk pack_openmldbsdk)
-    set_target_properties(openmldb_cxx_sdk PROPERTIES
-    IMPORTED_LOCATION ${CMAKE_CURRENT_BINARY_DIR}/libopenmldbsdk.a
-    )
     add_executable(openmldb_api_test openmldb_api_test.cc)
-    # add ${BIN_LIBS} because minicluster
-    target_link_libraries(openmldb_api_test openmldb_cxx_sdk ${BIN_LIBS} ${GTEST_LIBRARIES})
+    target_link_libraries(openmldb_api_test ${BIN_LIBS} ${GTEST_LIBRARIES})
 endif()
 
 FILE(GLOB USER_HEADER 


src/sdk/CMakeLists.txt
```diff
@@ -298,14 +298,8 @@ COMMENT "pack static cxx sdk and all depend static libs into one"
...
 if(TESTING_ENABLE)
-    add_library(openmldb_cxx_sdk STATIC IMPORTED GLOBAL)
-    add_dependencies(openmldb_cxx_sdk pack_openmldbsdk)
-    set_target_properties(openmldb_cxx_sdk PROPERTIES
-    IMPORTED_LOCATION ${CMAKE_CURRENT_BINARY_DIR}/libopenmldbsdk.a
-    )
     add_executable(openmldb_api_test openmldb_api_test.cc)
-    # add ${BIN_LIBS} because minicluster
-    target_link_libraries(openmldb_api_test openmldb_cxx_sdk ${BIN_LIBS} ${GTEST_LIBRARIES})
+    target_link_libraries(openmldb_api_test ${BIN_LIBS} ${GTEST_LIBRARIES})
```
vagetablechicken, line 302: Openmldb api test can run? No openmldb api lib.

dl239: yes, `${BIN_LIBS}` contains the sdk lib

src/sdk/CMakeLists.txt
```diff
@@ -298,14 +298,8 @@ COMMENT "pack static cxx sdk and all depend static libs into one"
 )
```
vagetablechicken, line 298: I used to fix Malformed archive error by rm before ar. Maybe ar -crs, remove’-T’ works?

dl239: rm `-T` work well, but it will generate about 7G static file.

aceforeverd: What't the reason of malformed archive from latest main ?

src/sdk/CMakeLists.txt
vagetablechicken: OK, but we should test that, standalone openmldb api lib can work. May be we should test in other places, e.g. a standalone cpp test with onebox cluster, or docker-compose test. Let's create a issue about it.

src/sdk/CMakeLists.txt
dl239: has not found the root cause. it compiling fails since #2846 merged

vagetablechicken: libopenmldbsdk.a and sdk lib in ${BIN_LIBS} have the same part, may be it's the root cause. Seems like link again. Or just ar `-T` bug 😢

vagetablechicken (approved):

src/sdk/CMakeLists.txt
dl239: but it seems strange that why it can works before. #2846 does not modify the lib

vagetablechicken: > has not found the root cause. it compiling fails since #2846 merged

👍 So it seems like no big change in cmakelists, maybe reach some resources limit?

tobegit3hub (approved):
LGTM

zhanghaohit (approved):

PR #3097, opened 2023-02-23
fix: rm curator when building docker

dl239:
#3089 has removed `curator` in docker-compose. it also need to rm when building docker
WX20230223-160406@2x


tobegit3hub (approved):
LGTM

vagetablechicken (approved):

PR #3112, opened 2023-03-01
feat: get job fail status for show job commands

### LINKED ISSUE #3078: add error msg when GetJobResultSet failed
https://github.com/4paradigm/OpenMLDB/blob/4045757cdfd19dc8d01c1f6e9b7dd6511160ebbc/src/sdk/sql_cluster_router.cc#L4194-L4208

If query job status failed(e.g. job table is broken...), we can't know, GetJobResultSet just return an empty resultset.

tobegit3hub:
* Resolve https://github.com/4paradigm/OpenMLDB/issues/3078 .

If the internal table is corrupt, we will get the error message now.

show_job_fail

### DIFF
--- a/src/sdk/sql_cluster_router.cc
+++ b/src/sdk/sql_cluster_router.cc
@@ -1837,7 +1837,13 @@ std::shared_ptr<hybridse::sdk::ResultSet> SQLClusterRouter::HandleSQLCmd(const h
                 *status = {StatusCode::kCmdError, "Failed to parse job id: " + cmd_node->GetArgs()[0]};
                 return {};
             }
-            return this->GetJobResultSet(job_id);
+
+            auto rs = this->GetJobResultSet(job_id, status);
+            if (status->IsOK()) {
+                return rs;
+            } else {
+                return {};
+            }
         }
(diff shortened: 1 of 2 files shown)


dl239 (approved):

src/sdk/sql_cluster_router.cc
```diff
@@ -1837,7 +1837,13 @@ std::shared_ptr<hybridse::sdk::ResultSet> SQLClusterRouter::HandleSQLCmd(const h
                 *status = {StatusCode::kCmdError, "Failed to parse job id: " + cmd_node->GetArgs()[0]};
                 return {};
             }
-            return this->GetJobResultSet(job_id);
+
+            auto rs = this->GetJobResultSet(job_id, status);
+            if (status->IsOK()) {
```
vagetablechicken: No need to check if status is ok? If status is not ok, GetJobResultSet will return `{}`
Or you can prepend the status msg in `else`, to let us know where we got the error, in show job, select into or others

tobegit3hub: Okay.

vagetablechicken (approved):

PR #3141, opened 2023-03-13
feat(udf doc gen): one udf doc fix and udf generate title 

aceforeverd:
There are two fixes:
- fixes the doc example for `join`
- fixes title of the generated udf markdown file


dl239 (approved):

tobegit3hub (approved):
LGTM

PR #3193, opened 2023-04-05
feat(udf): add_months

aceforeverd:
1. Add the `add_months` generally same as in [spark sql](https://spark.apache.org/docs/latest/api/sql/index.html#add_months)
2. Make general improve to udf registry system, allow developer to register external function more safely. 

The improvement make it possible to register void external function like `void fn(Args..., Ret*, bool*)` with `ExternalFuncRegistryHelper::args<Args...>(fn)`. Yet previously, you must go with `void fn(Args..., Ret*, bool*)` with `ExternalFuncRegistryHelper::args<Args...>(reinterpret_cast<void*>(fn))`

### DIFF
--- a/cases/query/simple_query.yaml
+++ b/cases/query/simple_query.yaml
@@ -985,3 +985,12 @@ cases:
       data: |
         1, 10, 1000
         2, 30, 2000
+  - id: 108
+    inputs:
+      - name: t1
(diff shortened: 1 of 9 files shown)


dl239 (approved):

tobegit3hub (approved):
LGTM

PR #3199, opened 2023-04-06
docs: english version of application architecture

TanZiYen:


lumianph:
the same issue to the chinese version, refer to the pr of chinese version

lumianph:
the figures need to be translated to English @TanZiYen

lumianph (approved):

PR #3217, opened 2023-04-13
fix: clear-all

### LINKED ISSUE #3210: `clear-all.sh` script has not clear the hdd/ssd data directory
`clear-all.sh` script has not clear the hdd/ssd data directory. refer https://github.com/4paradigm/OpenMLDB/blob/v0.7.3/release/sbin/clear-all.sh#L37

dl239:
close #3210


release/sbin/clear-all.sh
```diff
@@ -35,6 +35,19 @@ rm_dir() {
...
+  conf_file="conf/tablet.flags.template"
+  dirname=""
+  while IFS= read -r line
+  do
+    if echo "$line" | grep -q '^#'; then
+      continue
+    fi
+    if echo "$line" | grep -v "zk_root_path" | grep -q "root_path" ||
+        echo "$line" |  grep -q "openmldb_log_dir"; then
+      cur_dirname=$(echo "${line}" | awk -F '=' '{print $2}')
```
aceforeverd: kind unsafe
```sh
echo 'root_path=/ abc' | awk -F '=' '{print $2}'
```

is `/` and `abc`, it will erase the whole system

aceforeverd: If possible, it should never runs under root

release/sbin/clear-all.sh
```diff
@@ -35,6 +35,26 @@ rm_dir() {
...
+      dirname="${dirname} ${cur_dirname}"
+    fi
+  done < "$conf_file"
+  for item in $dirname
```
aceforeverd: Don't do the dirty check. A common fix to that problem is double quoting the path, so `rm / abc` become `rm "/ abc"`. 
You can declare `dirname` as array, or manually concat with double quotes.

release/sbin/clear-all.sh
```diff
@@ -35,6 +35,18 @@ rm_dir() {
...
+  conf_file="conf/tablet.flags.template"
+  dirname=()
+  while IFS= read -r line
+  do
+    if echo "$line" | grep -q '^#'; then
+      continue
+    fi
+    if echo "$line" | grep -v "zk_root_path" | grep -q "root_path" ||
+        echo "$line" |  grep -q "openmldb_log_dir"; then
+      dirname[${#dirname[*]}]=$(echo "${line}" | awk -F '=' '{print $2}')
```
aceforeverd: dirname+=

release/sbin/clear-all.sh
```diff
@@ -48,8 +60,15 @@ else
       rm_dir "$host" "$dir"
     else
       echo "clear tablet data and log in $dir with endpoint $host:$port "
-      cmd="cd $dir && rm -rf recycle db logs"
-      run_auto "$host" "$cmd"
+      for item in "${dirname[@]}"
+      do
+        if [[ $item == "/" ]]; then
```
aceforeverd: there are many cases elsewhere like `/usr`, we can't hard check it. 

If the array `dirname` is constructed corrected, simply make the cmd as
```bash
cmd="cd $dir && rm -rf \"$item\""
```

aceforeverd (changes requested):

release/sbin/clear-all.sh
dl239: what dose it mean?

tobegit3hub (approved):
LGTM

release/sbin/clear-all.sh
aceforeverd: array concat: https://stackoverflow.com/a/1951523

release/sbin/clear-all.sh
dl239: `/` is protected by `preserve-root` option default.

aceforeverd (approved):

PR #3220, opened 2023-04-14
fix: add deployment usage in java quickstart

### LINKED ISSUE #2920: add example for CallablePreparedStatement
CallablePreparedStatement -> RequestPreparedStatement -> PreparedStatement
It can do
```
stmt = create()
for i:
  stmt.setXX()
  stmt.executeQuery() # in dataBuild, it'll clearParameters()
```

- [x] Add a best practice and ut for it.

It's not good to create the stmt every time, cuz ctor is a little complex.

https://github.com/4paradigm/OpenMLDB/blob/f349d7242b1f08ac3b467bdbc2deb507ab73883e/java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/CallablePreparedStatement.java#L31

vagetablechicken:
close #2920


Makefile
```diff
@@ -148,19 +148,19 @@ thirdparty-fast:
 	    if [ "$$new_zetasql_version" != "$(ZETASQL_VERSION)" ] ; then \
 		echo "[deps]: thirdparty up-to-date. reinstall zetasql from $(ZETASQL_VERSION) to $$new_zetasql_version"; \
 		$(MAKE) thirdparty-configure; \
-		$(CMAKE_PRG) --build $(THIRD_PARTY_BUILD_DIR) --target zetasql; \
+		$(CMAKE_PRG) --build $(THIRD_PARTY_BUILD_DIR) -j $(NPROC) --target zetasql; \
```
aceforeverd, line 151: does NPROC improve speed ?  this is the parallelism of build all thirdparty deps, parallelism of each dep is handled inside each `cmake/Fetch*.cmake`

vagetablechicken: Yes, speed faster. I can't find -j or other options in Fetch*.cmake, where?

dl239 (approved):

Makefile
aceforeverd: zetasql uses bazel, which has its own parallelism strategy  builtin, no need to specify NPROC ?

vagetablechicken: I can’t recall about zetasql, I'll revert this change, and check it when I build the old stdc++ capable image.

Makefile
```diff
@@ -148,7 +148,7 @@ thirdparty-fast:
...
```
aceforeverd, line 151: revert here as well

aceforeverd (approved):

Makefile
vagetablechicken: already done https://github.com/4paradigm/OpenMLDB/pull/3220/commits/631f6eaddbc4d744cfe03b0405fa6aa921d9c77f, but seems we can't see it

PR #3226, opened 2023-04-21
refactor: use `CreateTask` function to create all kinds of task

dl239:


src/nameserver/name_server_impl.h
```diff
@@ -467,135 +450,26 @@ class NameServerImpl : public NameServer {
...
-
-    std::shared_ptr<Task> CreateSendSnapshotTask(const std::string& endpoint, uint64_t op_index,
-                                                 ::openmldb::api::OPType op_type, uint32_t tid, uint32_t remote_tid,
-                                                 uint32_t pid, const std::string& des_endpoint);
-
-    std::shared_ptr<Task> CreateLoadTableTask(const std::string& endpoint, uint64_t op_index,
-                                              ::openmldb::api::OPType op_type, const std::string& name, uint32_t tid,
-                                              uint32_t pid, uint32_t seg_cnt, bool is_leader,
-                                              ::openmldb::common::StorageMode storage_mode);
+    std::shared_ptr<Task> CreateTask(const std::shared_ptr<TaskMeta>& task_meta);
```
aceforeverd, line 453: can simplified with a template wrapper,  checkout https://github.com/4paradigm/OpenMLDB/blob/0920bf256d10a1c578a1da7ed7bfdb17d23efab7/hybridse/include/node/node_manager.h#L64-L68

dl239: `CreateTask` function is not only creating Task object, but also doing some init works like binding different execute function. 
Absolutely, using template can simplify the code of calling `CreateTask` instead of using shared_ptr. The downside is having to move the implement(much more code) to the header file

aceforeverd: You can make a template wrapper over the current `CreateTask` in header, no need to move impls into header

dl239: will update in #3288 later

aceforeverd (approved):

vagetablechicken (approved):

PR #3237, opened 2023-04-26
feat(docs): add the default html page under root of doc site

aceforeverd:
Then `{page_url}/` will automatically redirect to `{page_url}/zh`.


vagetablechicken (approved):

lumianph (approved):

lumianph:
we may add an index page showing two different entries for the Chinese and English docs later.

PR #3241, opened 2023-04-28
feat: add `skip_index_check` option

dl239:
if we deploy a SQL contains `LAST JOIN`, `kLatestTime` index is required. if there is another index type like `kAbsoluteTIme`, it will deploy failed. So we add `skip_index_check` option to skip index checking


vagetablechicken (approved):

tobegit3hub (approved):
LGTM

PR #3312, opened 2023-06-09
feat: add other os build workflow

### LINKED ISSUE #3302: build cxx binary/so for lower glibc version or other os
If released binary/so runs on lower glibc host, it'll get error:
```
libstdc++.so.6: version `GLIBCXX_3.4.19' not found
```

- [ ] it's better to fix the exception throw in LibraryLoader, so that we can see the better error msg.

We check the host system lib by script
```
#!/bin/bash
set -x
cat /etc/redhat-release
ldd --version
strings /lib64/libc.so.6 | grep ^GLIBC_
strings /usr/lib64/libstdc++.so.6 |grep ^GLIBC
```

For example, if we want run binary/so on glibc 2.12 and libstdc++.so.6.0.13(GLIBCXX_3.4.13), we can build it on a old centos image.

Build steps:
1. use old image, `ghcr.io/4paradigm/centos6_gcc7_hybridsql`
2. yum install -y bison bison-devel byacc cppunit-devel patchdevtoolset-8-gcc devtoolset-8-gcc-c++, install build tools.
3. create `/etc/os-release` for zetasql
4. rocksdb don't build sst_dump&ldb, or add -lrt
5. Download https://github.com/sub-mod/bazel-builds/releases for zetasql, build_zetasql_parser.sh直接关了bazel test? add -lrt
6. add func in script(centos6 coreutils doesn't have `realpath`)
```
realpath () { 
[[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}" 
}
```
Better to add a new Dockerfile? The repo is private and we should modify zetasql build scripts, so add a script for patch is better. If we want a new Dockerfile, thirdparty build scripts need update.

We can test the built files, e.g. `strings libsql_jsdk.so | grep ^GLIBC`

vagetablechicken:
close #3302 
workflow for centos6, macos 10/11, just dispatch trigger
build script for centos6

### DIFF
--- a/.github/workflows/other-os-build.yml
+++ b/.github/workflows/other-os-build.yml
@@ -1,0 +1,256 @@
+# OpenMLDB build on other os, only trigger by dispatch event and only upload artifacts, not release.
+# You should download artifacts manually. And we won't do test in this workflow.
+name: other os build
+
+on:
+  workflow_dispatch:
+    inputs:
+      OS:
+        description: 'os name: centos6, macos11, macos10'
+        required: true
+        default: 'centos6'
+      JAVA_SDK:
+        description: 'java sdk enable: ON, OFF'
+        required: true
+        default: 'OFF' # turn on later
+      PYTHON_SDK:
+        description: 'python sdk enable: ON, OFF'
+        required: true
+        default: 'OFF'
+
+env:
+  GIT_SUBMODULE_STRATEGY: recursive
+  NPROC: 2 # default Parallel build number for GitHub's Linux runner
+  EXAMPLES_ENABLE: OFF # turn off hybridse's example code
+  HYBRIDSE_TESTING_ENABLE: OFF # turn off hybridse's test code
+
+jobs:
+  centos6: # glibc version is 2.12, glibcxx version is 3.4.13
+    if: ${{ github.event.inputs.OS == 'centos6' }}
+    runs-on: ubuntu-latest
+    env:
+      OS: linux
+      SQL_JAVASDK_ENABLE: ${{ github.event.inputs.JAVA_SDK }}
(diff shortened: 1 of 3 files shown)


.github/workflows/other-os-build.yml
```diff
@@ -0,0 +1,256 @@
...
+    if: ${{ github.event.inputs.OS == 'centos6' }}
+    runs-on: ubuntu-latest
+    env:
+      OS: linux
+      SQL_JAVASDK_ENABLE: ${{ github.event.inputs.JAVA_SDK }}
+      SQL_PYSDK_ENABLE: ${{ github.event.inputs.PYTHON_SDK }} # python whl will be built when make, no prerequirement
+      TESTING_ENABLE: OFF
+      NPROC: 8
+    steps:
+      - uses: actions/checkout@v3
```
aceforeverd, line 41: give the option  to build upon release tag, and tweak release artifact name

steps/centos6_build.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
```
dl239, line 1: add copyright

.github/workflows/other-os-build.yml
vagetablechicken: You mean a option `VERSION` to let user define the release version?

steps/centos6_build.sh
vagetablechicken: done

dl239 (approved):

.github/workflows/other-os-build.yml
vagetablechicken: done

aceforeverd (approved):

PR #3325, opened 2023-06-19
build(ci): reduce coverage disk usage

vagetablechicken:
- add coverage cmake for local test
- add option COVERAGE_NO_DEPS to separate make and run, and skip building unuse test
- rm in coverage workflow


cmake/CodeCoverage.cmake
aceforeverd, line 1: new file ? CodeCoverage.cmake is in the cmake-modules submodules

Makefile
```diff
@@ -46,6 +46,9 @@ endif
 ifdef COVERAGE_ENABLE
     OPENMLDB_CMAKE_FLAGS += -DCOVERAGE_ENABLE=$(COVERAGE_ENABLE)
 endif
+ifdef COVERAGE_NO_DEPS
+	OPENMLDB_CMAKE_FLAGS += -DCOVERAGE_NO_DEPS=$(COVERAGE_NO_DEPS)
```
aceforeverd: indent issue

src/sdk/CMakeLists.txt
```diff
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-if(TESTING_ENABLE)
+# coverage test won't run tests below, but they are *_test, so if COVERAGE_NO_DEPS, disable build
+if(TESTING_ENABLE AND NOT COVERAGE_NO_DEPS)
```
aceforeverd, line 16: it do not update the `test_list` variable, why coverage pull the targets in sdk ?

This commit introduce : https://github.com/4paradigm/OpenMLDB/blob/10721cd05540ba51f3c0eb309bca819cf9832eb8/src/CMakeLists.txt#L98 ?

CMakeLists.txt
```diff
@@ -285,10 +286,15 @@ link_directories(
 add_subdirectory(src)
 
 if (COVERAGE_ENABLE AND CMAKE_COMPILER_IS_GNUCXX)
-  # FIXME(aceforeverd): api_server_test failed to run on debug mode, skip temporarily
-  SETUP_TARGET_FOR_COVERAGE_LCOV(NAME coverage
-    EXECUTABLE ctest --force-new-ctest-process --output-on-failure -E "api_server_test"
-    DEPENDENCIES ${test_list})
+  if (COVERAGE_NO_DEPS)
+        SETUP_TARGET_FOR_COVERAGE_LCOV(NAME coverage
+        EXECUTABLE ctest --force-new-ctest-process --output-on-failure -E "api_server_test")
```
aceforeverd, line 292: by design, `make coverage-cpp` only runs `coverage` target and should pull dependent targets `coverage` job depends.

cmake/CodeCoverage.cmake
vagetablechicken: cmake 3.21/3.20 don't have it, which version you use?

src/sdk/CMakeLists.txt
vagetablechicken: I split build and run in workflow:
`make first 60 tests, clean .o/... files, make other tests, clean, run tests`, we should know all test target names, just grep the target name end with _test, so the tests here is including, and hard to filter, so disable in cmake.

CMakeLists.txt
vagetablechicken: But we delete files in build, if we run make again, it'll build again. COVERAGE_NO_DEPS can ignore it.

cmake/CodeCoverage.cmake
vagetablechicken: git submodule update will download it

src/sdk/CMakeLists.txt
vagetablechicken: leave it

aceforeverd (approved):

lqy222 (approved):

PR #3328, opened 2023-06-20
fix: refresh table info after zk updated

vagetablechicken:
cicd workflow cpp job failed on deploy, cuz tablet may overwrite index by old zk info(slow notify, and no version, so it replace new), so make tablet sync zk info after zk info update.

### DIFF
--- a/src/nameserver/name_server_impl.cc
+++ b/src/nameserver/name_server_impl.cc
@@ -17,16 +17,15 @@
 #include "nameserver/name_server_impl.h"
 
 #include <algorithm>
-#include <set>
-#include <random>
-#include <iterator>
 #include <iostream>
+#include <iterator>
+#include <random>
+#include <set>
 #include <vector>
 
-
+#include "absl/strings/numbers.h"
 #include "absl/strings/str_cat.h"
 #include "absl/strings/str_split.h"
-#include "absl/strings/numbers.h"
 #include "absl/time/time.h"
 #include "nameserver/system_table.h"
 #include "sdk/db_sdk.h"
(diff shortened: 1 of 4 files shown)


src/nameserver/name_server_impl.cc
```diff
@@ -8584,6 +8582,22 @@ bool NameServerImpl::AddIndexToTableInfo(const std::string& name, const std::str
         }
     }
     UpdateZkTableNode(table_info);
+    // refresh tablet here, cuz this func may be called by task
```
dl239, line 8585: get all tablet by `GetAllHealthTablet` and call refresh

vagetablechicken: tid is just one, I'll fix it, and filter tablet by health and exists in table info. GetAllHealthTablet has a lock in it, so just filter health in here

src/test/util.cc
```diff
@@ -182,15 +182,15 @@ void ProcessSQLs(sdk::SQLRouter* sr, std::initializer_list<absl::string_view> sq
     hybridse::sdk::Status status;
     for (auto sql : sqls) {
         sr->ExecuteSQL(std::string(sql), &status);
-        EXPECT_TRUE(status.IsOK()) << "running sql=" << sql << " failed: code=" << status.code << ", msg=" << status.msg
+        ASSERT_TRUE(status.IsOK()) << "running sql=" << sql << " failed: code=" << status.code << ", msg=" << status.msg
```
aceforeverd: we need `EXPECT*`, some context require all SQLs executed even failure

aceforeverd (changes requested):

src/test/util.cc
vagetablechicken: OK, but it's not a good idea to expect when init(e.g. create db/table, deploy...), if create failed, the log will be too long and may confuse us.

aceforeverd (approved):

dl239 (approved):

PR #3337, opened 2023-06-28
fix: fix alter table with db name and add ut

tobegit3hub:
* Resolve the incorrect code in https://github.com/4paradigm/OpenMLDB/blob/6bb57489f2072f4f3a669892104909ce402f8451/src/sdk/sql_cluster_router.cc#L2651 .

### DIFF
--- a/src/sdk/sql_cluster_router.cc
+++ b/src/sdk/sql_cluster_router.cc
@@ -2648,7 +2648,7 @@ std::shared_ptr<hybridse::sdk::ResultSet> SQLClusterRouter::ExecuteSQL(
 
             std::string db_name;
             if (!plan->db_.empty()) {
-                db_name = (plan->table_.begin(), plan->table_.end());
+                db_name = plan->db_;
             } else {
                 db_name = db;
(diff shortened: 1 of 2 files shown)


dl239 (approved):

src/sdk/sql_cluster_router.cc
```diff
@@ -2648,7 +2648,7 @@ std::shared_ptr<hybridse::sdk::ResultSet> SQLClusterRouter::ExecuteSQL(
 
             std::string db_name;
             if (!plan->db_.empty()) {
-                db_name = (plan->table_.begin(), plan->table_.end());
+                db_name = plan->db_;
```
aceforeverd: `db_name = {plan->db_.begin(), plan->db_.end()};`

tobegit3hub: okay

aceforeverd: never mind

aceforeverd (approved):

PR #3350, opened 2023-07-10
fix: external udf load

vagetablechicken:


hybridse/src/udf/dynamic_lib_manager.cc
```diff
@@ -44,9 +45,13 @@ base::Status DynamicLibManager::ExtractFunction(const std::string& name, bool is
         }
     }
     if (!so_handle) {
-        void* handle = dlopen(file.c_str(), RTLD_LAZY);
-        CHECK_TRUE(handle != nullptr, common::kExternalUDFError,
-                   "can not open the dynamic library: " + file + ", error: " + dlerror())
+        // use abs path to avoid dlopen search failed in yarn mode 
+        char abs_path_buff[PATH_MAX];
+        if (realpath(file.c_str(), abs_path_buff) == NULL) {
```
vagetablechicken, line 50: check realpath cover symbolic path?

vagetablechicken: no need, we just use abs to dlopen, handle_map_ store the file(relative, and it's ok to be a symlink)

aceforeverd (approved):

dl239 (approved):

PR #3359, opened 2023-07-12
fix(udf): try abs path after dlopen failed and more error catch

vagetablechicken:


tobegit3hub (approved):
LGTM

hybridse/src/udf/dynamic_lib_manager.cc
```diff
@@ -45,14 +45,25 @@ base::Status DynamicLibManager::ExtractFunction(const std::string& name, bool is
...
     }
     if (!so_handle) {
-        // use abs path to avoid dlopen search failed in yarn mode
-        char abs_path_buff[PATH_MAX];
-        if (realpath(file.c_str(), abs_path_buff) == NULL) {
-            return {common::kExternalUDFError, "can not get real path " + file};
+        void* handle = dlopen(file.c_str(), RTLD_LAZY);
+        if (handle == nullptr) {
+            std::string err_msg;
+            err_msg = "can not open the dynamic library: " + file + ", error: " + dlerror() + ", try to use abs path";
```
aceforeverd, line 51: dlerror may return NULL, simple wrap it !

hybridse/src/udf/dynamic_lib_manager.cc
```diff
@@ -45,14 +45,25 @@ base::Status DynamicLibManager::ExtractFunction(const std::string& name, bool is
...
+            LOG(WARNING) << err_msg;
+            // try to use abs path to avoid dlopen failed but it only works in the same path(e.g. in yarn mode)
+            char abs_path_buff[PATH_MAX];
```
aceforeverd, line 54: reset `errno` to `0` here

hybridse/src/udf/dynamic_lib_manager.cc
vagetablechicken: https://linux.die.net/man/3/dlopen dlsym may be null, so it needs to call dlerror before, but dlopen is ok, see the example

hybridse/src/udf/dynamic_lib_manager.cc
vagetablechicken: https://linux.die.net/man/3/realpath
`If there is no error, realpath() returns a pointer to the resolved_path.

Otherwise, it returns a NULL pointer, the contents of the array resolved_path are undefined, and errno is set to indicate the error.`

set 0 before realpath is meaningless

aceforeverd (approved):

PR #3362, opened 2023-07-13
fix: external udf error handle

### LINKED ISSUE #3356: external udf error handle
drop func -> nameserver call tablets drop, return if error -> tablets drop func, meaningless error msgs -> nameserver rm it in zk

If error in the middle, we can't drop the func again.

Workaround way:
1. rm function in zk
2. restart nameserver and taskmanagers，tablets are able to re-create func.

vagetablechicken:
closes #3356
create function error handle
drop function only fails on zk failed or not exsits. Even drop function failed on tablet or taskmanager, won't fail the whole op.(tablet and taskmanager can create udf again when it has the udf, just replace)

P.S. tablet RegisterDynamicUdf will RemoveDynamicUdf first if exists. Taskmanager only has a map, just replace, batch will use zk meta, taskmanager only pass the file to spark.


hybridse/src/udf/udf_library.cc
```diff
@@ -256,29 +256,29 @@ Status UdfLibrary::RemoveDynamicUdf(const std::string& name, const std::vector<n
...
     }
     if (!HasFunction(canonical_name)) {
-        return Status(kCodegenError, "can not find the function " + canonical_name);
+        return Status(kCodegenError, "can not find the function in udf table: " + canonical_name);
     }
     if (IsUdaf(canonical_name)) {
         std::lock_guard<std::mutex> lock(mu_);
         if (table_.erase(canonical_name) <= 0) {
-            return Status(kCodegenError, "can not find the function " + canonical_name);
+            return Status(kCodegenError, "udaf function not present in udf table: " + canonical_name);
```
aceforeverd, line 264: continue instead of return better ?

vagetablechicken: It's to overwrite udf in the next time, so if failed, just report it, no need to hide

src/nameserver/name_server_impl.cc
```diff
@@ -9733,16 +9729,16 @@ void NameServerImpl::CreateFunction(RpcController* controller, const CreateFunct
...
     std::vector<std::shared_ptr<TabletInfo>> succ_tablets;
+    std::string error_msgs;
     for (const auto& tablet : tablets) {
         std::string msg;
         if (!tablet->client_->CreateFunction(request->fun(), &msg)) {
-            PDLOG(WARNING, "create function failed. endpoint %s, msg %s",
-                    tablet->client_->GetEndpoint().c_str(), msg.c_str());
-            response->set_msg(msg);
+            error_msgs.append("create function failed on " + tablet->client_->GetEndpoint() + ", reason: " + msg + ";");
             break;
```
vagetablechicken: remove break

dl239 (approved):

hybridse/src/udf/udf_library.cc
aceforeverd: not hiding but make this function atomic. For udaf there are multiple states to be handled, so either delete nothing, or delete everything.

vagetablechicken: delete may be failed too, there's no everything. You try delele 3 maps, but every map deletetion could fail, I can't see any benifit to change the logic here.

tobegit3hub (approved):
LGTM

PR #3366, opened 2023-07-14
fix: api server write json

### LINKED ISSUE #3347: APIServer return incorrect result for boolean values
Create table like this.

```
create table fzzz_test.request (user_code STRING NOT NULL, material_id STRING NOT NULL, distance DOUBLE, delivery_ta DOUBLE,  system_time TIMESTAMP NOT NULL);
```

Deploy the following SQL.

```
select system_time, hour(system_time) as system_time_hour, distance, delivery_ta, (delivery_ta >= 80 or (distance >= 500 and distance <= 3000)) as delivery_flag from request
```

Test with the following data.

```
{ "input": [ [ "foo", "foo", 1.01, 1.01, 1635247427000 ] ] }
```

Focus on the output the last value and get `true` which is expected to `false`.

vagetablechicken:
closes #3347

### DIFF
--- a/src/apiserver/api_server_impl.cc
+++ b/src/apiserver/api_server_impl.cc
@@ -947,7 +947,7 @@ void WriteValue(JsonWriter& ar, std::shared_ptr<hybridse::sdk::ResultSet> rs, in
         }
         default: {
             LOG(ERROR) << "Invalid Column Type";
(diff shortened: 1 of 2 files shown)


tobegit3hub (approved):
LGTM

dl239 (approved):

PR #3373, opened 2023-07-18
build: install openmldb with rpath

aceforeverd:
LD_LIBRARY_PATH is evil, removing. Replace with the
- `$ORIGIN` token for linux
- or `@rpath` for darwin

Other fixes:
- tests improves to openmldb-jdbc
- testing `CREATE FUNCTION` in openmldb-jdbc
- improve onebox scripts with pidfile
- maven-commit-id-plugin: default to offline mode


vagetablechicken:
Doc about udf dir should fix https://openmldb.ai/docs/zh/main/openmldb_sql/udf_develop_guide.html?highlight=udf#id9

And I strongly recommend this pr don't be released before it has enough tests(integration). We should add tests asap

aceforeverd:
> Doc about udf dir should fix https://openmldb.ai/docs/zh/main/openmldb_sql/udf_develop_guide.html?highlight=udf#id9
> 
> And I strongly recommend this pr don't be released before it has enough tests(integration). We should add tests asap

No worry, I'll add tests in this PR. Original `LD_LIBRARY_PATH` is never consider stable, since it was never tested, from CICD

vagetablechicken:
> > Doc about udf dir should fix https://openmldb.ai/docs/zh/main/openmldb_sql/udf_develop_guide.html?highlight=udf#id9
> > And I strongly recommend this pr don't be released before it has enough tests(integration). We should add tests asap
> 
> No worry, I'll add tests in this PR. Original `LD_LIBRARY_PATH` is never consider stable, since it was never tested, from CICD

Very Good! Original `LD_LIBRARY_PATH` at least works on user's production env and no bug report now, so it could be a stable one. I don't want too many changes all in one version.

FTR We'll consider 0.8.2 as a stable version about udf and hive, if new version get error, test on 0.8.2

aceforeverd:
openmldb-batch test fails on mac:
> ERROR: something wrong with flag 'logtostderr' in file 'src/logging.cc'.  One possibility: file 'src/logging.cc' is being linked both statically and dynamically into this executable.

aceforeverd:
TODO:
- [ ] ~~support full const project in offline, #3405~~ replace test with a row project SQL
- [ ] P1 fix running java tests on Mac, depends #3400, #3407

PR #3376, opened 2023-07-18
feat: support const project online

### LINKED ISSUE #2759: feat: support const project in request mode
`select 1, func(12)` 
without table source should be fine in request mode, no performance reason to disable

### LINKED ISSUE #3311: Fail to validate SQL with db name and "SELECT 100"
The SQLs like "SELECT 100" will fail to validate.

b58d4feeff1d98fb7a3531bc0a48a898

The SQL without db name like "SELECT * from entities" may work but "SELECT * from SYSTEM_FEATURE_PLATFORM.entities" does not.

aceforeverd:
close #2759, #3311

### DIFF
--- a/hybridse/src/testing/engine_test_base.h
+++ b/hybridse/src/testing/engine_test_base.h
@@ -429,6 +429,12 @@ class BatchRequestEngineTestRunner : public EngineTestRunner {
                 offset += row_num;
             }
         }
+
+        if (request_rows_.empty()) {
+            // batch request rows will empty for const projects
+            // workaround by add the one empty row, which passes batch request tests for const project
+            request_rows_.push_back(Row());
+        }
         return Status::OK();
     }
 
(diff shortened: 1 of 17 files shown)


hybridse/src/testing/engine_test_base.h
```diff
@@ -271,14 +271,20 @@ class RequestEngineTestRunner : public EngineTestRunner {
...
+        }
         for (auto in_row : request_rows_) {
             Row out_row;
             int run_ret = request_session->Run(in_row, &out_row);
             if (run_ret != 0) {
                 return_code_ = ENGINE_TEST_RET_EXECUTION_ERROR;
                 return Status(common::kRunError, "Run request session failed");
             }
-            if (!has_batch_request) {
+            if (!has_batch_request && !request_db_name.empty() && !request_name.empty()) {
```
aceforeverd: rm second cond

aceforeverd:
TODO check if there is: a runner requires non-empty requests, but empty request is allow

aceforeverd:
> ~~TODO check if there is: a runner requires non-empty requests, but empty request is allow~~

TODO: support const project procedure. It is not fixed by this PR, since a request procedure requires a request table

vagetablechicken (approved):

lqy222 (approved):

PR #3388, opened 2023-07-21
feat: upgrade docker version

dl239:


python/openmldb_tool/setup.py
```diff
@@ -18,7 +18,7 @@
 
 setup(
     name='openmldb-tool',
-    version='0.7.0a0',
+    version='0.8.2',
```
aceforeverd: keep the `a0`, version number in main is pre-release, not final number

steps/prepare_release.sh
aceforeverd, line 1: If the change is about the py number, this change is unnecessary since https://github.com/4paradigm/OpenMLDB/pull/3388/files#r1270313819

dl239: change the version of CMAKE, python, java

tobegit3hub (approved):
LGTM

steps/prepare_release.sh
```diff
@@ -69,6 +72,9 @@ if [[ ${#ARR[@]} -gt 3 ]]; then
     # refer: https://www.python.org/dev/peps/pep-0440/#pre-releases
     PY_VERSION="${ARR[0]}.${ARR[1]}.${ARR[2]}${ARR[3]}"
 fi
+if [[ ${VERSION} =~ "SNAPSHOT" ]]; then
+    PY_VERSION="${ARR[0]}.${ARR[1]}.${ARR[2]}a0"
```
aceforeverd: conflict with line 73. I intend to run as :
```sh
./prepare_release.sh 1.2.3.a0
```

then the `PY_VERSION` is `1.2.3a0`. For cpp, `.a0` is not token; for java, `.a0` always replaced with `-SNAPSHOT`.  See the comment at top of the file. 

See if you require more semantic, or we can revert the change.

aceforeverd (approved):

PR #3392, opened 2023-07-24
fix: tablet rpc return, tool log, link https

vagetablechicken:


dl239 (approved):

src/tablet/tablet_impl.cc
```diff
@@ -3507,6 +3507,7 @@ void TabletImpl::GetTableFollower(RpcController* controller, const ::openmldb::a
     if (info_map.empty()) {
         response->set_msg("has no follower");
         response->set_code(::openmldb::base::ReturnCode::kNoFollower);
+        return;
```
aceforeverd, line 3510: what's the bug behavior like ?

vagetablechicken: no follower but return ok, rewrite msg and code by code below

aceforeverd (approved):

PR #3404, opened 2023-07-28
refactor: convert schema to `sdkSchema` to reduce `JNI` call

kewenkang:
This change add sdkSchema converted from schema and replace some JNI function call of schema with sdkSchema.

Currently, in checkIdx() and checkDataType() functions, JNI functions (such as GetColumnCnt() and GetColumnType()) are called.

> From our investigation, the GetColumnCnt() and GetColumnType() spend about 1/6 time of getXXX().

After this change, when creating an instance of SQLResultSet, schema will be converted to sdkSchema and later JNI function calls will be replaced with java function calls to reduce time.


dl239:
have you tested setting one row in preparestatement ? 
convertSchema will call four JNI functions with every filed. 
https://github.com/4paradigm/OpenMLDB/blob/main/java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/sdk/Common.java#L58-L62

dl239:
refer #3445

PR #3408, opened 2023-07-31
fix: kafka connector dependency

vagetablechicken:
openmldb-jdbc needs commons-io, so pack it


vagetablechicken:
Give a message sample in auto schema way, just k:v map

vagetablechicken:
And kafka server/broker 1.1.1 is compatible but if connect distributed, it requires auto create topic, needs to check about auto create topic option

aceforeverd (approved):

lqy222 (approved):

Issue #3403, opened 2023-07-27
Optimize `SQLRequestRow` and `RequestPreparedStatement`


dl239:
- uncache the value of base type in `RequestPreparedStatement`
- rm `stringsLen` from `RequestPreparedStatement`
- replace list by bitmap to check whether the field set
- optimize the check in PreparedStatement, such as `checkNull`

PR #3420, opened 2023-08-03
Hcml

1944462162:


aceforeverd:
What's is about ?

java/openmldb-jdbc/src/main/java/com/_4paradigm/openmldb/jdbc/RequestPreparedStatement.java
```diff
@@ -258,6 +258,32 @@ public void setTimestamp(int i, Timestamp timestamp) throws SQLException {
         currentDatas.set(i - 1, ts);
     }
 
+    public void setRow(Map<String, Object> data) throws SQLException {
```
dl239, line 261: Do not add methods that are not in jdbc. https://docs.oracle.com/javase/8/docs/api/java/sql/PreparedStatement.html. We are plan to  optimize the performance of SDK, refer https://github.com/4paradigm/OpenMLDB/issues/3403, thank you

PR #3456, opened 2023-08-23
feat: deploy with bias

### LINKED ISSUE #3428: sql deploy: allow customize auto-created indexes from user
# Problem

see #3409, deploy stmt by default, create the index(as well as the TTL definition) for a table, to fulfill the requirements of the online SQL, if index not yet exists. For example, SQL with `window (... rows_range 30d preceding and current row)` will set TTL to 30d.

The issue is, if a request sent with a ts older than current system time, query result may not correct, cause some of the rows for the request row has deleted according to the 30d expire TTL rule. 

User should have the ability to customize the TTL definition for their needs.  

# Solution proposed: 

## Two new deploy options

- range_bias: auto added range index will increased by
- rows_bias: auto added rows index will increased by

**Type**

- `integral | interval | string`: e.g. `*_bias = 1, *_bias = 1d, *_bias = 'inf'`

**Special values**
- `*_bias = 'inf'`: the TTL going to add is **Unlimited**

**Default value**
 `0`, which reserved backward compatibility for deploy stmts. 

Bias values always expects non-negative, a negative input, like `-1`, causes SQL error. 

e.g 
```sql
deploy
   select .... from t1 window w as (partition by pat order by ts rows_range 30d preceding and current row) 
   options ( range_bias = 1d)
```

will create the index with ttl `31d`, instead of `30d` previously.

### Drawbacks

- If there is multiple tables & indexes going to create in a single deploy stmt, two options may still not enough. User may customize themselves after deploy, with `alter table alter index` stmt.

## alter table alter index ?

P1

vagetablechicken:
closes #3428 

set bias in demo cases before release


docs/zh/openmldb_sql/deployment_manage/DEPLOY_STATEMENT.md
```diff
@@ -170,6 +175,16 @@ deploy demo options(SYNC="false") SELECT t1.col1, t2.col2, sum(col4) OVER w1 as
     WINDOW w1 AS (PARTITION BY t1.col2 ORDER BY t1.col3 ROWS BETWEEN 2 PRECEDING AND CURRENT ROW);
 ```
 
+#### 设置偏移
+
+如果你并不希望数据根据deploy的索引淘汰，或者希望晚一点淘汰，可以在deploy时设置偏移，常用于数据时间戳并不实时的情况、测试等情况。如果deploy后的索引ttl为abs 3h，但是数据的时间戳是3h前的(以系统时间为基准)，那么这条数据就会被淘汰，无法参与计算。设置一定时间或永久的偏移，则可以让数据更久的停留在在线表中。
+
+时间偏移，单位可以是`ms`、`s`、`m`、`h`、`d`，也可以是整数，单位为`ms`，也可以是`inf`，表示不限制；如果是行数偏移，可以是整数，单位是`row`，也可以是`inf`，表示不限制。0表示不偏移。
```
vagetablechicken: create table ttl can't be ms?

docs/en/reference/sql/deployment_manage/DEPLOY_STATEMENT.md
```diff
@@ -173,6 +178,16 @@ deploy demo options(SYNC="false") SELECT t1.col1, t2.col2, sum(col4) OVER w1 as
     WINDOW w1 AS (PARTITION BY t1.col2 ORDER BY t1.col3 ROWS BETWEEN 2 PRECEDING AND CURRENT ROW);
 ```
 
+#### BIAS
+
+If you don't want data to be expired by the deployed index, or want data expired later, you can set bias when deploy, which is usually used in the case of data timestamp is not real-time, test, etc. If the index ttl after deploy is abs 3h, but the data timestamp is 3h ago(based on system time), then the data will be eliminated and cannot participate in the calculation. Setting a certain time or permanent bias can make the data stay in the online table for a longer time.
+
+Range bias can be `ms`, `s`, `m`, `h`, `d`, or integer(unit is ms), or `inf`. Rows bias can be integer, or `inf`. 0 means no bias.
```
vagetablechicken: inf = infinite

docs/zh/openmldb_sql/deployment_manage/DEPLOY_STATEMENT.md
vagetablechicken: 不限制->永不淘汰

src/sdk/sql_cluster_router.cc
```diff
@@ -3588,13 +3611,18 @@ hybridse::sdk::Status SQLClusterRouter::GetNewIndex(const TableInfoMap& table_ma
             // the existed ones
             std::vector<::openmldb::common::ColumnKey> new_indexs;
             for (auto& column_key : extract_column_keys) {
-                auto index_id = openmldb::schema::IndexUtil::GetIDStr(column_key);
+                // add bias on extracted index, 0 means unbounded
+                auto new_column_key = bias.AddBias(column_key);
+                LOG(WARNING) << "add bias on index " << column_key.ShortDebugString() << " to "
```
vagetablechicken: log info

aceforeverd: hide log ? in case printed in cli

vagetablechicken: will convert to info

docs/zh/openmldb_sql/deployment_manage/DEPLOY_STATEMENT.md
vagetablechicken: > create table ttl can't be ms?

ref `ColumnIndexNode`, interval has no ms

dl239 (approved):

lqy222 (approved):

src/sdk/sql_cluster_router.cc
```diff
@@ -4392,5 +4420,65 @@ std::shared_ptr<hybridse::sdk::ResultSet> SQLClusterRouter::GetNameServerJobResu
     return rs;
 }
 
+common::ColumnKey Bias::AddBias(const common::ColumnKey& index) const {
+    if (!index.has_ttl()) {
+        LOG(WARNING) << "index has no ttl, skip bias";
```
aceforeverd, line 4425: this log should hide too ?

vagetablechicken: Just protection, highly unlikely ever to happen.

PR #3547, opened 2023-10-11
refactor(codegen): null safe for struct ir builder 

### LINKED ISSUE #926: Codegen: NativeValue represent const null safely

OpenMLDB generates IR for expressions using LLVM SDK. We use `NativeValue` to represent an LLVM value with a null flag.
For instance, we can create a Const Null of given type by calling:
```c++
NativeValue::CreateNull(type);
```
`CreateNull` essentially invoke:
```c++
NativeValue NativeValue::CreateNull(::llvm::Type* ty) {
    return NativeValue(nullptr, nullptr, ty);
}
```
Basically, a const null will be represented by a NativeValue with null raw `llvm::Value*` and a null flag.

When we try to obtain a Value from the NativeValue, the ConstNull NativeValue will return a UndefineValue instead.
```c++
::llvm::Value* NativeValue::GetValue(::llvm::IRBuilder<>* builder) const {
    if (IsConstNull()) {
        return ::llvm::UndefValue::get(GetType());
    } else if (IsMem()) {
        return builder->CreateLoad(raw_);
    } else {
        return raw_;
    }
}
```

The problem is when we try to construct an String type NULL, we will create a Native Value with undefined StringRef. This means the StringRef.data and StringRef.size are undefined. This is very dangerous.

In this issue, we are going to improve the Const Null Native Value for Date, StringRef and Timestamp. Ensure they can be used safely.

- [ ] Introduce a CreateNull method for StringIrBuilder, TimestampIRBuilder, DateIRBuilder. The `NativeValue` of a const null should be initialized with safe data and pointer. For instance, they should be initialized as `StringRef(nullptr, 0)`, `Date(0)`, `Timestamp(0)`
- [ ] Review Date, Timestamp, StringRef functions, make sure they access these structures safely.

- [ ] Handling NULL when it is necessary. 
- [ ] Add introduction and description in built-in functions development guide.

**Additional context**

aceforeverd:
fixes https://github.com/4paradigm/OpenMLDB/issues/926

### DIFF
--- a/cases/query/const_query.yaml
+++ b/cases/query/const_query.yaml
@@ -126,15 +126,55 @@ cases:
       columns: ['c1 bool', 'c2 int16', 'c3 int', 'c4 double', 'c5 string', 'c6 date', 'c7 timestamp' ]
       rows:
         - [ true, 3, 13, 10.0, 'a string', '2020-05-22', 1590115420000 ]
+
+  # =================================================================================
+  # Null safe for structure types: String, Date, Timestamp and Array
+  # creating struct from:
+  # 1. NULL liternal (const null)
+  # 2. another supported date type but fails to cast, e.g. timestamp(-1) returns NULL of timestamp
+  #
+  # casting to array type un-implemented
+  # =================================================================================
   - id: 10
+    desc: null safe for date
     mode: procedure-unsupport
     sql: |
       select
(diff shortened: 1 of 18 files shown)


lqy222 (approved):

tobegit3hub (approved):
LGTM

PR #3554, opened 2023-10-13
feat(sql): window without order by

aceforeverd:
Rules:
- ALLOWED for ROWS-type WINDOW
- NOT ALLOWED for RANGE-type WINDOW with offset PRECEDING/FOLLOWING frame
- NOT ALLOWED for WINDOW with attribute EXCLUDE CURRENT_TIME

Usage notice for WINDOW without ORDER BY clause:
- Without ORDER BY, rows are processed in an unspecified order, neither which rows are going to included in the window or the ordering among rows
- General-Purpose Window Functions, e.g `lag, first_value, ...`, depend on the sort ordering specified by the ORDER BY clause of the associated window definition. In which case, MAY returns undetermined result
- For batch mode, the query result is undetermined, for request mode, the request row is ensured to be **BEFORE OR THE LAST ROW** from window frame

### DIFF
--- a/cases/function/window/error_window.yaml
+++ b/cases/function/window/error_window.yaml
@@ -17,15 +17,17 @@ debugs: []
 version: 0.5.0
 cases:
   - id: 0
-    desc: no order by
+    desc: RANGE-type WINDOW with offset PRECEDING/FOLLOWING requires ORDER BY
     inputs:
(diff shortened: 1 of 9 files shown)


aceforeverd:
TODO afterwards: fix auto-create-index logic during deployment, I don't cover that.

dl239 (approved):

lqy222 (approved):

Issue #3416, opened 2023-08-02
pure virtual method called


vagetablechicken:
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007fa21d102537 in __GI_abort () at abort.c:79
#2  0x00007fa21d4ac7ec in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007fa21d4b7966 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007fa21d4b79d1 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007fa21d4b869f in __cxa_pure_virtual () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x0000557144278b05 in brpc::Controller::OnVersionedRPCReturned (this=0x7fa1f0053fd0, info=..., new_bthread=<optimized out>, saved_error=<optimized out>) at /home/ace/deps/build/src/brpc/src/brpc/controller.cpp:637
#7  0x000055714427929e in brpc::Controller::HandleSendFailed (this=0x7fa1f0053fd0) at /home/ace/deps/build/src/brpc/src/brpc/controller.cpp:1007
#8  0x0000557144277df8 in brpc::Controller::IssueRPC (this=0x7fa1f0053fd0, start_realtime_us=<optimized out>) at /home/ace/deps/build/src/brpc/src/brpc/controller.cpp:1159
#9  0x0000557144270832 in brpc::Channel::CallMethod (this=0x55714827e940, method=0x5571481e5b60, controller_base=0x7fa1f0053fd0, request=0x7fa1f0054330, response=0x7fa1f00543b0, done=0x0) at /home/ace/deps/build/src/brpc/src/brpc/channel.cpp:576
#10 0x00005571432939c4 in openmldb::api::TabletServer_Stub::GetTaskStatus (this=0x55714b60b320, controller=0x7fa1f0053fd0, request=0x7fa1f0054330, response=0x7fa1f00543b0, done=0x0) at src/proto/tablet.pb.cc:46385
[...]

PR #3557, opened 2023-10-17
build: upgrade thirdparty to 0.6.0

aceforeverd:
1. upgrade thirdparty, which upgrade brpc to 1.6.0
2. calls some teardown function before nameserver destructor, it mitigate cases like #3416


src/nameserver/name_server_impl.cc
```diff
@@ -5004,19 +5006,21 @@ void NameServerImpl::UpdateTableStatus() {
...
-        if (!kv.second->client_->GetTableStatus(tablet_status_response)) {
-            PDLOG(WARNING, "get table status failed! endpoint[%s]", kv.first.c_str());
-            continue;
-        }
-        for (int pos = 0; pos < tablet_status_response.all_table_status_size(); pos++) {
-            std::string key = absl::StrCat(tablet_status_response.all_table_status(pos).tid(), "_",
-                              tablet_status_response.all_table_status(pos).pid(), "_", kv.first);
-            pos_response.emplace(key, tablet_status_response.all_table_status(pos));
+
+        for (const auto& kv : tablet_ptr_map) {
```
dl239: what's the reason for the update?

tobegit3hub (approved):
LGTM

src/nameserver/name_server_impl.cc
aceforeverd: see #3416. I just forward the workaround previously, without it `single_tablet_test` will not teardown correctly. 
PLS check if there are any side effects

dl239: there are some problems if send rpc request under the lock

aceforeverd: reverted

dl239 (approved):

PR #3559, opened 2023-10-18
feat: full inspect

vagetablechicken:
1. inspect for full cluster report
2. only make rpc hint to be optional(need extra lib pb and bs4), we can call rpc in base version
3. docs about full inspect
4. docs about diag tool commands and add inspect in must read

### DIFF
--- a/python/openmldb_tool/diagnostic_tool/diagnose.py
+++ b/python/openmldb_tool/diagnostic_tool/diagnose.py
@@ -83,7 +83,7 @@ def status(args):
 
     # --diff with dist conf file, conf_file is required
     if args.diff:
-        assert flags.FLAGS.conf_file, "need --conf_file/-f"
+        assert flags.FLAGS.conf_file, "need -f,--conf_file"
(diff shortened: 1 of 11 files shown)


python/openmldb_tool/diagnostic_tool/diagnose.py
```diff
@@ -96,39 +98,60 @@ def status(args):
...
+    # we use `show table status` instead of partition inspection, cuz it's simple and quick
+    warn_tables = table_ins(connect)
+
+    # if has unhealthy tables, do partition and ops check, otherwise skip
+    hints = []
+    if warn_tables:
+        # 3. ns ops
+        related_ops = ops_ins(connect)
+        # 4. partition level and get some hint about table
+        hints = partition_ins(server_map, related_ops)
```
vagetablechicken: check offset of healthy tables

aceforeverd (approved):

lqy222 (approved):

PR #3565, opened 2023-10-23
feat: last join on window

aceforeverd:
Supporting online for this kind of SQLs:
```sql
SELECT ... FROM request
LAST JOIN (
  SELECT ID, count(val) over w as agg
  FROM t2 
  WINDOW w as ( PARTITION BY part ... )
) tx ON request.ID = tx.ID
```

**RESTRICTIONS**

1. The JOIN condition, `tx.ID`, where ID must resolve to the column from actual table, which is `t2` in the SQL above, not from the window agg result
2. Table t2 should at least have two indexes, one index key referred by `tx.ID`, another index key referred by WINDOW PARTITION BY column

### DIFF
--- a/hybridse/src/plan/planner.cc
+++ b/hybridse/src/plan/planner.cc
@@ -272,7 +272,7 @@ base::Status Planner::CreateSelectQueryPlan(const node::SelectQueryNode *root, n
         auto first_window_project = dynamic_cast<node::ProjectListNode *>(project_list_vec[1]);
         node::ProjectListNode *merged_project =
             node_manager_->MakeProjectListPlanNode(first_window_project->GetW(), true);
-        if (!is_cluster_optimized_ && !enable_batch_window_parallelization_ &&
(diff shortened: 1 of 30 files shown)


hybridse/src/plan/planner.cc
```diff
@@ -272,7 +272,8 @@ base::Status Planner::CreateSelectQueryPlan(const node::SelectQueryNode *root, n
         auto first_window_project = dynamic_cast<node::ProjectListNode *>(project_list_vec[1]);
         node::ProjectListNode *merged_project =
             node_manager_->MakeProjectListPlanNode(first_window_project->GetW(), true);
-        if (!is_cluster_optimized_ && !enable_batch_window_parallelization_ &&
+        // TODO(ace): make the split in the optimizer, like clustrer optimzed
```
aceforeverd: nop

lqy222 (approved):

dl239 (approved):

Issue #3615, opened 2023-11-21
support SQL in offline mode


aceforeverd:
support those features in offline mode #3576 #3565 #3533 #3554 #3590

PR #3619, opened 2023-11-22
feat(offline): support & test new SQLs from online mode

aceforeverd:
- fix WINDOW without ORDER BY in batch mode
- support WINDOW without ORDER BY offline, only work if SkewWindowOpt is off 
- support last join (window/last join) offline

Ref https://github.com/4paradigm/OpenMLDB/issues/3615


tobegit3hub (approved):
LGTM

vagetablechicken (approved):

PR #3627, opened 2023-11-24
test(#3544): cluster opt default on

### LINKED ISSUE #3544: make `enable_distsql` defaults on
tablet server uses this flag `enable_distsql` to enable cluster optimization https://github.com/4paradigm/OpenMLDB/blob/b685e648e4467d86ce1ad8233b9428fb50313e09/src/tablet/tablet_impl.cc#L210-L214. It is on with release configurations: `release/conf/tablet.flags.template`, but no in `src/flags.cc`. Which as a results, won't enabled in mini_cluster tests, e.g `sql_sdk_test`.

There is a PR previously #3295, but turns out breaks existing tests. It means there exists cases that is OK with request mode, but not request(cluster) mode.

#3531 is a example, deployment tests not fail for those tests, simply because `enable_distsql` not turned on.

aceforeverd:
close #3544


aceforeverd:
hard to work around. batch mode tests require clusteropt off but request mode tests require clusteropt on

PR #3630, opened 2023-11-27
feat(sql router): add interface for sql to dag

aceforeverd:
```cpp
DAG SQLToDAG(const std::string& sql, hybridse::sdk::Status* status);
```

### DIFF
--- a/src/sdk/sql_router.h
+++ b/src/sdk/sql_router.h
@@ -80,6 +80,22 @@ class ExplainInfo {
     virtual const std::string& GetRequestDbName() = 0;
 };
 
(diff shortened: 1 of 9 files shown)


src/sdk/sql_router.h
```diff
@@ -80,6 +80,22 @@ class ExplainInfo {
     virtual const std::string& GetRequestDbName() = 0;
 };
 
+struct DAG {
```
aceforeverd: -> `DAGNode`

vagetablechicken (approved):

dl239 (approved):

PR #3631, opened 2023-11-28
feat: put returns failure if the server memory usage exceeds the specified ratio

### LINKED ISSUE #3591: Importing a large volume of data may lead to server out-of-memory (OOM)

dl239:


src/tablet/tablet_impl.cc
```diff
@@ -437,6 +442,17 @@ bool TabletImpl::CheckGetDone(::openmldb::api::GetType type, uint64_t ts, uint64
     return false;
 }
 
+void TabletImpl::UpdateMemoryUsage() {
+    base::SysInfo info;
+    if (auto status = base::GetSysMem(&info); status.OK()) {
+        system_memory_usage_rate_.store(info.mem_used * 100 / info.mem_total, std::memory_order_relaxed);
```
dl239: divided by zero

src/base/sys_info_test.cc
```diff
@@ -0,0 +1,48 @@
...
+
+TEST_F(SystemInfoTest, GetMemory) {
+    base::SysInfo info;
+    auto status = base::GetSysMem(&info);
+    ASSERT_TRUE(status.OK());
+    ASSERT_GT(info.mem_total, 0);
+    ASSERT_GT(info.mem_used, 0);
+    ASSERT_GT(info.mem_available, 0);
+    ASSERT_EQ(info.mem_total, info.mem_used + info.mem_available);
+    printf("total:%lu\n", info.mem_total);
```
dl239: delete

src/base/sys_info.h
```diff
@@ -0,0 +1,83 @@
...
+    uint64_t mem_total = 0;       // unit is kB
+    uint64_t mem_used = 0;        // unit is kB
+    uint64_t mem_available = 0;   // unit is kB
+};
+
+base::Status GetSysMem(SysInfo* info) {
+#if defined(__linux__)
+    FILE *fd = fopen("/proc/meminfo", "r");
+    if (fd == nullptr) {
+        return {ReturnCode::kError, "fail to open meminfo file"};
```
aceforeverd, line 51: print `errno` info

src/client/tablet_client.cc
```diff
@@ -202,19 +202,26 @@ bool TabletClient::UpdateTableMetaForAddField(uint32_t tid, const std::vector<op
 }
 
 bool TabletClient::Put(uint32_t tid, uint32_t pid, uint64_t time, const std::string& value,
-                       const std::vector<std::pair<std::string, uint32_t>>& dimensions) {
+                       const std::vector<std::pair<std::string, uint32_t>>& dimensions,
+                       int memory_usage_limit) {
```
dl239, line 206: add msg

src/base/sys_info.h
```diff
@@ -0,0 +1,83 @@
...
+#include "absl/strings/ascii.h"
+#include "absl/strings/match.h"
+#include "absl/strings/numbers.h"
+#include "absl/strings/string_view.h"
+#include "base/status.h"
+
+namespace openmldb::base {
+
+constexpr const char* MEM_TOTAL = "MemTotal";
+constexpr const char* MEM_AVAILABLE = "MemAvailable";
```
aceforeverd: acc https://access.redhat.com/solutions/406773, `MemAvailable` do not has this filed util RHEL 7, which means CentOS 6 do not have `MemAvailable`

src/base/sys_info_test.cc
```diff
@@ -0,0 +1,48 @@
...
+};
+
+TEST_F(SystemInfoTest, GetMemory) {
+    base::SysInfo info;
+    auto status = base::GetSysMem(&info);
+    ASSERT_TRUE(status.OK());
+    ASSERT_GT(info.mem_total, 0);
+    ASSERT_GT(info.mem_used, 0);
+    ASSERT_GT(info.mem_available, 0);
+    ASSERT_EQ(info.mem_total, info.mem_used + info.mem_available);
```
aceforeverd: pls compare output of `free` to see there is no big difference

src/base/sys_info.h
```diff
@@ -0,0 +1,83 @@
...
+
+struct SysInfo {
+    uint64_t mem_total = 0;       // unit is kB
+    uint64_t mem_used = 0;        // unit is kB
+    uint64_t mem_available = 0;   // unit is kB
+};
+
+base::Status GetSysMem(SysInfo* info) {
```
aceforeverd, line 47: why not use `sysinfo` directly ? checkout https://man7.org/linux/man-pages/man2/sysinfo.2.html

dl239: i cannot get the real used memory from sysinfo(). It looks like there is no cached memory

src/base/sys_info.h
```diff
@@ -0,0 +1,102 @@
...
+#include "absl/strings/string_view.h"
+#include "base/status.h"
+
+namespace openmldb::base {
+
+constexpr const char* MEM_TOTAL = "MemTotal";
+constexpr const char* MEM_BUFFERS = "Buffers";
+constexpr const char* MEM_CACHED = "Cached";
+constexpr const char* MEM_FREE = "MemFree";
+constexpr const char* SLAB = "Slab";
```
aceforeverd: man free
> cache  Memory used by the page cache and slabs (Cached and SReclaimable in /proc/meminfo)

dl239: what's the problem? pls let me know more detals

aceforeverd: @dl239  I think you should use `SReclaimable` instead of `Slab`

dl239: I see the manual of `free` is below under centos 7.5. centos 7.7 shows as you quote.
> cache  Memory used by the page cache and slabs (Cached and Slab in /proc/meminfo

in fact, the computation of `MemAvailable` is more complicated, refer [`si_mem_available`](https://elixir.bootlin.com/linux/v6.6/source/mm/show_mem.c).
we can use `SReclaimable`  as a rough estimate. refer [proc docs](https://www.kernel.org/doc/Documentation/filesystems/proc.txt)
> MemAvailable: An estimate of how much memory is available for starting new
              applications, without swapping. Calculated from MemFree,
              SReclaimable, the size of the file LRU lists, and the low
              watermarks in each zone.
              The estimate takes into account that the system needs some
              page cache to function well, and that not all reclaimable
              slab will be reclaimable, due to items being in use. The
              impact of those factors will vary from system to system.

lqy222 (approved):

aceforeverd (approved):

Issue #891, opened 2021-12-13
build(clean): allow delete proto/swig generated files in project


zhanghaohit:
two minor bugs in Makefile:
- `make clean` doesn't clean the swig genereated files
- `HYBRIDSE_CMAKE_DEPS_FLAGS := -DHYBRIDSE_TESTING_ENABLE=OFF -DEXAMPLES_ENABLE=OFF -DPYSDK_ENABLE=OFF -DJAVASDK_ENABLE=OFF` 
  these options are cached, simply overriding cannot change these values.

## Expected Behavior
- `make clean` at least provide an option to clean the generated files, including swig, proto and so on.
- By simply `make JAVASDK_ENABLE=ON`, we can make the hybridse javasdk

## Current Behavior
- `make clean` only delete the `build` and `hybridse/build`
- `make JAVASDK_ENABLE=ON` does not work as expected

aceforeverd:
1. Problem 1 Does it affected to proto generated files?
2. Problem 2, a solution might rm `HYBRIDSE_
CMAKE_DEPS_FLAGS`, but increase compile time.

 What about the convention: not compile tests when hitting `make`, while keep test targets still available in build directory?

zhanghaohit:
> 1. Problem 1 Does it affected to proto generated files?
> 2. Problem 2, a solution might rm `HYBRIDSE_ CMAKE_DEPS_FLAGS`, but increase compile time.
> 
> What about the convention: not compile tests when hitting `make`, while keep test targets still available in build directory?

1. Problem 1. If the .proto file or swig definition changes, does it re-compile atomically by executing `make` again? If not, we have to delete all the generated files by `make clean`.  Otherwise, users have to manually find and delete if they change proto or swig.
2. Problem 2. Maybe we can change the default value in the `hybridse/CMakeLists.txt`?

I think not compiling test by default is better? we can just keep the test target as what we currently do.

aceforeverd:
> > 1. Problem 1 Does it affected to proto generated files?
> > 2. Problem 2, a solution might rm `HYBRIDSE_ CMAKE_DEPS_FLAGS`, but increase compile time.
> > 
> > What about the convention: not compile tests when hitting `make`, while keep test targets still available in build directory?
> 
> 1. Problem 1. If the .proto file or swig definition changes, does it re-compile atomically by executing `make` again? If not, we have to delete all the generated files by `make clean`.  Otherwise, users have to manually find and delete if they change proto or swig.
[...]

PR #3636, opened 2023-11-28
build(proto): improve protobuf generation

### LINKED ISSUE #1090: build(proto java): generate to one module is sufficient
https://github.com/4paradigm/OpenMLDB/blob/1e643c58b58f93a62eb8f5b79e3b92e68b0f62f1/src/CMakeLists.txt#L21

proto generated java output to four java modules, redundant codes.  

Fix: 
make sure one *.proto output to one java module, other java modules depends by module

1. output to openmldb-common only.  ​
2. or different proto file output to different java modules based on usage.

aceforeverd:
1. tracking all generated files (.cpp, .h, .java), instead of .cpp file only
2. auto clean all generated files on clean target, resolving #891 for proto files
3. simplify dependency in java modules, fixes #1090
   - proto generated files to openmldb-common only, instead of multiple projects
   - openmldb-native now depends on openmldb-common, no direct dependency to openmldb-common from other modules

### DIFF
--- a/hybridse/src/proto/CMakeLists.txt
+++ b/hybridse/src/proto/CMakeLists.txt
@@ -12,80 +12,43 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-include_directories(${INCLUDE_DIRECTORIES}
-        ${PROJECT_SOURCE_DIR}/src
-        ${PROJECT_BINARY_DIR}/src)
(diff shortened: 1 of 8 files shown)


tobegit3hub (approved):
LGTM

dl239 (approved):

PR #3641, opened 2023-11-29
build: support build on no tag

vagetablechicken:
if REF set empty, branch or commit, failed on grep HEAD. Ignore this error, just use snapshot to be the version. The doc is correct, no update about it.

zetasql lib built by workflow run is wierd, but can't solve it now. So we add zetasql lib url to replace it in workflow.


aceforeverd (approved):

tobegit3hub (approved):
LGTM

PR #3649, opened 2023-11-30
fix: throw null pointer exception if the set value is null in `RowBuilder` 

### LINKED ISSUE #3637: no null protect in FlexibleRowBuilder  setString
https://github.com/4paradigm/OpenMLDB/blob/b005cd11ba30c9645804ed1a0fc4f715326e3e82/java/openmldb-common/src/main/java/com/_4paradigm/openmldb/common/codec/FlexibleRowBuilder.java#L340

if val is null, we'll get an error:
```
java.lang.NullpointerException: null
at com._4paradigm.openmldb.common.codec.FlexibleRowBuilder.setstring(FlexibleRowBuilder.java:337)
at com._4paradigm.openmldb,sdk.impl.BatchcallablepreparedstatementImpl,setstring(BatchcallablereparedstatementImpl.java:215)
```

dl239:
close #3637

### DIFF
--- a/java/openmldb-common/src/main/java/com/_4paradigm/openmldb/common/codec/ClassicRowBuilder.java
+++ b/java/openmldb-common/src/main/java/com/_4paradigm/openmldb/common/codec/ClassicRowBuilder.java
@@ -266,14 +266,14 @@ public class ClassicRowBuilder implements RowBuilder {
 
     @Override
     public boolean appendString(String val) {
+        if (!check(DataType.kVarchar) && !check(DataType.kString)) {
+            return false;
+        }
         if (val == null) {
             return appendNULL();
         }
         byte[] bytes = val.getBytes(CodecUtil.CHARSET);
         int length = bytes.length;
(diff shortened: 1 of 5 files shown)


java/openmldb-common/src/main/java/com/_4paradigm/openmldb/common/codec/ClassicRowBuilder.java
```diff
@@ -205,6 +205,9 @@ public boolean appendTimestamp(Timestamp val) {
         if (!check(DataType.kTimestamp)) {
             return false;
         }
+        if (val == null) {
+            return appendNULL();
```
vagetablechicken, line 209: Not all jdbc driver can do this `setXX(1,null)->setNull(1)`, it's a good feature, but we should add some desc about it in java sdk doc?

vagetablechicken (approved):

lqy222 (approved):

PR #3696, opened 2024-01-09
feat(udf): map data type

aceforeverd:
- deps: zetasql -> 0.3.3
- parser: map data type
- parser: create table with array/map data type
- parser & node: extract array/map value with []operator
- udf: map functions
- udf: map_keys function
- codegen: construct map data type
- codegen: access map value by key with []operator
- type system: NULL and VOID type
- refactor & IWYU

TODO
- []operator support for array type
- codegen: use vector type & SIMD instructions instead of array ?


tobegit3hub (approved):
LGTM

dl239 (approved):

PR #3716, opened 2024-01-23
feat: support absandlat/absorlat gc in disktable

### LINKED ISSUE #2748: disk table `DeleteIndex` cannot delete the data if there are multiple indexes with same column family
**Bug Description**
For `DiskTable`, multiple indexes with same `key` column will share the same column family. It is not easy to only delete the data associated with one single index that may share the same column family with others.

Potential solutions:
- re-structure the disk table index: one column family corresponds to one index

### LINKED ISSUE #2992: Support `kAbsAndLat` TTL type for disk table

dl239:
current pr is based #3711, close #2992

- fix #2748 
- support absandlat/absorlat gc in disktable


tobegit3hub (approved):
LGTM

PR #3746, opened 2024-02-04
feat: offline map type

aceforeverd:
feat
-----
- query with map data type from offline
- native row builder for offline 

limitation 
------
- not able to native encode empty map value, native row builder need some work to take the correct type for empty map, since a empty map is translated to `map()` expression that will inferred as `map<void, void>`, it does not consider type in input schema
- each row will encoded with distinct native instance due to #3748, a jit can not add multiple modules
- unsaferow encoding for map values


tobegit3hub (approved):
LGTM

vagetablechicken (approved):

PR #3765, opened 2024-02-21
feat: support map data type in yaml testing framework

aceforeverd:
allow map data type in YAML testing framework for input definition, you
may define input table's schema as:

  columns: ["col1 int", "col2 map<int, string>"]

which is exactly the same as the SQL syntax in CREATE TABLE statement.

And define table values with `inputs[*].inserts`:

  inserts:
    - insert into t1 values (1, map(12, "abc"))

just write down the insert statement SQL.

*LIMITATIONS*

- unsupported: map data type in `expect` field
- INSERT statement with custom columns definition, implementation
  limits, each insert value expression must matches exactly to the table column definition

### DIFF
--- a/src/sdk/sql_cluster_router.cc
+++ b/src/sdk/sql_cluster_router.cc
@@ -43,6 +43,7 @@
 #include "brpc/channel.h"
 #include "cmd/display.h"
 #include "codec/encrypt.h"
+#include "codegen/insert_row_builder.h"
 #include "common/timer.h"
 #include "glog/logging.h"
 #include "nameserver/system_table.h"
@@ -62,7 +63,6 @@
 #include "sdk/split.h"
 #include "udf/udf.h"
 #include "vm/catalog.h"
-#include "codegen/insert_row_builder.h"
 
 DECLARE_string(bucket_size);
 DECLARE_uint32(replica_num);
--- a/cases/query/udf_query.yaml
+++ b/cases/query/udf_query.yaml
@@ -619,3 +619,22 @@ cases:
     sql: |
       select c1 + 8 from (select 9 as c1)
 
+  - id: 18
+    mode: request-unsupport
+    inputs:
+      - name: t1
+        columns: ["col1 int32", "std_ts timestamp", "col2 map<int,string>"]
+        indexs: ["index1:col1:std_ts"]
+        # insert map values by insert stmts only
+        inserts:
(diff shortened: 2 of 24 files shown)


src/sdk/sql_cluster_router.cc
```diff
@@ -62,6 +62,7 @@
 #include "sdk/split.h"
 #include "udf/udf.h"
 #include "vm/catalog.h"
+#include "codegen/insert_row_builder.h"
```
dl239: reorder the new header file

dl239:
Is there  performance comparison between codegen encoding and the original encoding?

src/cmd/sql_cmd_test.cc
```diff
@@ -3286,6 +3286,53 @@ TEST_P(DBSDKTest, CreateIfNotExists) {
...
+    absl::BitGen gen;
+    auto db = absl::StrCat("db_", absl::Uniform(gen, 0, std::numeric_limits<int32_t>::max()));
+    auto table = absl::StrCat("tb_", absl::Uniform(gen, 0, std::numeric_limits<int32_t>::max()));
+
+    ProcessSQLs(sr, {
+                        "set session execute_mode = 'online'",
+                        absl::StrCat("create database ", db),
+                        absl::StrCat("use ", db),
+                        absl::Substitute("create table $0 (id string, val map<int, string>)", table),
+                        absl::Substitute("insert into $0 values ('1', map(12, '23')) ", table),
```
dl239: how to insert map fields with placeholder

aceforeverd: not possible util codegen insert row support parameters

aceforeverd:
> Is there performance comparison between codegen encoding and the original encoding?

we should do some benchmark after codegen insert row generally stable (current alpha, interface may change).

src/cmd/sql_cmd_test.cc
dl239: just user interface. and one sql that contains map  placeholder only execute codegen at the first run time?

aceforeverd: it will codegen a function with extra function parameter, this parameter type might be `codec::Row` or `node::ExprListNode`, I'd prefer design it later

tobegit3hub (approved):
LGTM

vagetablechicken (approved):

Issue #3760, opened 2024-02-20
docker image for developers to build openmldb from scratch without internet for native codes


tobegit3hub:
**Describe the feature you'd like**

Now we have provide a docker image to build openmldb from scratch in http://openmldb.ai/docs/zh/main/deploy/compile.html . It will download some files from github and require users to download the openmldb source.

We may provide a better docker image so that developers can build source code without internet.

**Additional context**

Base on the docker image of "4pdosc/hybridsql:0.8" and download the required files when `make configure`. Add the source code of openmldb in docker image without `.git` files.

aceforeverd:
A few things may require internet resource:
- pre-compiled thirdparty
- pre-compiled zetasql
- cmake configure time deps: rapidjson & farmhash

break down for the full-offline dockerfile: 
1. hybridsql image already bunded zetasql & thirdparty, but leaving zetasql version outdated, update version number in https://github.com/4paradigm/OpenMLDB/blob/d760b5078950d13bcd0b98b88d132567cb47e446/docker/Dockerfile#L18. latest version is defined in `third-party/cmake/FetchZetasql.cmake`
2. bundle rapidjson & farmhash directly in source code, see how `contrib/` directory works
3. since bump new version for thirdparty or zetasql in `third-party` directory will forget to update `docker/Dockerfile` easily, it may necessary to add a simple test to verify the version is exact same from the two places.
4. include OpenMLDB source code in another Dockerfile

tobegit3hub:
To deploy and start openmldb in this docker image, we still need to install `rsync`.

PR #3773, opened 2024-02-22
feat: offline docker

QiChenX:
feature

https://github.com/4paradigm/OpenMLDB/issues/3760
bundle rapidjson & farmhash directly in source code

bundle rapidjson & farmhash directly in source code, allowing users to compile OpenMLDB offline


docker/Dockerfile
aceforeverd, line 1: this Dockerfile does not need update

aceforeverd (changes requested):

aceforeverd:
@QiChenX pls make framhash and rapidjson as submoudle in contrib directory. And duing cmake configure, you should check if the target submodule directory is empty, is not, run `git submodule <target_directory>` to initicate that submodule directory, this avoid the case when git clone does not with `--submodules`

PR #3774, opened 2024-02-22
fix: coverage

dl239:
When `glog` compiles a static library, it will hide some symbols in v0.6.0 (refer [here](https://github.com/google/glog/blob/v0.6.0/CMakeLists.txt#L64)) so that coverage cannot find the symbols in linking stage when it relies on the dynamic libraries
```
2024-02-23T06:19:25.8332883Z /usr/bin/ld: cmd/single_tablet_test: hidden symbol `_ZN6google10LogMessageC1EPKcii' in /home/runner/work/OpenMLDB/OpenMLDB/.deps/usr/lib/libglog.a(logging.cc.o) is referenced by DSO
2024-02-23T06:19:25.8335221Z /usr/bin/ld: final link failed: bad value
2024-02-23T06:19:25.8336221Z collect2: error: ld returned 1 exit status
```
A quick temporary solution is to compile the glog dynamic library before running coverage


.github/workflows/coverage.yml
```diff
@@ -33,6 +33,8 @@ on:
 jobs:
   coverage:
     runs-on: ubuntu-20.04
+    container:
```
aceforeverd: don't do that, ubuntu is required

aceforeverd:
That should the common issue when linking a shared library against static libraray, see https://stackoverflow.com/questions/23696585/what-does-exactly-the-warning-mean-about-hidden-symbol-being-referenced-by-dso

So I've try bundle source code of glog (plug gflags and anything necessary), then we can link at least gflag & glog from source

aceforeverd (approved):

aceforeverd:
it build glog_shared, and FindPackage(glog) will find shared library first ?

dl239:
> Details

It seems so

dl239:
> it build glog_shared, and FindPackage(glog) will find shared library first ?

It seems so

tobegit3hub (approved):
LGTM

PR #3777, opened 2024-02-25
build(deps): thirdparty builds

aceforeverd:
- CMP0135 NEW
- glog: enable libunwind, don't build in source
- llvm: rm z3 dependency, macOS build may implicitly enabled
- libunwind: rm feature minidebuginfo, this requires lzma dependency
- gflags 2.2.2

PR #3780, opened 2024-02-26
fix: delete with same ts

dl239:
- fix integration-test
- fix hybridse-ci compiling error on macos


.github/workflows/hybridse-ci.yml
```diff
@@ -61,7 +61,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          brew install coreutils
```
dl239, line 64: revert here after removing llvm z3 dependency in #3777 @aceforeverd

aceforeverd (approved):

vagetablechicken (approved):

PR #3782, opened 2024-02-27
feat: set `NPROC` in intergration test

dl239:
`SELFHOST-INTEGRATION-TEST` takes too much time in build stage now as `OpenMLDB` use one thread to compile by default. we can set `NPROC` to compile in parallel


tobegit3hub (approved):
LGTM

aceforeverd (approved):

Issue #3779, opened 2024-02-26
After inserting data with precompiled SQL, when queried, certain fields have additional null characters ('\x00') appended at the end


yht520100:
**Bug Description**
Service Version: Code from the main branch after version 0.8.4, compiled and deployed on macOS 13.5.1.
Client Version: The current project branch python-sdk, accessing the database using the dbapi method.

After inserting data with precompiled SQL, when queried, certain fields have additional null characters ('\x00') appended at the end, as shown in the fourth column:
`
[(1001, '2022-05-01', 'province1', 'city1\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', 1, 1590738991000), 
(1002, '2022-05-02', 'province2', 'city2\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', 2, 1590738992000), 
(1003, '2022-05-03', 'province3', 'city3\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', 3, 1590738993000), 
(1004, '2022-05-04', 'province0', 'city4\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', 4, 1590738994000)]
`

**Expected Behavior**
The field contents should not have additional null characters appended and should match the inserted content.

**Relation Case**

**Steps to Reproduce**

1. Use the python-sdk code from the main branch after version 0.8.4.
2. Run the test cases below to reproduce the corresponding error.

```
import sys
import os
from openmldb.dbapi import connect
import pytest

from .case_conf import OpenMLDB_ZK_CLUSTER, OpenMLDB_ZK_PATH

class TestOpenmldbDBAPI:
    cursor_without_db = None
    cursor = None
    db_name = "dbapi_test"

    @classmethod
    def setup_class(cls):
        # connect without db to create it
        db = connect(zk=OpenMLDB_ZK_CLUSTER, zkPath=OpenMLDB_ZK_PATH)
        cls.cursor_without_db = db.cursor()
        cls.cursor_without_db.execute(
            f"create database if not exists {cls.db_name};")
        # then we can use the db to connect
        db = connect(database=cls.db_name,
                     zk=OpenMLDB_ZK_CLUSTER,
                     zkPath=OpenMLDB_ZK_PATH)
        cls.cursor = db.cursor()
        cls.cursor.execute("create database if not exists {};".format(cls.db_name))
        assert cls.db_name in cls.cursor.get_databases()

    def recreate_table(self, table, schema):
[...]

PR #3788, opened 2024-02-28
fix: '\x00' Issue in Python SDK Precompiled SQL Inserts

yht520100:
Bug fix

* https://github.com/4paradigm/OpenMLDB/issues/3779
Resolve the issue of extra '\x00' characters in string fields when inserting data using precompiled SQL in the Python SDK.


python/openmldb_sdk/openmldb/dbapi/dbapi.py
```diff
@@ -291,44 +291,50 @@ def execute(self, operation, parameters=()):
...
+                continue
             if isinstance(row, tuple):
-                if isinstance(row[i], str):
-                    str_size += len(row[i])
+                col_value = row[i]
             elif isinstance(row, dict):
                 if name not in row:
                     raise DatabaseError("col {} data not given".format(name))
                 if row[name] is None:
                     if schema.IsColumnNotNull(idx):
```
vagetablechicken: tuple need check not null too?

yht520100: ok，updated

vagetablechicken: > ok，updated

I don't see any change about check not null when row is an instance of tuple, you only check when row is dict

yht520100: updated, please review

yht520100: > > ok，updated
> 
> I don't see any change about check not null when row is an instance of tuple, you only check when row is dict

updated, please review

python/openmldb_sdk/tests/dbapi_parameterized_test.py
```diff
@@ -0,0 +1,133 @@
...
+import sys
+import os
+from openmldb.dbapi import connect
+import pytest
+from datetime import datetime
+
+from .case_conf import OpenMLDB_ZK_CLUSTER, OpenMLDB_ZK_PATH
+
+
+class TestOpenmldbDBAPI:
```
vagetablechicken: Test class name should be the same with file name

yht520100: updated

vagetablechicken (approved):

aceforeverd (approved):

tobegit3hub (approved):
LGTM

PR #3799, opened 2024-03-11
test: add automatic table cleanup after go sdk tests

oh2024:
* **What kind of change does this PR introduce?**
This PR introduces a test enhancement in the Go SDK. It adds automatic cleanup logic for database table created during tests, ensuring that tests can be rerun without the need for manual table deletion.

Currently, the tests in the Go SDK create demo table in the database to verify various functionalities. However, it's not automatically cleaned up after the tests are completed. This behavior requires manual intervention to delete the table before tests can be rerun, leading to inefficiencies in the development process.

With this change, the Go SDK tests now include cleanup logic that automatically deletes demo table created during the testing process. This ensures that the database is restored to its initial state after the tests complete, allowing for tests to be rerun multiple times without manual cleanup. This improvement streamlines the testing process, making it more efficient and less error-prone.

### DIFF
--- a/go/go_sdk_test.go
+++ b/go/go_sdk_test.go
@@ -37,6 +37,15 @@ func Test_driver(t *testing.T) {
 		_, err := db.ExecContext(ctx, createTableStmt)
 		assert.NoError(t, err, "fail to exec %s", createTableStmt)
 	}
+
+	defer func() {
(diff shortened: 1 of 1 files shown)


tobegit3hub (approved):
LGTM

aceforeverd (approved):

PR #3804, opened 2024-03-14
feat: Add Object Table for Authz in __INTERNAL_DB

oh2024:
**What kind of change does this PR introduce?** 
This PR introduces a new feature that adds an object table to OpenMLDB for authorization purposes. The table is accessible in the __INTERNAL_DB database and is named OBJECT. This feature allows for enhanced security measures by facilitating permission checks and other authorization mechanisms through the newly introduced table.

**What is the current behavior?** 
Currently, OpenMLDB lacks a dedicated object table for authorization purposes, limiting the ability to implement fine-grained access control and permission checks within the database.

**What is the new behavior (if this is a feature change)?**
With this change, OpenMLDB introduces an OBJECT table under the __INTERNAL_DB database, which can be interacted with using the DESCRIBE OBJECT; command. This table is designed to store authorization-related information, enabling developers and database administrators to implement more sophisticated authz mechanisms. This feature enhances OpenMLDB's security model by allowing for detailed permission settings and access control at a granular level.

### DIFF
--- a/src/nameserver/system_table.h
+++ b/src/nameserver/system_table.h
@@ -19,6 +19,7 @@
 
 #include <memory>
 #include <string>
+
 #include "absl/strings/string_view.h"
 #include "base/status.h"
 #include "gflags/gflags.h"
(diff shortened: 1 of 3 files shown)


src/nameserver/system_table.h
```diff
@@ -49,11 +50,11 @@ enum class SystemTableType {
     kGlobalVariable = 3,
     kDeployResponseTime,
     kUser,
+    kObject,
```
aceforeverd, line 53: `object` is too general, pick a better name for privilege

oh2024: What about `UserPrivilege`?

PR #3813, opened 2024-03-19
test: add a test case on memory usage  comparison with Redis.

gaoboal:
- add a test case on memory usage  comparison with Redis;
- update benchmark test document;


aceforeverd (approved):

tobegit3hub (approved):
LGTM

PR #3819, opened 2024-03-20
fix: correct errors in the examples for 'LOAD DATA INFILE' statement.

gaoboal:
Add parameter mode='append' for the examples of 'LOAD DATA INFILE' statement in online mode.

The SQL statement cannot be executed correctly in the examples because the default value of the 'mode' parameter is 'error_if_exists', but in online mode, only mode='append' is supported.


vagetablechicken (approved):

tobegit3hub (approved):
LGTM

Issue #3733, opened 2024-01-29
Select constant data will not return value row


tobegit3hub:
**Bug Description**

Now we execute "select 1" will generate dataframe with one colume but not data.

tobegit3hub:
Resolved.

PR #3825, opened 2024-03-22
fix: select constant return empty value

Matagits:
Fix the problem that when select constant data and then it will generate dataframe with one colume but not data, and add corresponding test code. And when sql expression contains cast expression to select constant in offline mode, it will throw exception.

Issue#3733. https://github.com/4paradigm/OpenMLDB/issues/3733
Plus, when sql expression contains cast expression to select constant in offline mode, it will throw exception.

When select constant data, it will generate dataframe with one row of corresponding constant data. And also support cast expression when select constant in offline mode.


tobegit3hub (approved):
LGTM

tobegit3hub:
Please resolve the cicd issue in https://github.com/4paradigm/OpenMLDB/actions/runs/8419907957/job/23053523375?pr=3825 .

aceforeverd (approved):

PR #3835, opened 2024-03-26
feat: add nameserver auth

oh2024:
What kind of change does this PR introduce?
This PR introduces a significant security enhancement by implementing RPC authenticator functionality to ensure all connections are authenticated server-side. This is a feature change aimed at improving the security of server-client communications.

What is the current behavior?
Currently, there is no authentication performed on the server side for incoming connections. Authentication is solely handled within the SDK, leaving a potential security gap for server-client communications.

What is the new behavior (if this is a feature change)?
With this change, all connections to the server will undergo authentication at the beginning of the rpc connection. This server-side authentication is facilitated by the newly introduced Authenticator class, which supports two types of authentication: service authentication and user authentication. For the sake of maintaining backward compatibility, the system will initially allow all services carrying the "default" token and users with the username "root". This PR lays down the foundational framework for our enhanced security measures, focusing on authentication at the server side. Future PRs will expand on this by incorporating user verification against tablet data and authenticating service tokens via a Zookeeper service token record. This implementation is a critical step toward tightening security and providing a robust authentication mechanism for our system.


src/auth/authenticator.h
```diff
@@ -0,0 +1,46 @@
...
+using AuthToken = std::variant<ServiceToken, UserToken>;
+
+inline AuthToken g_auth_token;
+
+class Authenticator : public brpc::Authenticator {
+ public:
+    using GetUserPasswordFunc = std::optional<std::string> (*)(const std::string&);
+    Authenticator() : getUserPassword(DefaultGetUserPassword) {}
+    Authenticator(GetUserPasswordFunc getUserPasswordFunc) : getUserPassword(getUserPasswordFunc) {}
+    int GenerateCredential(std::string* auth_str) const;
```
aceforeverd: add `override` specifier

src/auth/authenticator.cc
```diff
@@ -0,0 +1,62 @@
+#include "authenticator.h"
+
+#include "base/glog_wrapper.h"
```
vagetablechicken: this file needs log header?

tobegit3hub (approved):
LGTM

vagetablechicken:
@oh2024 too much failure in checks, do you pass all in your localhost?

oh2024:
integration test will fail because old sdk is used

src/tablet/drop_procedure_before_drop_table_test.cc
```diff
@@ -125,160 +125,35 @@ void ShowTable(::openmldb::RpcClient<::openmldb::nameserver::NameServer_Stub>& n
...
-    request.set_show_all(true);
+    request.set_show_all(false);
     bool ok = name_server_client.SendRequest(&::openmldb::nameserver::NameServer_Stub::ShowTable, &request, &response,
                                              FLAGS_request_timeout_ms, 1);
     ASSERT_TRUE(ok);
     ASSERT_EQ(response.table_info_size(), size);
 }
 
-TEST_F(SqlClusterTest, RecoverProcedure) {
+TEST_F(SqlClusterTest, DropProcedureBeforeDropTable) {
```
vagetablechicken: Why rename this test file? `drop_procedure_before_drop_table_test.cc` is not a good name. You'd better make the file name be the same with the test class, not the test method name.

vagetablechicken (changes requested):
no docs about privileges, add in auth.md after #3879

src/tablet/drop_procedure_before_drop_table_test.cc
oh2024: I need to separate the RecoverProcedure and DropProcedureBeforeDropTable for the tests to pass. There's probably some bugs in the tests that we can fix later

src/auth/user_access_manager.cc
```diff
@@ -0,0 +1,82 @@
...
+                                     std::shared_ptr<nameserver::TableInfo> user_table_info)
+    : user_table_iterator_factory_(std::move(iterator_factory)), user_table_info_(user_table_info) {
+    StartSyncTask();
+}
+
+UserAccessManager::~UserAccessManager() { StopSyncTask(); }
+
+void UserAccessManager::StartSyncTask() {
+    sync_task_running_ = true;
+    sync_task_thread_ = std::thread([this] {
```
aceforeverd, line 36: query table every 0.1 seconds ? Not good practice  if my understand is correct, instead, you should lazy fetch the table with help of zk. 

Register an event listener to zk > on table values update, notify zk > zk will send data updated event to every subscriber .

oh2024: Yeah this is just temporary solution to get latest user credentials. In the future we will use a distributed transaction to update user table cache on all nodes.

vagetablechicken:
merge after 0.9.0 released, cherry-pick is heavy

vagetablechicken (approved):

PR #3841, opened 2024-03-31
feat(codegen): map data type patches

aceforeverd:
1. feat: add simple debug facility `printLog` in codegen so JITed code is able to print simple logs runtime
2. fix: **extract map value** when value is NULL, current codegen still inspect the NULL value (especially value is a struct ptr like string), so a safe null struct should always constructed, instead of a `UndefValue`
3. feat(draft): make LLJIT accept `JItOptions` as well, it does not have real difference now, but should expected to support later.
4. fix: gen create table sql in sql_sdk_test


tobegit3hub (approved):
LGTM

vagetablechicken (approved):

PR #3847, opened 2024-04-03
feat(map): cast null or map() to explicit map type

aceforeverd:
Enhancements to new insert row builder:

**Auto casting to target type when necessary, which usually happens with insert statement while a default schema defined by target table.**

With the patch (whenever `InsertRowBuilder` used), it can
1. automatically construct safe value for `NULL` place holder, so insert statement like `insert to t1 values (1, null, null, map())` simply work for table t1 `(int, timestamp, map<int, string>, map<timestamp, int>)`
2. supports constructing null or empty values from offline engine.

### DIFF
--- a/cases/plan/create.yaml
+++ b/cases/plan/create.yaml
@@ -798,7 +798,7 @@ cases:
           |  +-0:
           |  |  +-node[kColumnDesc]
           |  |    +-column_name: column1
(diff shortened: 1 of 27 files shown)


tobegit3hub (approved):
LGTM

vagetablechicken (approved):

PR #3858, opened 2024-04-09
build: report status if patch failed

### LINKED ISSUE #3857: glog patch failed when BUILD_SHARED_LIBS
If the patch failed in glog cmakelists when BUILD_SHARED_LIBS=ON, it report:
```
CMake Error: install(EXPORT "glog-targets" ...) includes target "glog" which requires target "gflags_nothreads_shared" that is not in any export set.
CMake Error in contrib/glog/CMakeLists.txt:
  export called with target "glog" which requires target
  "gflags_nothreads_shared" that is not in this export set, but in multiple
  other export sets:
  /xxx/OpenMLDB2/build/contrib/gflags/gflags-nonamespace-targets.cmake,
  /xxx/OpenMLDB2/build/contrib/gflags/gflags-targets.cmake.

  An exported target cannot depend upon another target which is exported
  multiple times.  Consider consolidating the exports of the
  "gflags_nothreads_shared" target to a single export.
```

It's not the dup targets in contrib/gflags, the root cause is glog origin cmake will find the gflags in .deps, then be dup with contrib/glfags.

- [x] report message if patch failed
- [x] why patch failed? no patch binary 🤣

vagetablechicken:
closes #3857

### DIFF
--- a/contrib/glog-cmake/CMakeLists.txt
+++ b/contrib/glog-cmake/CMakeLists.txt
@@ -1,10 +1,15 @@
 set(WITH_GTEST OFF)
 set(WITH_GFLAGS ON)
 set(WITH_UNWIND ON)
-
 execute_process(
   COMMAND patch -p 1 -N -i ${CMAKE_SOURCE_DIR}/contrib/glog-cmake/find-gflags.patch
   WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/contrib/glog
+  RESULT_VARIABLE ret
+  ECHO_OUTPUT_VARIABLE
+  ECHO_ERROR_VARIABLE
+  COMMAND_ECHO STDOUT
 )
-
+if(NOT ret EQUAL "0")
+    message(STATUS  "Bad exit status '${ret}', be careful")
+endif()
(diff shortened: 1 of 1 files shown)


contrib/glog-cmake/CMakeLists.txt
```diff
@@ -1,10 +1,13 @@
...
-
 execute_process(
   COMMAND patch -p 1 -N -i ${CMAKE_SOURCE_DIR}/contrib/glog-cmake/find-gflags.patch
   WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/contrib/glog
+  RESULT_VARIABLE ret
+  COMMAND_ECHO STDOUT
 )
-
+if(NOT ret EQUAL "0")
+    message(STATUS  "Bad exit status '${ret}', be careful")
```
aceforeverd: `WARNING`  better

vagetablechicken: OK

aceforeverd (approved):

tobegit3hub (approved):
LGTM

PR #3859, opened 2024-04-09
fix: ttl merge

### LINKED ISSUE #3837: mismatch ttl type in isexpire and ttlmerge
https://github.com/4paradigm/OpenMLDB/blob/4c6406e48d1ea9e73016ec01396575684675b799/src/storage/schema.h#L152

abs and lat should make all <abs_time and >lat_num expired, so it should be < || >, not &&, abs or lat is wrong too.
UT is wrong? `TEST_F(SchemaTest, TestIsExpired)`

TTLSt wrong too?
https://github.com/4paradigm/OpenMLDB/blob/b1b8b63424a13a78ffd70b93de4c6e2bf621f7f4/src/storage/schema.h#L111

reproduce way:
```
create database db;
use db;
-- drop table t1;
-- latest 可以只保留1，但absandlat不行，无论索引key单个还是多个
-- create table t1(c1 string, c2 string, c3 int, index(key=(c1), ttl_type=latest, ttl=1));
create table t1(c1 string, c2 string, c3 int, index(key=(c1), ttl_type=absandlat, ttl=(10d,1)));
set @@execute_mode='online';
insert into t1 values ("a","b", 1),("a","b", 2);
select * from t1;
```

## ttl type in isexpire used

&& means `is_expire = is_abs_expire && is_lat_expire`, || means `is_expire = is_abs_expire || is_lat_expire`. Don't misunderstrand it, e.g.  && means `valid = valid_inabs && valid_in_lat` -> `expire = is_abs_expire || is_lat_expire`, || expires more rows.

## mismatch in ttl merge
TTL merge is used in creating index, and we use `||` to merge abs ttl and lat ttl, it's wrong, to store both valid in abs and valid in lat, we should use `&&`.

vagetablechicken:
closes #3837 

merge logic ref https://gist.github.com/vagetablechicken/bbd5d084987e975f31d5723dcd908e1c


tobegit3hub (approved):
LGTM

aceforeverd (approved):

PR #3873, opened 2024-04-12
fix: ddl parser get dup col keys

### LINKED ISSUE #3830: ddl parse error: dup col
```
create database db;
use db;
create table t1(c1 int);
create table t2(c1 int);
create table t3(c1 int);
deploy d2 select * from t1 last join t2 on t1.c1=t2.c1 and t1.c1=t2.c1;
-- error: Error: [2000] table [db.t2] add index failed. duplicated col c1
deploy d2 select * from t1 last join t2 on t1.c1=t2.c1 last join t3 on t1.c1=t3.c1 and t2.c1=t3.c1;
-- error too
```

vagetablechicken:
closes #3830


aceforeverd (approved):

tobegit3hub (approved):
LGTM

PR #3888, opened 2024-04-22
fix: recoverdata support load disk table

### LINKED ISSUE #3882: recoverdata disk table
Recoverdata can't support to recover disk table now, and it will generate wrong metadata for disk table. Fix it ASAP.

We can test it on openmldb-compose, create a disk table and restart tablets, do recoverdata(if we use start-all.sh, it'll do recoverdata).

vagetablechicken:
close #3882 
Deprecate old loadtable(using tablet client), use http to send request, it's simple and clean.


Elliezza (approved):

tobegit3hub (approved):
LGTM

PR #3892, opened 2024-04-25
docs: change version to 0.9.0 in docs and pom

tobegit3hub:


Elliezza (approved):

aceforeverd:
what about the version number in C/java/python ?

tobegit3hub:
> what about the version number in C/java/python ?

Cicd will use tag name to deploy. We will update these version in post script after deploying.

aceforeverd (approved):

PR #3894, opened 2024-04-25
feat: use local logger instead of slf4j annotation which causes compile fail

tobegit3hub:
Fail when compiling java. Refer to https://github.com/4paradigm/OpenMLDB/actions/runs/8817992530/job/24205745582

7d469d1f4d1a689651cfa3d25805ef47


tobegit3hub:
The build progress is successful in https://github.com/4paradigm/OpenMLDB/actions/runs/8827345942 .

image

vagetablechicken (approved):

aceforeverd (approved):

PR #3896, opened 2024-04-25
fix: discrete signature use sign int

wyl4pd:
discrete signature use sign int64_t format instead of uint64_t


tobegit3hub (approved):
LGTM

aceforeverd (approved):

PR #3906, opened 2024-04-28
fix: make clients use always send auth info

oh2024:
Bug fix

Clients need a flag to enable BRPC authentication and send tokens. This causes a problem for the java sdk which cannot set GFLAGS to enable auth

Clients always send brpc auth tokens. Servers validate tokens or not based on the skip grant tables flag


tobegit3hub (approved):
LGTM

vagetablechicken (approved):

PR #3918, opened 2024-05-07
feat: tablets get user table remotely

oh2024:
Feature

All tablets get a copy of the user table. There will be problems when new tablets are added after deployment because replica size is fixed.

At most 2 tablets contain user table. Other tablets get them remotely


vagetablechicken (approved):
resolve the conflicts

tobegit3hub (approved):
LGTM

Issue #820, opened 2021-12-01
built-in function: Support `locate(substr, str[, pos])` function


jingchen2222:
**Is your feature request related to a problem? Please describe.**

### locate(substr, str[, pos])
Returns the position of the first occurrence of substr in str after position pos. The given pos and return value are 1-based.

### Examples:
```sql
> SELECT locate('bar', 'foobarbar');
 4
> SELECT locate('bar', 'foobarbar', 5);
 7
```

### Reference:
https://spark.apache.org/docs/2.3.0/api/sql/#locate

**Describe the solution you'd like**

We do provide a built-in function development guide for you. 
Please check [OpenMLDB Build-In Function Develop Guide](https://github.com/4paradigm/OpenMLDB/blob/main/docs/en/built_in_function_develop_guide.md) or the [Chinese version](https://github.com/4paradigm/OpenMLDB/blob/main/docs/cn/built_in_function_develop_guide.md) for help.

**Please make sure you:**
- [ ] Add built-in C++ function in src/udf/udf.h and src/udf/udf.cc
- [ ] Register function to default library from function `void DefaultUdfLibrary::IniStringUdf()` in src/udf/default_udf_library.cc.
- [ ] Add related unit test in src/codegen/udf_ir_builder_test.cc
- [ ] Documenting function 

**Describe alternatives you've considered**

**Additional context**

Mouray-Hutchinson:
Hello, I would like some clarification on the parameter and return types of this function. I've written the function as 1
but I am aware that it is likely something like 2 or 3. Please let me know so I can commit my changes. 

1. unsigned int locate(std::string substr, std::string str, unsigned int pos =0 ) {..}
2. int32_t locate( hybridse::codec::StringRef substr, hybridse::codec::StringRef str,int32_t pos =0 ) {...}
3. int32_t locate(string substr,string str,int32_t pos =0 ) {...}

jingchen2222:
> 

Hi Mouray-Hutchinson, very glad that you would like to help.
Firstly,  I would like to remove the `SELECT POSITION('bar' IN 'foobarbar')` since it might be conflict with our `IN` expression. Maybe we can simply support `locate('bar', 'foobarbar')` and `locate('bar', 'foobarbar', 5)` for this issue.
Secondly, we had better implement a function looks like `int32_t locate( hybridse::codec::StringRef substr, hybridse::codec::StringRef str,int32_t pos =0 )` in udf.cc

[...]

PR #3940, opened 2024-05-28
feat: add udf function locate

howdb:
* **feature**

[issue](https://github.com/4paradigm/OpenMLDB/issues/820)

* **add a new udf locate**

### DIFF
--- a/hybridse/src/udf/default_udf_library.cc
+++ b/hybridse/src/udf/default_udf_library.cc
@@ -911,6 +911,47 @@ void DefaultUdfLibrary::InitStringUdf() {
 
     RegisterAlias("substr", "substring");
 
+    RegisterExternal("locate")
+    .args<StringRef, StringRef>(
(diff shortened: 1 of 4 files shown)


hybridse/src/udf/default_udf_library.cc
```diff
@@ -911,6 +911,44 @@ void DefaultUdfLibrary::InitStringUdf() {
...
+        .args<StringRef, StringRef, int32_t>(
+            static_cast<int32_t (*)(StringRef*, StringRef*, int32_t)>(udf::v1::locate))
+        .doc(R"(
+            @brief Returns the position of the first occurrence of substr in str after position pos. The given pos and return value are 1-based.
+
+            Example:
+
+            @code{.sql}
+
+                select locate("wo", "hello world", 2);
```
vagetablechicken, line 940: Thx for contributing~
Could we give a more representative example? e.g. `locate("o", "hello world", 5)`, it will be 5, `locate("o", "hello world", 6)`, it will be 8. Or the examples in #820.

PR #3945, opened 2024-06-03
feat(udf): array_combine & array_join

aceforeverd:
Two UDFs added

## array_combine

```
array_combine (delimiter, array1, ...)
```

- param **delimiter**: string delimiter for each product
- param **array1**: variadic list of arrays, requires at least one. You may provide array of any primitive type as input arrays, like `intXX, string, date, timestamp`, but not array of composited types. For example, `ARRAY<MAP<int, int>>` is invalid
- return: array of string, each element is a product

**return array of strings for input array1, array2, ... doing cartesian product. Each product is joined with {delimiter} as a string. Empty string used if {delimiter} is null.**

```sql
select array_combine("-", ["1", "2"], ["3", "4"]);  -- ["1-3", "1-4", "2-3", "2-4"]
```

## array_join
```
array_join(arr, delimiter)
```

**Concatenates the elements of the given array using the delimiter. Any null value is filtered.**

```sql
select array_join(["1", "2"], "-");  -- "1-2"
```

### DIFF
--- a/cases/query/udf_query.yaml
+++ b/cases/query/udf_query.yaml
@@ -562,14 +562,16 @@ cases:
         array_join(["1", "2"], "") c2,
         array_join(["1", "2"], cast(null as string)) c3,
         array_join(["1", NULL, "4", "5", NULL], "-") c4,
+        array_join(array<string>[], ",") as c5
     expect:
       columns:
         - c1 string
         - c2 string
         - c3 string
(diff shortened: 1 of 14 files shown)


vagetablechicken (approved):

tobegit3hub (approved):
LGTM

PR #3946, opened 2024-06-04
fix: handle repeated sort keys and sort as int

oh2024:
bug fix

1. `json_array_sort` drops duplicate sort keys
2. `json_array_sort` sorts by string instead of int

1. `json_array_sort` does not drop duplicate sort keys 
2. `json_array_sort` sorts by int instead of string, and breaks ties using output column

### DIFF
--- a/hybridse/src/udf/default_defs/feature_zero_def.cc
+++ b/hybridse/src/udf/default_defs/feature_zero_def.cc
@@ -15,6 +15,8 @@
  */
 
 #include <algorithm>
+#include <charconv>
+#include <iostream>
 #include <queue>
 #include <string>
 #include <tuple>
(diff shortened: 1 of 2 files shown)


hybridse/src/udf/default_defs/feature_zero_def.cc
```diff
@@ -670,27 +672,30 @@ void json_array_sort(::openmldb::base::StringRef *json_array,
       continue;
     }
 
-    container.emplace(order_val, column_val);
+    int order_int;
+    auto [ptr_order, ec_order] = std::from_chars(order_val.data(), order_val.data() + order_val.size(), order_int);
+    if (ec_order != std::errc()) {
+        return;
```
aceforeverd, line 678: continue?

oh2024: he said return if not int

PR #3947, opened 2024-06-05
fix: handle repeated sort keys and sort as int

oh2024:
bug fix

1. `json_array_sort` drops duplicate sort keys
2. `json_array_sort` sorts by string instead of int

1. `json_array_sort` does not drop duplicate sort keys 
2. `json_array_sort` sorts by int instead of string, and breaks ties using output column

### DIFF
--- a/hybridse/src/udf/default_defs/feature_zero_def.cc
+++ b/hybridse/src/udf/default_defs/feature_zero_def.cc
@@ -15,6 +15,8 @@
  */
 
 #include <algorithm>
+#include <charconv>
+#include <iostream>
 #include <queue>
 #include <string>
 #include <tuple>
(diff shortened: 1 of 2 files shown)


hybridse/src/udf/default_defs/feature_zero_def.cc
```diff
@@ -615,7 +616,7 @@ void json_array_sort(::openmldb::base::StringRef *json_array,
 
   std::string_view order_ref(order->data_, order->size_);
   std::string_view column_ref(column->data_, column->size_);
-  std::map<std::string, std::string> container;
+  std::vector<std::pair<int, std::string>> container;
```
aceforeverd: Use int64

aceforeverd (approved):

PR #3952, opened 2024-07-05
refactor: fix compile for mcjit and improve to tests

aceforeverd:
- fix compile with `-DLLVM_EXT_ENABLE=ON`
- setup sql case base dir on compile, no need to specify `SQL_CASE_BASE_DIR` environment every time


tobegit3hub (approved):
LGTM

PR #3963, opened 2024-07-23
fix: select from JOB_INFO should always in online mode

aceforeverd:
Fix error when user set default `execute_mode` to offline:

```sql
set global execute_mode = 'offline';
select 1;
```


tobegit3hub (approved):
LGTM

PR #3964, opened 2024-07-24
ci: update create-pull-request action to v6 in udf-doc-gen workflow & rm deprecated file sync

Jayaprakash0511:
udf doc gen fail issue#3903 (bug).
Updated the create-pull-request (.github/worflows/udf-doc.yml) to v6 which is currently using v4 causing deprecated issues.

### DIFF
--- a/.github/workflows/udf-doc.yml
+++ b/.github/workflows/udf-doc.yml
@@ -54,7 +54,6 @@ jobs:
         if: github.event_name != 'pull_request'
         with:
           add-paths: |
-            docs/en/reference/sql/udfs_8h.md
             docs/zh/openmldb_sql/udfs_8h.md
           labels: |
             udf
--- a/.github/workflows/udf-doc.yml
+++ b/.github/workflows/udf-doc.yml
@@ -50,7 +50,7 @@ jobs:
           make -C hybridse/tools/documentation/udf_doxygen sync
 
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@v4
+        uses: peter-evans/create-pull-request@v6
         if: github.event_name != 'pull_request'
         with:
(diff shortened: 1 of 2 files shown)


aceforeverd:
@Jayaprakash0511 Hi, thanks for your interests. 

However, the workflow fails actually because the file `docs/en/reference/sql/udfs_8h.md` does not exists any more. 

you should at least update the configuration in both:
- https://github.com/4paradigm/OpenMLDB/blob/96976f50b53d9b18ed6f35d01be95a91b4e185a9/hybridse/tools/documentation/udf_doxygen/Makefile#L31-L31
- https://github.com/4paradigm/OpenMLDB/blob/96976f50b53d9b18ed6f35d01be95a91b4e185a9/.github/workflows/udf-doc.yml#L57

Jayaprakash0511:
> @Jayaprakash0511 Hi, thanks for your interests.
> 
> However, the workflow fails actually because the file `docs/en/reference/sql/udfs_8h.md` does not exists any more.
> 
> you should at least update the configuration in both:
> 
> * https://github.com/4paradigm/OpenMLDB/blob/96976f50b53d9b18ed6f35d01be95a91b4e185a9/hybridse/tools/documentation/udf_doxygen/Makefile#L31-L31
> * https://github.com/4paradigm/OpenMLDB/blob/96976f50b53d9b18ed6f35d01be95a91b4e185a9/.github/workflows/udf-doc.yml#L57

Thank you for your response! Seeing the errors , I thought there were only deprecated issues related to Buffer(), hence changed the pull request version to the latest. Now i have removed references to docs/en/reference/sql/udfs_8h.md as the file no longer exists.

aceforeverd (approved):

PR #4028, opened 2025-08-04
fix: fix cicd

dl239:
- `actions/upload-artifact@v3` and `actions/download-artifact@v3` has been deprecated. https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
- The macOS 12 Actions runner has been deprecated. https://github.com/actions/runner-images/issues/10721


dl239:
cpp test case will be fixed in another pr

PR #4114, opened 2026-05-13
fix: preserve env variables for SparkLauncher

### LINKED ISSUE #3926: hadoop.conf.dir env bug
https://github.com/4paradigm/OpenMLDB/blob/6569b42f81cb57393858b74b0658b02ae465d527/java/openmldb-taskmanager/src/main/scala/com/_4paradigm/openmldb/taskmanager/spark/SparkJobManager.scala#L48

After 1792700a97 , sparklauncher miss the env, so 0.8.3<=version<=0.9.0 can't use `hadoop.conf.dir` in taskmanager config.

bitbharat0509:
### What kind of change does this PR introduce?
Bug fix

This PR fixes an issue in `SparkJobManager` where Spark jobs launched through `SparkLauncher` were not preserving existing system environment variables.

---

### What is the current behavior?
Currently, `SparkJobManager` creates a new empty environment map and only adds `HADOOP_CONF_DIR` and `HADOOP_USER_NAME`.

As a result:
- Existing system environment variables such as `PATH`, `JAVA_HOME`, etc. are lost
- Spark jobs may fail when required environment variables are missing
- `hadoop.conf.dir` may not work correctly in some deployments

Root cause: the environment passed to `SparkLauncher` did not preserve existing system environment variables.

---

### What is the new behavior?
This PR initializes the environment map using `System.getenv()` to preserve existing system environment variables.

It then adds:
- `HADOOP_CONF_DIR`
- `HADOOP_USER_NAME`

Finally, the updated environment map is passed to `SparkLauncher`.

This ensures Spark jobs inherit required system environment variables while still supporting custom Hadoop configurations.  
Fixes:#3926

### DIFF
--- a/java/openmldb-taskmanager/src/main/scala/com/_4paradigm/openmldb/taskmanager/spark/SparkJobManager.scala
+++ b/java/openmldb-taskmanager/src/main/scala/com/_4paradigm/openmldb/taskmanager/spark/SparkJobManager.scala
@@ -41,9 +41,7 @@ object SparkJobManager {
     if (TaskManagerConfig.getHadoopUserName != null && TaskManagerConfig.getHadoopUserName.nonEmpty) {
       env.put("HADOOP_USER_NAME", TaskManagerConfig.getHadoopUserName)
     }
-    logger.info("HADOOP_CONF_DIR = " + env.get("HADOOP_CONF_DIR"))
-    logger.info("PATH exists = " + env.containsKey("PATH"))
-    
+
     val launcher = new SparkLauncher(env)
       .setAppResource(TaskManagerConfig.getBatchjobJarPath)
       .setMainClass(mainClass)
@@ -90,7 +88,7 @@ object SparkJobManager {
     val launcher = createSparkLauncher(mainClass)
 
     if (args.nonEmpty) {
-      launcher.addAppArgs(args: _*)
+      launcher.addAppArgs(args:_*)
     }
 
     if (localSqlFile.nonEmpty) {
--- a/java/openmldb-taskmanager/src/main/scala/com/_4paradigm/openmldb/taskmanager/spark/SparkJobManager.scala
+++ b/java/openmldb-taskmanager/src/main/scala/com/_4paradigm/openmldb/taskmanager/spark/SparkJobManager.scala
@@ -29,30 +29,27 @@ object SparkJobManager {
 
   private val logger = LoggerFactory.getLogger(this.getClass)
 
-  /**
-   * Create the SparkLauncher object with pre-set parameters like yarn-cluster.
-   *
-   * @param mainClass the full-qualified Java class name
-   * @return the SparkLauncher object
-   */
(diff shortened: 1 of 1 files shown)


bitbharat0509:
I checked the failing `cicd / cpp` logs.

The failing test seems to be `DBSDK/DBSDKTest.Truncate/1` in the `sql_cmd_test` suite. From the logs, the failure happens during table cleanup/drop operations, where OpenMLDB tries to rename table directories from `db_root` to `recycle_root`, but the source directory is already missing:

`fail to rename ... db_root/... to ... recycle_root/... with error No such file or directory`

This looks unrelated to this PR because my change only updates `SparkJobManager.scala` to preserve environment variables when creating `SparkLauncher`.

Please let me know if you want me to investigate this CI failure further.

java/openmldb-taskmanager/src/main/scala/com/_4paradigm/openmldb/taskmanager/spark/SparkJobManager.scala
```diff
@@ -29,32 +29,27 @@ object SparkJobManager {
...
-   * @param mainClass the full-qualified Java class name
-   * @return the SparkLauncher object
-   */
   def createSparkLauncher(mainClass: String): SparkLauncher = {
 
-    val launcher = new SparkLauncher()
-      .setAppResource(TaskManagerConfig.getBatchjobJarPath)
-      .setMainClass(mainClass)
+    val env: java.util.Map[String, String] =
+      new java.util.HashMap[String, String](System.getenv())
```
dl239: hi @bitbharat0509  thanks for your contribution. Using the environment variables of the submission node may cause conflicts in cluster mode such as `PATH` and `JAVA_HOME`

bitbharat0509:
Hi @dl239, thanks for the review that makes sense.

You're right that copying all variables from `System.getenv()` could introduce conflicts in cluster mode (e.g. `PATH`, `JAVA_HOME`).

I'll update the patch to avoid inheriting all system environment variables and instead only pass the required ones (`HADOOP_CONF_DIR` and `HADOOP_USER_NAME`) to `SparkLauncher`.

dl239:
> Hi @dl239, thanks for the review that makes sense.
> 
> You're right that copying all variables from `System.getenv()` could introduce conflicts in cluster mode (e.g. `PATH`, `JAVA_HOME`).
> 
> I'll update the patch to avoid inheriting all system environment variables and instead only pass the required ones (`HADOOP_CONF_DIR` and `HADOOP_USER_NAME`) to `SparkLauncher`.

hi @bitbharat0509 please init the env with an empty map as described in the comment above.

bitbharat0509:
@dl239  Thanks for the clarification
 
I'll update the implementation to initialize SparkLauncher with an empty   environment map and explicitly add only the required variables (HADOOP_CONF_DIR and HADOOP_USER_NAME).

dl239 (approved):

PR #4122, opened 2026-05-17
fix: support scientific notation for float literals

### LINKED ISSUE #3872: Support scientific notation representation float number

```sql
CREATE TABLE demo_table1 (
	id BIGINT, 
	name STRING, 
	age BIGINT, 
	score FLOAT(53), 
	ts FLOAT(53), 
	dt DATE
)

INSERT INTO demo_table1 (id, name, age, score, ts, dt) VALUES (1, 'Alice', 25, 1.1e0, 1712911045, '2024-01-01')
```

support float value `1.1e0` for field `score`

**Additional context**

bitbharat0509:

Bug fix

Float literals using scientific notation, such as `1.1e0`, may fail when converted as float literals.

In the float32 path, the literal image was truncated by one character before conversion:

`literal->image().size() - 1`

For a value like `1.1e0`, this turns the literal into `1.1e`, which is invalid.

Fixes #3872

This PR preserves the full float literal image when converting float32 literals, allowing values like `1.1e0` to be parsed correctly.

### DIFF
--- a/hybridse/src/planv2/ast_node_converter.cc
+++ b/hybridse/src/planv2/ast_node_converter.cc
@@ -454,7 +454,7 @@ base::Status ConvertExprNode(const zetasql::ASTExpression* ast_expression, node:
             bool is_null;
             if (literal->is_float32()) {
                 float float_value = 0.0;
-                hybridse::codec::StringRef str(literal->image().size() - 1, literal->image().data());
+                hybridse::codec::StringRef str(literal->image().size(), literal->image().data());
                 hybridse::udf::v1::string_to_float(&str, &float_value, &is_null);
                 CHECK_TRUE(!is_null, common::kSqlAstError, "Invalid float literal: ", literal->image());
                 *output = node_manager->MakeConstNode(float_value);
--- a/hybridse/src/planv2/ast_node_converter.cc
+++ b/hybridse/src/planv2/ast_node_converter.cc
@@ -454,7 +454,7 @@ base::Status ConvertExprNode(const zetasql::ASTExpression* ast_expression, node:
             bool is_null;
             if (literal->is_float32()) {
                 float float_value = 0.0;
-                hybridse::codec::StringRef str(literal->image().size(), literal->image().data());
+                hybridse::codec::StringRef str(literal->image().size() - 1, literal->image().data());
                 hybridse::udf::v1::string_to_float(&str, &float_value, &is_null);
                 CHECK_TRUE(!is_null, common::kSqlAstError, "Invalid float literal: ", literal->image());
                 *output = node_manager->MakeConstNode(float_value);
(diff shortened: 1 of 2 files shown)


hybridse/src/planv2/ast_node_converter.cc
```diff
@@ -454,7 +454,7 @@ base::Status ConvertExprNode(const zetasql::ASTExpression* ast_expression, node:
             bool is_null;
             if (literal->is_float32()) {
                 float float_value = 0.0;
-                hybridse::codec::StringRef str(literal->image().size() - 1, literal->image().data());
+                hybridse::codec::StringRef str(literal->image().size(), literal->image().data());
```
DSFans2014: If a float value ends with `f`(e.g. 3.4f), then `is_float32` will be true. Before calling `string_to_float`, we need to remove the `f`.

refer https://github.com/4paradigm/zetasql/blob/a87b2f5b34be1b129d89eb8308a54dc1d4c28d0f/zetasql/parser/parse_tree_manual.h#L3697-L3701

DSFans2014: To support scientific notation, the handling logic should be added in the else block
https://github.com/4paradigm/OpenMLDB/blob/8f4f4af6530bc8efcad24ddb0e0c90cbab73477f/hybridse/src/planv2/ast_node_converter.cc#L461-L467

bitbharat0509:
Hi @DSFans2014 
Thanks for the review.

I reverted the earlier float32 change since that approach was incorrect for literals ending with `f`.

I traced the insert flow through `ast_node_converter`, `string_to_double`, `node_adapter`, and `RowBuilder2`, and I couldn’t find an obvious parsing issue in those paths.

To validate the original report properly, I added a regression test for inserting scientific notation (`1.1e0`) into a float column. This should help confirm whether the issue still reproduces or if the current implementation already handles it correctly.

DSFans2014:
> Hi @DSFans2014 Thanks for the review.
> 
> I reverted the earlier float32 change since that approach was incorrect for literals ending with `f`.
> 
> I traced the insert flow through `ast_node_converter`, `string_to_double`, `node_adapter`, and `RowBuilder2`, and I couldn’t find an obvious parsing issue in those paths.
> 
> To validate the original report properly, I added a regression test for inserting scientific notation (`1.1e0`) into a float column. This should help confirm whether the issue still reproduces or if the current implementation already handles it correctly.

hi @bitbharat0509 I checked the code and verified it, and it does support scientific notation.
The `strtod` function that is called inside `string_to_double` supports scientific notation. refer https://en.cppreference.com/cpp/string/byte/strtof

the test result as below:
ScreenShot_2026-05-19_103411_683

Sorry for the misleading information in my previous comment. And thank you very much for your attention and contribution to OpenMLDB.

